PR #2, opened 2020-06-24
demonstrate lerna use

wiese:
This shows 
* how lerna is added to the project
* two packages `tokens` & `vue-components` are managed via lerna
* how `tokens` (which is in fact private) is installed as a dep of `vue-components`
  * `tokens` has a simulated `dist` folder which contains something that is used in `vue-components`
* how the usual suspects of root-level npm commands (`install`, `test`) invoke the respective commands on the sub packages

TODO
* does this all make sense with us trying to have a storybook which shows information based on the `tokens` and the `vue-components` package?


lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "packages": [
+    "packages/*"
```
jakobw, line 3: I would like it better if the packages are visible on the root level of the repository without the extra `packages/` directory. If there are more practical benefits then I'd of course be fine with having it, but if this glob here is the only one, then I'd rather list them here one by one.

packages/vue-components/package.json
```diff
@@ -0,0 +1,24 @@
...
+    "url": "git+https://github.com/wmde/wikit.git"
+  },
+  "scripts": {
+    "test": "echo \"Tests of vue-components\""
+  },
+  "bugs": {
+    "url": "https://github.com/wmde/wikit/issues"
+  },
+  "dependencies": {
+    "@wmde/wikit-tokens": "file:../tokens"
```
jakobw, line 22: This makes my head hurt a little. The difference between depending on a published npm package and pointing to the source of the package as a file dependency as it's done here is that the published package will of course contain only what's specified in the `files` field in its `package.json`, whereas the file dependency contains whatever the directory contains. This means that this sub package isn't self sufficient in that you would always first have to build the `tokens/` package or otherwise the components package would be missing the build output.

That said, I guess that's the whole point of lerna? The sub package doesn't need to be self-sufficient and can be interdependent because I'm not supposed to install/build/publish anything from here and instead do it from the root?

wiese:
This is replaced by #6, doing it proper.

PR #3, opened 2020-06-25
Vue comps

wiese:
Really rough way of demoing the use of the (only) component library's storybook to show the tokens as well to reduce the amount of system we have to maintain.


wiese:
This is replaced by #9, #10, #12

PR #4, opened 2020-06-25
Add test workflow for CI

tzhelyazkova:
Setup CI with GitHub Workflows following https://help.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,22 +1,14 @@
 # This workflow will do a clean install of node dependencies, build the source code and run tests
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
+on: push
 name: Test
-
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
-
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.x'
+    - name: Install
+      run: npm install
+    - name: Test
+      run: npm run test
```
wiese, line 19: Call me old-fashioned but I like my newlines at the ends of files.
```suggestion
      run: npm run test

```

.github/workflows/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.x'
+    - name: Install
+      run: npm install
```
wiese, line 17: Obviously both do very similar things but, among others reasons, I saw a difference of 5s in run time. I assume `ci` does not run the update check.

```suggestion
      run: npm ci
```

PR #5, opened 2020-06-25
add npm root package

wiese:
The repository root will become the top layer of a lerna project.
We want to build CI first, so add a basic npm layer to be run there.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "wikit",
+  "private": true,
+  "license": "BSD-3-Clause",
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "wikit",
+  "private": false,
```
jakobw, line 3: Shouldn't this one say `true`? We publish sub packages, if ever, right?

wiese: Ha. I have no idea how I got this one wrong.

wiese: Thank you for paying attention!

jakobw:
Needs a rebase, but otherwise good to go!

PR #6, opened 2020-06-26
Add initial lerna setup

jakobw:
This adds the two packages `tokens` and `vue-components`, and models
the dependency between the two.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
     "url": "git+https://github.com/wmde/wikit.git"
   },
   "scripts": {
-    "test": "echo \"wikit tests\" && exit 0"
+    "test": "lerna run test"
+  },
+  "devDependencies": {
+    "lerna": "^3.22.1"
   }
 }
--- a/tokens/package.json
+++ b/tokens/package.json
@@ -1,0 +1,22 @@
+{
+  "name": "@wmde/wikit-tokens",
+  "version": "0.0.0",
+  "description": "Design tokens for wikit in different formats",
+  "author": "The Wikidata team",
+  "homepage": "https://github.com/wmde/wikit/tree/master/tokens#readme",
+  "license": "BSD-3-Clause",
+  "main": "dist/index.json",
+  "files": [
+    "dist"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wmde/wikit.git"
+  },
+  "scripts": {
+    "test": "echo \"wikit tokens tests\" && exit 0"
+  },
+  "bugs": {
+    "url": "https://github.com/wmde/wikit/issues"
+  }
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,5 +12,6 @@ ARG GID=1000
 
 RUN addgroup --gid $GID runuser && adduser --uid $UID --gid $GID --disabled-password --gecos "" runuser
 RUN npm install --global npm@6.14.5
+RUN npm install --global lerna@3.22.1
 
 USER runuser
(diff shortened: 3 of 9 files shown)


tokens/package.json
```diff
@@ -0,0 +1,22 @@
+{
```
wiese, line 1: Maybe I'm being inconsistent, but do we want this to be public or not?

I don't mind it, but I'm pondering: Not (even remotely) every change to the vue-components will result in a change to the tokens. Would be weird to publish no-diff changes in the tokens every time. Should we use [`--independent`](https://github.com/lerna/lerna/tree/master/commands/init#--independent) then? Or am I missing something?

package.json
```diff
@@ -8,5 +8,8 @@
   },
   "scripts": {
     "test": "echo \"wikit tests\" && exit 0"
+  },
+  "devDependencies": {
+    "lerna": "^3.22.1"
```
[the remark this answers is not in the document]
wiese: I guess `package-lock.json` is addressed already.

tokens/package.json
```diff
@@ -0,0 +1,22 @@
...
+  "main": "dist/index.json",
+  "files": [
+    "dist"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/wmde/wikit.git"
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1"
```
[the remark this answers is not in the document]
wiese: And with "kicking them out" I actually mean: make them emit a message that the sub-module was tested and exit with 0.

tokens/package.json
jakobw: As far as I know they would get published together, even with no-diff changes. I'm in favor of just making this private for now until there is a use case.

tokens/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@wmde/wikit-tokens",
+  "version": "0.0.0",
+  "description": "Design tokens for wikit in different formats",
+  "author": "The Wikidata team",
+  "homepage": "https://github.com/wmde/wikit#readme",
+  "license": "BSD-3-Clause",
+  "main": "dist/index.json",
+  "files": [
+    "dist"
```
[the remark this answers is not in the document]
jakobw: Added!

tokens/package.json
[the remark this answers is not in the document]
jakobw: Done

tokens/package.json
wiese: So after some team discussion we concluded that our existing apps already want/need a more elegant way to get to token than the build script that exists copied over and over. Publishing the tokens right away is the way to go then.
We looked a vue-cli and how it manages its sub-packages and concluded that it is also not absurd to increment all version when publishing a change to only one sub-package and that it is a nice touch to be able to verify all packages are in sync by seeing the same version number in all of them (in consuming package.jsons).

package.json
```diff
@@ -8,5 +8,8 @@
   },
   "scripts": {
     "test": "echo \"wikit tests\" && exit 0"
```
wiese, line 10: In order for our CI to be able to pick up the sub packages we either have to tell it to `lerna run test` or instruct the root level package's test script (it is currently hooked up to CI) to do it instead.

I'm in favor of the latter (even if this is not what the lerna authors intended) because everyone knows how to interpret a `package.json` but only few will expect a test target other than `npm test` from being run in CI.

jakobw: Yup, I agree. Changed the test script to invoke lerna.

PR #14, opened 2020-06-29
Filter out inherited wikimedia-ui-base variables

jakobw:
Same as https://github.com/wmde/wikit-css/pull/25


tokens/.style-dictionary/config.js
```diff
@@ -0,0 +1,51 @@
...
+	include: ['node_modules/wikimedia-ui-base/tokens.json'],
+	source: ['properties/**/*.json'],
+	platforms: {
+		scss: {
+			transformGroup: 'scss',
+			buildPath: 'dist/',
+			files: [{
+				destination: '_variables.scss',
+				format: 'scss/variables',
+				filter: 'removePrefixedVars'
```
wiese, line 13: `removePrefixedVars` sounds rather generic but by the time it gets registered it is highly opinionated already (looking for the wmf-prefix). Maybe we should reflect this fact in its name to make the config more self-explanatory.

```suggestion
				filter: 'removeBaseTokensByWMF'
```

wiese: (changed my mind about the name and edited)

tokens/.style-dictionary/config.js
```diff
@@ -0,0 +1,52 @@
...
+					filter: 'removePrefixedVars',
+				} ],
+			},
+			json: {
+				transformGroup: 'web',
+				buildPath: 'dist/',
+				files: [ {
+					destination: 'index.json',
+					format: 'json',
+					filter: 'removePrefixedVars',
```
wiese, line 41: It appears([0]) we can even use a pointer to a named function so no need for magic strings.

[0]: https://amzn.github.io/style-dictionary/#/formats?id=filtering-tokens

jakobw: Oh great! Changed it and also moved it to a separate file.

PR #16, opened 2020-06-30
docs: some details about storybook

wiese:


docs/.eslintrc.json
```diff
@@ -0,0 +1,16 @@
+{
+	"extends": [
+		"wikimedia",
+		"wikimedia/node",
+		"plugin:json/recommended"
+	],
```
jakobw, line 6: While `"wikimedia/node"` makes sense for our `tokens/` dir which only contains node.js code for building the tokens, the `docs/` and `vue-components/` folders are client code. I suggest something like the following eslint config:
```suggestion
	"extends": [
		"wikimedia",
		"wikimedia/client",
		"wikimedia/language/es6",
		"plugin:json/recommended"
	],
	"parserOptions": {
		"sourceType": "module"
	},
```

jakobw:
> Why change from arrow function expression to `function`? We are not using `this` anywhere.

I also prefer declaring functions using the `function` keyword. In this case it makes very little practical difference, but makes it slightly more readable in my opinion. I like to stick to arrow functions for callbacks, and `function` for... functions.

tzhelyazkova:
> > Why change from arrow function expression to `function`? We are not using `this` anywhere.
> 
> I also prefer declaring functions using the `function` keyword. In this case it makes very little practical difference, but makes it slightly more readable in my opinion. I like to stick to arrow functions for callbacks, and `function` for... functions.

Okay, so it is a matter of opinion. It seems two against one, so I'm fine with using `function`, but then we should somehow express that in common linting rules. e.g.
https://eslint.org/docs/rules/prefer-arrow-callback

Can't find anything about not using `=>` when it's not a callback.

jakobw:
Latest force-push is a no-diff rebase.

PR #17, opened 2020-06-30
tokens: add basic jest setup

jakobw:


tokens/tests/structureForStyleDictionary.spec.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const { structureForStyleDictionary } = require( './../build/wikimedia-ui-base/lib' );
+
+describe( 'structureForStyleDictionary', () => {
+	it( 'prefixes imported variables', function () {
```
wiese, line 6: Let's make this arrow syntax so that we have a good example for copy-paste development (me).

tokens/tests/.eslintrc.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
```
wiese, line 1: IDK - this just does not feel right.
I don't like the quotes in JSON either, but this boilerplate even less.

tokens/tests/structureForStyleDictionary.spec.js
jakobw: Done

tokens/tests/.eslintrc.js
jakobw: Done

PR #19, opened 2020-06-30
Add list of referenced tokens to json build

jakobw:
https://github.com/wmde/wikit-css/pull/24 but with tests and actually using it in storybook

Bug: [T256022](https://phabricator.wikimedia.org/T256022)

### DIFF
--- a/docs/src/lib/flattenTokenTree.js
+++ b/docs/src/lib/flattenTokenTree.js
@@ -16,7 +16,7 @@ export function flattenTokenTree( nestedTokens ) {
 
 		return tokens.concat( {
 			name: node.name,
-			referencedTokens: node.original.value.replace( /{|}|.value/gi, '' ),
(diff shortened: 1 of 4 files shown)


docs/src/lib/flattenTokenTree.js
```diff
@@ -16,7 +16,7 @@ export function flattenTokenTree( nestedTokens ) {
 
 		return tokens.concat( {
 			name: node.name,
-			referencedTokens: node.original.value.replace( /{|}|.value/gi, '' ),
+			referencedTokens: node.attributes.tokens.join( ', ' ),
```
wiese, line 19: I'm a bit unsure if gluing this by comma here is going too far into the presentation territory already (could be done just as well in the respective story) OTOH it deduplicates which is good. And it is currently a no-op anyway (no situations of two tokens informing another AFAICT).

PR #23, opened 2020-06-30
Visualize color opacity tokens in Storybook

tzhelyazkova:
- Rename 'hex' to 'value' because 'hex' does not make sense for color opacity values.
- The opacity is visualized in a native `html` `table`, because the docs addon does not seem to have a component of that sorts


tzhelyazkova:
Now that we have new function name for traversing the tokens, and also have renamed the info object, this PR is pretty much all wrong. Closing, will follow up in another one.

PR #25, opened 2020-06-30
Add opacity tokens page to Storybook

tzhelyazkova:


tzhelyazkova:
> Looks good code-wise. After seeing the shiny color token page I'm a bit sad to only have a table here, but I guess we can iterate on that if needed :D
> Needs a manual rebase after the structural changes from #24 (`opacity` is no longer part of `color`).

Yep, I didn't find anything in the docs addon that can help us illustrate opacity. If we decide to mix the opacity with the colors I guess that should happen in the Visual Guide section and with a lot of help from UX.

PR #31, opened 2020-07-02
linting: prefer declaration style of function definition

wiese:
This was discussed in https://github.com/wmde/wikit/pull/16/ and
suggests on of the two mentioned approaches.

See https://eslint.org/docs/rules/func-style#declaration

### DIFF
--- a/docs/.eslintrc.json
+++ b/docs/.eslintrc.json
@@ -14,6 +14,7 @@
 			"imports": "always-multiline",
 			"exports": "always-multiline",
 			"functions": "always-multiline"
-		} ]
+		} ],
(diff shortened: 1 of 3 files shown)


wiese:
Maintaining two independent configuration pains me. I have yet to understand what is the proper way of avoiding this. Of course, eslint is capable of traversing the tree looking for configuration files, but this requires the folders to be part of the same package with all the plugins centrally installed. This is not the case with us as we _want_ the packages to be able to track their dependencies independently. Maybe we should make an exception for eslint et al (they are not really _dependencies_ of these packages anyway) and hoist it and its root configuration to the wrapping lerna package... Either way a conceptually independent step.

PR #34, opened 2020-07-02
Use Chromatic with Storybook

tzhelyazkova:
Login with your GitHub account and you'll be able to see the wmde/wikit Builds
https://www.chromatic.com/builds?appId=5efdb3b5f65950002286285d

### DIFF
--- a/.github/workflows/chromatic.yaml
+++ b/.github/workflows/chromatic.yaml
@@ -1,0 +1,17 @@
+
+on: push
+name: Test Deploy
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+    - name: Set up node
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.x'
+    - name: Install
+      run: npm ci
+    - name: Deploy Storybook
+      run: npm run chromatic
--- a/docs/.storybook/main.js
+++ b/docs/.storybook/main.js
@@ -1,4 +1,11 @@
 module.exports = {
 	stories: [ '../src/**/*.stories.(js|mdx)' ],
-	addons: [ '@storybook/addon-docs' ],
+	addons: [
+		{
+			name: "@storybook/addon-docs",
+			options: {
+				configureJSX: true,
+			}
+		}
+	],
 };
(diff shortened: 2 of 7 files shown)


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        npm install
+        npm run build:storybook
+    - uses: chromaui/action@v1
+      with: 
+        projectToken: 3evi132fpys
+        token: ${{ secrets.GITHUB_TOKEN }}
```
wiese, line 15:
```suggestion
        projectToken: 3evi132fpys
        token: ${{ secrets.GITHUB_TOKEN }}
        storybookBuildDir: docs/dist

```

This way it knows which folder to upload, which was already build be the `build:storybook` script.

https://github.com/chromaui/action#usage-all-options

We should, hope that's possible, also try and get a new `projectToken` and hide it behind a github secret.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,15 @@
+name: "Chromatic"
+on: push
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        npm install
```
wiese, line 10:
```suggestion
        npm ci
```

tzhelyazkova:
The Storybook is built https://5efdb3b5f65950002286285d-levymnrwym.chromatic.com/?path=/story/design-tokens-colors--page
but the CLI (and respectively the action) fails to verify the instance of Storybook because we don't have any stories yet.

.github/workflows/chromatic.yaml
```diff
@@ -0,0 +1,17 @@
+
+on: push
+name: Test Deploy
+jobs:
+  build-test:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
```
wiese, line 9: As discussed before:

I can not prove it b/c I don't know chromatic good enought to understand where the problems would show but I assume that this statement [here](https://github.com/chromaui/action#checkout-depth) is true no matter if we do the chromatic upload via their github action or npm script - if there is no git history from the `actions/checkout` action, then there is nothing to upload.

Consequently we must either use v1 (which seems to do all the history by default) _or_ add `fetch-depth: 0` and use v2 of `actions/checkout`. I'm in favor of the former.

```suggestion
      uses: actions/checkout@v1
```

PR #35, opened 2020-07-02
Improve typography example phrase

jakobw:
:fox_face:


jakobw:
I managed to accidentally sneak that into #41. I also see things cut off. Odd. We can hope for that to get fixed in storybook, or we can manipulate the line-height within `Typeset` if it gets sufficiently annoying.

PR #36, opened 2020-07-02
docs: use link icon to illustrate anchor capability

wiese:
Dabbling to demonstrate a possible solution.
Of course, should be applied to all token pages alike if it makes sense.


wiese:
Is solving a non problem. Something which came up in https://phabricator.wikimedia.org/T256861#6274287.
If you think it's silly we can drop it, but I thought it would be cool to massage this into the reusable component we are planning.

wiese:
Should go after #41

wiese:
> Should go after #41

Done

PR #38, opened 2020-07-02
docs: illustrate transitions

wiese:
Bug: [T256859](https://phabricator.wikimedia.org/T256859)


wiese:
Should wait for #41

tzhelyazkova:
> Should wait for #41

oh :ok:

wiese:
> The duration table changes cell size when the transition rectangle is hovered..but only sometimes.

Yes. That's my sloppy job on the presenter. The width becomes smaller so that you don't hover it anymore which in turn stops the transition.

wiese:
> Should wait for #41

Done

PR #42, opened 2020-07-03
TokenTable: fix token links

jakobw:
This applies the default link styles and behavior from storybook to token links and prevents them from opening the story frame instead of navigating to the token anchor within the storybook.


wiese:
Obviously!

PR #43, opened 2020-07-03
docs: align TokenTable with ColorPallete

wiese:
Somewhat takes away from the semantics, but aligns those token pages visually 
and creates a bit more room for the presenters. Also somewhat explains to the
users the difference between primary values and values influenced by other
tokens.

Should probably recycle the very elements if we really mean to achieve
identical layout.

See
https://github.com/storybookjs/storybook/blob/b6136e1539c85d253504391a7d3f65e2c1239143/lib/components/src/blocks/ColorPalette.tsx#L170


wiese:
Rebased

PR #46, opened 2020-07-03
docs: add line-heights

wiese:
Bug: [T256855](https://phabricator.wikimedia.org/T256855)

### DIFF
--- a/docs/src/10-font-line-height.stories.mdx
+++ b/docs/src/10-font-line-height.stories.mdx
@@ -1,0 +1,20 @@
+import { Meta, Typeset } from '@storybook/addon-docs/blocks';
+import { TokenTable } from './lib/TokenTable';
+import { font } from '@wmde/wikit-tokens';
+import styles from './styles/line-height.css';
+
+<Meta title="Design Tokens|Font line-height" />
+
+## line-height
+
+<TokenTable
+    tokens={font[ 'line-height' ]}
+    valueCell={( value ) => (
+        <div>
+            <pre>{ value }</pre>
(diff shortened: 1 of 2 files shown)


wiese:
Want to group this with the rest of the font stuff but does not make much sense while there are multiple PRs in progress competing for those incrementing ids.

wiese:
rebased

docs/src/10-font-line-height.stories.mdx
```diff
@@ -0,0 +1,18 @@
...
+<Meta title="Design Tokens|Font line-height" />
+
+## line-height
+
+<TokenTable
+    tokens={font[ 'line-height' ]}
+    valueCell={( value ) => (
+        <div>
+            <pre>{ value }</pre>
+            <div className='lineHeight' style={{ lineHeight: value }}>demo</div>
```
tzhelyazkova, line 15:
```suggestion
            <div className='lineHeight' style={{ lineHeight: value }}>
            The quick brown fox jumps over the lazy dog
            </div>
```
I think the visuals can benefit from a longer sentence. It makes the line height visible between lines as well.
Like here https://www.lightningdesignsystem.com/design-tokens/#category-line-height

wiese: Hmm. I guess for me it became less clear, or maybe I misunderstood.

tzhelyazkova: Ok, we can ask UX first, before we merge this one.
I think line-height is most important in paragraphs.

wiese:
I also removed the horizontal alignment which did not make sense with an intentional multiline.

PR #49, opened 2020-07-03
chromatic: name build nicely

wiese:
This is shown in every PR so naming it nicely makes sense.

Bug: [T256864](https://phabricator.wikimedia.org/T256864)

Screenshot_2020-07-03 docs add break-points by wiese · Pull Request #45 · wmde wikit


wiese:
Made me wondering... We should probably only perform the chromatic step if the request of the tests succeeded to go easy on them resources.

PR #50, opened 2020-07-03
chromatic: upload only when otherwise green

wiese:
Building and uploading storybook is quite expensive and probably only
makes sense if a patch is otherwise fine. Consequently, we can chain it
behind the rest of the tests which already exists (which get renamed to
reflect that this pipeline is now a bit more involved).

Bug: [T256864](https://phabricator.wikimedia.org/T256864)


wiese:
Updated the branch protection to match the new name in the action.

PR #52, opened 2020-07-03
docs: reflect clicked anchors in browser URL

wiese:
So that the deep links can be used to communicate between people.

Bug: [T256861](https://phabricator.wikimedia.org/T256861) (mentioned there)

There was a bug report of sorts in https://phabricator.wikimedia.org/T256861#6276737 and this hopefully solves it.


wiese:
rebased

PR #53, opened 2020-07-04
docs: alternative ways to present tokens

wiese:
Adds
* a page showing all tokens, inferring their presenter
* a dashboard grouping tokens into the categories globals & aliases (all others, in fact)


wiese:
Beef moved into #54

PR #55, opened 2020-07-06
docs: style referenced tokens

wiese:
Create UX-desired style for referenced tokens in TokenTable.
Unfortunately the specification only happened in chat.

Also replace the "primary value" information and custom style with a
dash because that apparently is more clear.


wiese:
> Snarky commit message :D

Not my brightest moment.

PR #57, opened 2020-07-06
Make token presenter inference optional

jakobw:


docs/src/05-font-size.stories.mdx
```diff
@@ -1,9 +1,10 @@
...
 import { TokenTable } from './lib/TokenTable';
+import FontSize from './lib/presenters/FontSize';
 import { dimension } from '@wmde/wikit-tokens';
 
 <Meta title="Design Tokens|Font size" />
 
 ## Font size
 
-<TokenTable tokens={dimension.font.size} />
+<TokenTable tokens={dimension.font.size} Presenter={FontSize} />
```
wiese, line 10: I'm probably missing a coffee or two but what do we gain by e.g. this? The same presenter would be used automatically, right?

PR #58, opened 2020-07-06
Improve shadow tokens visualization

tzhelyazkova:
Style changes are based on feedback from UX provided in [T256858](https://phabricator.wikimedia.org/T256858#6277435)

### DIFF
--- a/docs/src/lib/presenters/Shadows.jsx
+++ b/docs/src/lib/presenters/Shadows.jsx
@@ -3,8 +3,8 @@ import React from 'react';
 
 export default function ( { token } ) {
 	return (
-		<div className="shadows" style={{ boxShadow: token.value }}>
-			<pre>{ token.value }</pre>
+		<div>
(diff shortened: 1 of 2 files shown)


docs/src/lib/presenters/Shadows.jsx
```diff
@@ -3,7 +3,8 @@ import React from 'react';
 
 export default function ( { token } ) {
 	return (
-		<div className="shadows" style={{ boxShadow: token.value }}>
+		<div>
+			<div className="shadows" style={{ boxShadow: token.value }} />
 			<pre>{ token.value }</pre>
```
jakobw, line 8:
```suggestion
```

With 7fe1a2ffe0d9c17d903f931148398c7a597c2ca4 the token value is always shown, and no longer part of the presenter.

tzhelyazkova: oh cool, thanks!

PR #60, opened 2020-07-07
Improve line height storybook page

tzhelyazkova:
Improvements made are based on feedback provided by UX here [T256855](https://phabricator.wikimedia.org/T256855#6277459)

### DIFF
--- a/docs/src/lib/presenters/LineHeight.css
+++ b/docs/src/lib/presenters/LineHeight.css
@@ -3,7 +3,11 @@
  * Apply to an element (with text content), pass a custom line-height.
  */
 .lineHeight {
-    background-color: #36c;
-    color: #fff;
-    width: 200px;
+    border-radius: 4px;
+    background: rgb(255, 255, 255) none repeat scroll 0% 0%;
+    box-shadow: rgba(0, 0, 0, 0.1) 0px 1px 3px 0px;
+    border: 1px solid rgba(0, 0, 0, 0.1);
+    margin: 25px 0px 40px;
+    padding: 30px 20px;
+    font-size: 16px;
 }
--- a/docs/src/lib/presenters/LineHeight.jsx
+++ b/docs/src/lib/presenters/LineHeight.jsx
@@ -4,7 +4,8 @@ import React from 'react';
 export default function ( { token } ) {
 	return (
 		<div className='lineHeight' style={{ lineHeight: token.value }}>
-			The quick brown fox jumps over the lazy dog
+			The quick brown jumps<br />
+			over the lazy dog<br />
 		</div>
 	);
 }


wiese:
I assume the motivation to replicate `Typeset` even though it can be used as an independent "swatch" (c.f. `FontSize.jsx`) lies in getting it to go multi-line.

Technically I guess this is ok then, but I have the hardest time wrapping my head around how those requirements (and consequently also how we implement them) make sense. Let's see if UX green-lights this.

tzhelyazkova:
> I assume the motivation to replicate `Typeset` even though it can be used as an independent "swatch" (c.f. `FontSize.jsx`) lies in getting it to go multi-line.
> 
> Technically I guess this is ok then, but I have the hardest time wrapping my head around how those requirements (and consequently also how we implement them) make sense. Let's see if UX green-lights this.

I looked at Typeset's implementation, and they do not take line height as a property. That, to me, means Typeset will be "misused" for displaying line heights.

wiese:
As said already

> I have the hardest time wrapping my head around how those requirements [...] make sense

so I'd like to hear @jakobw's opinion. Good to be merged if he does not veto as I don't think there is anything _technically_ wrong with it, I just don't understand the _why_.

tzhelyazkova:
> As said already
> 
> > I have the hardest time wrapping my head around how those requirements [...] make sense
> 
> so I'd like to hear @jakobw's opinion. Good to be merged if he does not veto as I don't think there is anything _technically_ wrong with it, I just don't understand the _why_.

This assumes you're talking about the requirement to have the text in multiple lines (without the blue background). If so:
I think it's a pretty standard way to display `line-height`, because it's ultimately about paragraphs. 
MDN does it https://developer.mozilla.org/en-US/docs/Web/CSS/line-height
The lightning design system as well https://www.lightningdesignsystem.com/design-tokens/#category-line-height

jakobw:
I think presenting line-height by having multiple lines makes perfect sense. The different font sizes per individual token don't make a lot of sense to me, are bound to look really silly the next time the token names are changed, and arguably make it harder to see differences. Neither of the two pages linked in the comment above do that. I'm not vetoing it if that's what it's supposed to do :man_shrugging:

wiese:
> This assumes you're talking about the requirement to have the text in multiple lines

No - multiple lines make total sense.

It is the change of font _size_ depending on the token value (line-height), thereby changing two variables at once and making it impossible to grasp to impact of one or the other. I simply do not understand how it helps and was hoping for someone to open my eyes about this one, because I still perceive it as detrimental.

SaiSan-WMDE:
Line-height is strictly font-size dependent. I encouraged applying the different line-heights to its corresponding font sizes because I thought that it would be nice to make that line-height/font-size connection explicit here by using TypeSet (which displays the font size applied by default). But since said block cannot be used, I'd agree that we can go back to using 16px: although imprecise. Line-height is not about how much space is created between lines of text, but about the readability improvement of a given text-size. I believe that the introduction of alias tokens in this page will make more apparent the need to establish this connection. Hopefully and thankfully, the token names will support understanding.

wiese:
Incorporated the feedback. This also addresses what @jakobw already mentioned earlier, wrong font sizes the "next time the token names are changed", which happens also when alias tokens are added which do not follow the global's naming convention.

tzhelyazkova:
I think we established that tying `line-height` to `font-size` is the right thing to do. And it will probably be done in the Visual Guide section of Storybook. It's anther question if this visual guide will be using tokens or not.

I'd be happy if @jakobw merges this as the only one who did not amend the code so far, so we can finally be done with this small but very expensive (money wise) change. peace :v:

PR #62, opened 2020-07-07
document connection between tokens and docs

wiese:

### DIFF
--- a/tokens/README.md
+++ b/tokens/README.md
@@ -1,1 +1,24 @@
 # Wikit Tokens
+
+This package holds the Wikit design tokens and tools needed to build them.
+
+## Package organization
+
+We organized the tokens in three categories
+* globals
+* aliases
+* component-level
+
+Each category serves its own purpose.
+
+    ⚠️ TBD
+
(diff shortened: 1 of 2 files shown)


tokens/README.md
```diff
@@ -1 +1,24 @@
...
+## Package organization
+
+We organized the tokens in three categories
+* globals
+* aliases
+* component-level
+
+Each category serves its own purpose.
+
+    ⚠️ TBD
```
wiese, line 14: I guess @SaiSan-WMDE may want to chime in here in a follow-up commit.

SaiSan-WMDE: ""Global tokens reflect the elementary visual foundations of the system. They are context agnostic (e.g. "accent-30"). Aliases are abstractions that communicate the intended context of use of a token. They mostly represent styles that are shared among several components (e.g. "border-color-progressive-active"). Finally, component-level tokens are used to represent every value associated with the style of a component (e.g."button-primary-progressive-border-color-active").""

Let me know if this matches expectations, and if it's not too wordy.

wiese: Makes sense to me and matches what we discussed so far.

Given you have the tools and skills, feel free to turn this into an edit on this branch.
You can do this e.g. also (we live in the future) via the "insert a suggestion" feature right here in the comment editor.

PR #63, opened 2020-07-07
Improve json structure

SaiSan-WMDE:
This is an attempt to adapt the json files to a logic structure that can be consistently interpreted by the documentation pages, so we can make sure that all current and future tokens are displayed in Storybook.

I tried to comply with a set of "rules", but this quite an error prone task, so I would appreciate feedback regarding the whole structure and the changes applied to the files. The rules are:

- Root level token branches are shown on an individual sections
- Second levels are captured as subpages (nested within the sections) and include an h2 that equals the name of the subpage
- Third levels (if they exist) are captured as sections inside the subpages, and introduced by an h3
- Further nesting levels are shown below h2 or h3 headlines (depending on the previous number of levels) and in their respective tables
- Tokens inside those sections are show in the order they are stored in the source JSON file. Tokens extracted from the global.json file are displayed on top, tokens extracted from the alias.json file are included

An overview of the structure of the pages is also provided in [this spreadsheet](https://docs.google.com/spreadsheets/d/1Od9m25uBMMmbPirazvwYQl8QVyx0FSm0Eo0jIMj4_no/edit#gid=1159537700&range=1:1).


jakobw:
Closing this as its contents moved into #73 which is now merged.

PR #64, opened 2020-07-07
correct wrong border color value

SaiSan-WMDE:
Visualization helped detect that an incorrect value had been provided for the token border.color.neutral.focus


wiese:
Rebased

PR #65, opened 2020-07-08
modify width and height of the Shadows presenter

SaiSan-WMDE:
Changing the height and width of the Shadows element to make it consistent with the Color and Transition presenters in our token documentation pages.

Bug: [T256858](https://phabricator.wikimedia.org/T256858)


wiese:
rebased

PR #68, opened 2020-07-09
docs: demo line heights at various font sizes

wiese:
The idea was mentioned in a meeting - maybe it helps illustrate the line
heights.

Bug: [T256855](https://phabricator.wikimedia.org/T256855)

### DIFF
--- a/docs/src/lib/presenters/LineHeight.css
+++ b/docs/src/lib/presenters/LineHeight.css
@@ -11,3 +11,12 @@
     padding: 30px 20px;
     font-size: 16px;
 }
+
+.lineHeightContainer:not(:last-of-type) {
+    margin-bottom: 2ex;
+}
+
+.lineHeightContainer dt {
+    margin-bottom: 1ex;
+}
+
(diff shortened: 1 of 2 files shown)


SaiSan-WMDE:
Hey, Pablo! I think you indeed quickly mentioned the idea of creating a "matrix" for the line-height page during our meeting, and I appreciate the exploration, but we didn't get to discuss or align on this.

So, I know I'm leaving my feedback here without being asked for it, but I would not recommend continuing on this direction. First, because we just agreed that showcasing only 1 variable per token page (therefore unifying the font-size here to make it irrelevant, using 16px) was the most consistent and logical thing to do. This was also preferred by the members of the UX team (as I shared in our channel earlier). Secondly, if sizes are introduced as a variables here, then we would have a tokens page and a visual guidelines page contradicting each other (some of the combinations displayed are not recommended or part of the system's typographic foundations), which could be confusing. Hope you agree!

SaiSan-WMDE:
Hey, @wiese! I appreciate the exploration but I would not recommend continuing on this direction.

First, because we just agreed that showcasing only 1 variable per token page was the most consistent and logical thing to do. This was also preferred by the members of the UX team (as I shared in our channel earlier). Secondly, if multiple sizes are introduced as variables here, then we would have a tokens page and a visual guidelines page contradicting each other (some of the combinations displayed are not recommended or part of the system's typographic foundations), which could be confusing. Hope you agree!

PR #69, opened 2020-07-09
Lint .jsx files

jakobw:
I set `"react/prop-types": "off"` because I felt that for our non-critical use case [this rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) was a little too strict.

https://phabricator.wikimedia.org/T257096

### DIFF
--- a/docs/.eslintrc.json
+++ b/docs/.eslintrc.json
@@ -2,7 +2,8 @@
 	"extends": [
 		"wikimedia",
 		"wikimedia/client",
-		"wikimedia/language/es6"
+		"wikimedia/language/es6",
+		"plugin:react/recommended"
 	],
 	"parserOptions": {
 		"sourceType": "module",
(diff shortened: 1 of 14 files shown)


tzhelyazkova:
> I set `"react/prop-types": "off"` because I felt that for our non-critical use case [this rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md) was a little too strict.
> 

makes sense :+1:

docs/.eslintrc.json
```diff
@@ -15,6 +19,7 @@
 			"exports": "always-multiline",
 			"functions": "always-multiline"
 		} ],
-		"func-style": [ "error", "declaration" ]
+		"func-style": [ "error", "declaration" ],
+		"react/prop-types": "off"
```
wiese, line 23: This made me curios. Will look into that a little.

PR #70, opened 2020-07-09
docs: validate our storybook component props via react

wiese:
Ensure the components communicate with each other as desired.

Allows us to remove the linting exception as there are no more
violations.

See https://reactjs.org/docs/typechecking-with-proptypes.html

Bug: [T257096](https://phabricator.wikimedia.org/T257096)

### DIFF
--- a/docs/.eslintrc.json
+++ b/docs/.eslintrc.json
@@ -19,7 +19,6 @@
 			"exports": "always-multiline",
 			"functions": "always-multiline"
 		} ],
-		"func-style": [ "error", "declaration" ],
-		"react/prop-types": "off"
+		"func-style": [ "error", "declaration" ]
 	}
 }
--- a/docs/package.json
+++ b/docs/package.json
@@ -46,6 +46,7 @@
     "eslint-config-wikimedia": "^0.16.2",
     "eslint-plugin-react": "^7.20.3",
     "npm-run-all": "^4.1.5",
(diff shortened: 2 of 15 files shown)


jakobw:
Hm, I wrote in PR #69 

> I set "react/prop-types": "off" because I felt that for our non-critical use case this rule was a little too strict.

and I meant that as "this looks a little tedious for not a lot of value, how about not doing that" rather than "I don't know how to fix it". I'm not strictly against it, though.

tzhelyazkova:
> Hm, I wrote in PR #69
> 
> > I set "react/prop-types": "off" because I felt that for our non-critical use case this rule was a little too strict.
> 
> and I meant that as "this looks a little tedious for not a lot of value, how about not doing that" rather than "I don't know how to fix it". I'm not strictly against it, though.

And I approved and said "makes sense" by which I meant I agree with his decision completely and would have done the same.
Prop types is indeed a bit too much for docs code..

wiese:
I don't agree with "tedious for not a lot of value" (seems to be the official way) or "too much for docs code" (fixing bugs in docs code is just as annoying as fixing bugs in production code and in some way it even _is_ the production code of this product) but I am not attached to it - just wanted to put it out there and learn something in the process.

Isn't the first PR that gets abandoned, but it's weird so see frowny faces for suggestions.

PR #71, opened 2020-07-09
Start storybook with a fixed port

jakobw:
We decided to try out running storybook with a dynamic port at the beginning of this project. While this worked brilliantly for getting things started, I found that allowing the port to change every time `docker-compose up storybook` is run gets annoying quickly.

I'm also happy to abandon this, if y'all feel differently about this.


docker-compose.yml
```diff
@@ -13,4 +13,6 @@ services:
     storybook:
         extends:
             service: node
-        command: 'npm run storybook'
+        ports:
+            - '${STORYBOOK_PORT}:${STORYBOOK_PORT}'
+        command: 'npm run storybook -- -- -- -p ${STORYBOOK_PORT} --ci'
```
jakobw, line 18: If someone who knows how computers work could tell me how to avoid this `-- -- --`, that'd be great.

wiese: This is the `docker-compose.yml` running the root npm script "storybook", i.e.

1. invokes root `npm run`
2. which invokes root `lerna`
3. which invokes docs `npm run`

but we want to pass arguments to the program which is invoked by the 3rd level.

```
npm run storybook -- -- -- -p ${STORYBOOK_PORT} --ci
                        ^--- stops arguments (e.g. -p) from being sent to docs-level npm run
                     ^--- stops arguments (e.g. -p) from being sent to root level lerna
                  ^--- stops arguments (e.g. -p) from being sent to root-level npm run
```

So if one really wanted to reduce the amount of "morse code" needed, pointing to the lower layers would make obsolete the respective amount of dashes. Of course, this somehow renders the root level script pointless.

jakobw: Yup, am aware that's how it works. It just looked funny which made me think there's a different way.

wiese: Sorry for stating the obvious. Thought I heard a question.

wiese:
Realized after the review (of course) that the [last line in the README](https://github.com/wmde/wikit/blame/b5d2a8bff568fc1554843a7e94a4136dd71b1d5e/README.md#L41) is now somewhat obsolete, but there is no harm either.

PR #73, opened 2020-07-10
Introduce TokenSection to produce separate Globals and Alias tables from a token object

jakobw:


jakobw:
I don't feel like clicking "approve" on a PR that I opened, but 7edae56 and the resulting storybook look good to me :+1:

jakobw:
The latest version updates the "Restructure tokens" commit to use the latest version of `tokens/properties/alias.json` from #63.

PR #79, opened 2020-07-14
Use custom kebab case function for generating token names

jakobw:
This overrides style-dictionary's own kebab case function to prevent it
from hyphenating parts of the path that contain numbers. For example,
style-dictionary produces the name `font-weight-style-h-1` from path
`[ 'font', 'weight', 'style', 'h1' ]`, but we want it to be
`font-weight-style-h1`.

I added some extra context to what's happening inside style-dictionary in [a comment on the task](https://phabricator.wikimedia.org/T257805#6303508).

### DIFF
--- a/tokens/.style-dictionary/config.js
+++ b/tokens/.style-dictionary/config.js
@@ -1,12 +1,16 @@
 'use strict';
 
-const { removeWikimediaUiBaseVars, getReferencedTokens } = require( './lib' ),
-	StyleDictionary = require( 'style-dictionary' ).extend( {
+const StyleDictionary = require( 'style-dictionary' ),
+	{ removeWikimediaUiBaseVars, getReferencedTokens, kebabCase } = require( './lib' ),
+	wikitStyleDictionary = StyleDictionary.extend( {
 		include: [ 'node_modules/wikimedia-ui-base/tokens.json' ],
 		source: [ 'properties/**/*.json' ],
 		platforms: {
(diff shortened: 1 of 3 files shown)


SaiSan-WMDE:
The patch has removed all the unwanted dashes from the style tokens' names. Thanks! :)

wiese:
This made me curios - not because of the demand to display it as desired but because it surprised me a lot that the flat token really should be mangled in this unexpected way (in my case, especially the unintuitive impact on e.g. SASS variables, implying a nesting where there is none). Which is also why I did not recheck the way the workaround in this patch works, but only investigated the "why" (is it annoying no one else in style-dictionary)?

Turns out we are not the only ones but even the "Design Systems Architect at Amazon" [recently wrote](https://github.com/amzn/style-dictionary/issues/300#issuecomment-590508455)

> It might be a good idea to take a look at migrating off lodash's string methods [...]

If anything, maybe we could add the ticket to our commit message to have a way to find it again should someone else be surprised by this divergence from style-dictionary defaults in our code base in the future (and maybe the possibility to switch back when it is fixed): https://github.com/amzn/style-dictionary/issues/300

wiese:
Rebased and added the style-dictionary issue to the commit message.

jakobw:
@wiese cool, thanks for finding that ticket. I had only seen the one lodash that I mentioned in the phab comment.

PR #80, opened 2020-07-14
docs: adjust subpage order per specification

wiese:
Also names them as seen in the spreadsheet.

As extrapolated from the feedback in https://phabricator.wikimedia.org/T256854#6301920

Bug: [T256854](https://phabricator.wikimedia.org/T256854)


SaiSan-WMDE:
This is great! 🙌🏻

PR #83, opened 2020-07-16
Add ADR for using TypeScript without Decorators

tzhelyazkova:
[T257813](https://phabricator.wikimedia.org/T257813)

### DIFF
--- a/docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
+++ b/docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
@@ -1,0 +1,37 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+
+<Meta title="Documentation|dev/ADRs/5) Use typescript" />
+
+# 5) Use TypeScript
+
+Date: 2020-07-16
+
+## Status
+
+Accepted
+
+## Context
+The components library that implements the design system will be implemented in Vue (see ADR #1). In addition to that we discussed what benefits we can gain by using [TypeScript](https://www.typescriptlang.org/).
+
+There are several products (as of date) in the WMDE realm that were built with Vue, TypeScript and [Class Component syntax](https://github.com/vuejs/vue-class-component): Termbox, Tainted References, Data Bridge, and we anticipate there will be more. This means the components library needs to support TypeScript apps.
+
+Using TypeScript together with Decorators, when developing Vue projects, used to be the recommended way, because Decorators provide full compatibility between Vue and TypeScript. The TypeScript support in Vue has significantly improved since then.
+
+Benefits of using TypeScript:
+ - It reduces the need for code-level documentation through static typing
+ - It can help with defining interfaces for code that is shared between components
+ - It can help for testing by adding better editor support and type checking
+ - Automatically enables the library to be consumed in TypeScript apps
+
+Downsides of using TypeScript:
+ - Slightly raises the entry barrier for newcomers to Vue
+ - If used with Decorators ([Component](https://github.com/vuejs/vue-class-component), [Prop](https://github.com/kaorun343/vue-property-decorator)) it changes the syntax significantly and the [Vue docs](https://vuejs.org/v2/guide/) become less usable. The Decorators are also not well documented (compared to the [Vue docs](https://vuejs.org/v2/guide/)). The TypeScript support in Vue has improved significantly with time, therefore we might not need the additional help of Decorators to make Vue and TypeScript work well together.
+
+
+## Decision
+Use TypeScript to build the components library (the package vue-components in this monorepo) which implements the design system.
+Don't use [class components](https://vuejs.org/v2/guide/typescript.html#Class-Style-Vue-Components). Use object notation to write Vue components (the default way).
+
+## Consequences
+Using TypeScript will make testing better. It will also make internal reusing of code easier to develop and navigate.
+Using an object notation syntax to develop Vue components (no Decorators) will ensure the ability to easily use the official Vue docs. We don't anticipate any downsides that come from not using Decorators.


docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+# 5) Use TypeScript
+
+Date: 2020-07-16
+
+## Status
+
+Accepted
+
+## Context
+The components library that implements the design system will be implemented in Vue (see ADR #1). In addition to that we disucssied what benefits we can gain by using [TypeScript](https://www.typescriptlang.org/).
```
jakobw, line 14: We did what?! :D
```suggestion
The components library that implements the design system will be implemented in Vue (see ADR #1). In addition to that we discussed what benefits we can gain by using [TypeScript](https://www.typescriptlang.org/).
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+There are several products (as of date) in the WMDE realm that were build with Vue, TypeScript and [Decorators](https://github.com/vuejs/vue-class-component): Termbox, Tainted References, Data Bridge, and we anticipate there will be more. This means the components library needs to support TypeScript apps.
```
jakobw, line 16:
```suggestion
There are several products (as of date) in the WMDE realm that were built with Vue, TypeScript and [Decorators](https://github.com/vuejs/vue-class-component): Termbox, Tainted References, Data Bridge, and we anticipate there will be more. This means the components library needs to support TypeScript apps.
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+Using TypeScript together with Decorators, when developing Vue projects, used to be the recomended way, because Decorators provide full compatibility between Vue and TypeScript. The TypeScript support in Vue has significantly improved since then.
```
jakobw, line 18:
```suggestion
Using TypeScript together with Decorators, when developing Vue projects, used to be the recommended way, because Decorators provide full compatibility between Vue and TypeScript. The TypeScript support in Vue has significantly improved since then.
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+Benefits of using TypeScript:
+ - It can help for internal reuse of behavior (e.g. a toggleable mixin)
```
jakobw, line 21: I think I'd have trouble understanding this point in two weeks once I lost the context of the meeting we had this morning. Trying to phrase it in a more generic way:
```suggestion
 - It can help with defining interfaces for code that is shared between components
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+ - It can help for testing
```
jakobw, line 22: Same here. A lot (all?) of the benefits we get from TS boil down to static typing and whatever benefits that gives us.
```suggestion
 - It can help for testing by adding better editor support and type checking
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+Accepted
+
+## Context
+The components library that implements the design system will be implemented in Vue (see ADR #1). In addition to that we disucssied what benefits we can gain by using [TypeScript](https://www.typescriptlang.org/).
+
+There are several products (as of date) in the WMDE realm that were build with Vue, TypeScript and [Decorators](https://github.com/vuejs/vue-class-component): Termbox, Tainted References, Data Bridge, and we anticipate there will be more. This means the components library needs to support TypeScript apps.
+
+Using TypeScript together with Decorators, when developing Vue projects, used to be the recomended way, because Decorators provide full compatibility between Vue and TypeScript. The TypeScript support in Vue has significantly improved since then.
+
+Benefits of using TypeScript:
```
jakobw, line 20: I'd like to add the argument that we talked about regarding code-level documentation and jsdoc feeling sluggish when you're used to type hints being baked into the code.

Something like: "It reduces the need for code-level documentation through static typing". Does that make sense?

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+ - It can help for internal reuse of behavior (e.g. a toggleable mixin)
+ - It can help for testing
+ - Automatically enables the library to be consumed in TypeScript apps
+
+Downsides of using TypeScript:
+ - Slightly raises the entry barrier for newcommers to Vue
```
jakobw, line 26:
```suggestion
 - Slightly raises the entry barrier for newcomers to Vue
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
```diff
@@ -0,0 +1,36 @@
...
+ - If used with Decorators ([Component](https://github.com/vuejs/vue-class-component), [Prop](https://github.com/kaorun343/vue-property-decorator)) it changes the syntax significantly and the [Vue docs](https://vuejs.org/v2/guide/) become less usable. The Decorators are also not well documented (compared to the [Vue docs](https://vuejs.org/v2/guide/)). The TypeScript support in Vue has improved significantly with time, therefor we might not need the additional help of Decorators to make Vue and TypeScript work well together.
```
jakobw, line 27: I think this explains the downsides of Class Components and its decorators really well!

Typo:
```suggestion
 - If used with Decorators ([Component](https://github.com/vuejs/vue-class-component), [Prop](https://github.com/kaorun343/vue-property-decorator)) it changes the syntax significantly and the [Vue docs](https://vuejs.org/v2/guide/) become less usable. The Decorators are also not well documented (compared to the [Vue docs](https://vuejs.org/v2/guide/)). The TypeScript support in Vue has improved significantly with time, therefore we might not need the additional help of Decorators to make Vue and TypeScript work well together.
```

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
jakobw, line 16: I think we should refer to what's called "Decorators" here as "Class Components" or "Class Component syntax". Decorators are only part of the package, and I'd only use the term when referring to the actual decorators like `@Component`, `@Prop` etc.

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
wiese: Looks like that made it in. :+1:

docs/src/doc/dev/adr/0005-use-typescript.stories.mdx
wiese: Looks like that made it in. +1

PR #84, opened 2020-07-16
Setup vue-components with vue-cli

tzhelyazkova:
## [T257815](https://phabricator.wikimedia.org/T257815) - Set up the package
This is an initial setup done with vue-cli.
All files have been auto-generated. Only `package.json` has been tweaked a bit because it existed before.
All additional lint, test etc. configs will be done as part of follow up tasks.

### Notes

1. I didn't introduce a docker service, because we haven't discussed if we want a dev entry point, or if we are going to use Storybook for preview during development. Either way this could be a separate PR.
2. I left the initial ~~`App.vue` and~~ `HelloWorld.vue` component~~s~~ because we've been talking about having a dummy component. Not sure if this is still valid, now that we have a Button story.


vue-components/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,62 @@
+<template>
```
wiese, line 1: The `logo.png` does not appear to be used anymore after the switch to the library build, so I guess we can remove it. (But I don't know how to comment on a binary file on github)

vue-components/public/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
```
wiese, line 1: This would be the dev entrypoint mentioned in the PR description. I think there is not too much point in having it (the `public` folder) in a library and we'll - as written - try and get storybook to a state where we can look at/develop the individual components.

vue-components/package.json
```diff
@@ -2,25 +2,51 @@
...
-  "license": "BSD-3-Clause",
+  "author": {
+    "name": "The Wikidata team"
+  },
+  "scripts": {
+    "serve": "vue-cli-service serve",
+    "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json",
+    "test:unit": "vue-cli-service test:unit",
+    "lint": "vue-cli-service lint",
+    "test": "echo \"wikit vue components tests\" && exit 0"
```
wiese, line 13: This is again a matter of coordinating this between the patch here and https://phabricator.wikimedia.org/T257817 but if we add the basic test here, I think it should also be run to prove it is functional.

Maybe
```suggestion
    "test": "npm run lint && npm run test:unit"
```
for now.

wiese:
Rebased

PR #86, opened 2020-07-16
docs: nowrap only for token names, not the whole column

jakobw:
A little birdie told me that the [token page for state transitions](https://wmde.github.io/wikit/?path=/docs/design-tokens-transitions-state-transitions--page#transition-interaction-button-focus-property) looked silly after my attempted improvement for avoiding word wraps on token names. I had optimistically slapped `white-space: nowrap;` on the whole table column which looks weird as soon as there are multiple token names.

This change now does what I had intended to do and avoids wrapping for individual token names.

### DIFF
--- a/docs/src/lib/TokenTable.jsx
+++ b/docs/src/lib/TokenTable.jsx
@@ -17,11 +17,17 @@ export function TokenTable( { tokens } ) {
 				{
 					tokens.map( ( token ) => (
 						<tr key={ token.name } id={ token.name }>
-							<td className="token-name">
+							<td>
 								<AnchorMdx href={ '#' + token.name }>🔗</AnchorMdx>
-							&nbsp;<strong>{ token.name }</strong>
+							&nbsp;<strong className="token-name">{ token.name }</strong>
 								<div className="referenced-tokens" title="value influenced by">
-									{ token.referencedTokens || '—' }
+									{token.referencedTokens.length > 0 ?
+										token.referencedTokens.map( ( token, i ) => [
+											i > 0 ? ', ' : '',
+											<span key={i} className="token-name">{token}</span>,
+										] ) :
+										'-'
+									}
 								</div>
 							</td>
 							<td>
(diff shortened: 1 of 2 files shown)


jakobw:
Hola @SaiSan-WMDE, the fix for the token name wrapping issue you spotted before went slightly wrong and looks silly, as seen [here](https://wmde.github.io/wikit/?path=/docs/design-tokens-transitions-state-transitions--page#transition-interaction-button-focus-property). With this improvement it looks like [this](https://5efdb3b5f65950002286285d-pryeodfljj.chromatic.com/?path=/docs/design-tokens-transitions-state-transitions--page).

SaiSan-WMDE:
Hola, @jakobw. It does look better, indeed. Thanks for fixing the fix 😄

wiese:
Splendid. Thanks.

> looked silly

Tell the birdy to watch its phrasing!

_Rebases_

docs/src/lib/TokenTable.jsx
```diff
@@ -17,11 +17,17 @@ export function TokenTable( { tokens } ) {
...
 						<tr key={ token.name } id={ token.name }>
-							<td className="token-name">
+							<td>
 								<AnchorMdx href={ '#' + token.name }>🔗</AnchorMdx>
-							&nbsp;<strong>{ token.name }</strong>
+							&nbsp;<strong className="token-name">{ token.name }</strong>
 								<div className="referenced-tokens" title="value influenced by">
-									{ token.referencedTokens || '—' }
+									{token.referencedTokens.length > 0 ?
+										token.referencedTokens.map( ( token, i ) => [
```
wiese, line 25: Explaining this to myself (correct me if something is not right)...

This
* always yields two entries
* first entry is a text node adding the comma for all but the first (0th) item
* second entry has no surprises
* react, upon seeing this, puts both nodes into the DOM, kind of like a fragment[0]
  * I assume, because the first node is text-only some of the rules don't apply (no key, no shared parent between the two of them)

[0]: https://reactjs.org/docs/fragments.html

PR #89, opened 2020-07-17
Components storybook

jakobw:
This demonstrates that component hot reloading works without additional configuration in storybook as long as we import the components from the `src/` directory. Lerna does the right thing and links these files as expected. Looking at the dependency from the docs package to the vue-components package this is fine, because the `src/` files are part of the package contents.

It does come with a risk of creating a disconnect between the standard usage of components and how we use storybook to develop them, think `import { Component } from '@wmde/wikit-vue-components';` vs `import Component from '@wmde/wikit-vue-components/src/components/Component';`. I think that this is probably fine for now and we can revisit this if it becomes a problem.

Bug: https://phabricator.wikimedia.org/T257825

### DIFF
--- a/docs/src/dummy.stories.js
+++ b/docs/src/dummy.stories.js
@@ -1,4 +1,4 @@
-import HelloWorld from '@wmde/wikit-vue-components/src/components/HelloWorld';
+import { HelloWorld } from '@wmde/wikit-vue-components';
 
 export default {
 	title: 'Components/Dummy',
--- a/docs/src/dummy.stories.js
+++ b/docs/src/dummy.stories.js
@@ -1,7 +1,15 @@
+import HelloWorld from '@wmde/wikit-vue-components/src/components/HelloWorld';
+
 export default {
-	title: 'Components',
+	title: 'Components/Dummy',
 };
 
-export function dummy() {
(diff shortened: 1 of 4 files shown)


jakobw:
@wiese thanks for spotting that.

I now turned it around to not use a src file, but directly use the commonjs build instead. You can start storybook and run `docker-compose run --rm node lerna run dev` in a separate terminal. Uploaded for early feedback, but I want to ponder this a bit more and add some docs.

wiese:
> You can start storybook and run `docker-compose run --rm node lerna run dev` in a separate terminal.

FWIW workflow-wise this sounds very intuitive to me.

On the other hand (this is a mere hunch), things like vue-docgen-loader[0] - on behalf of addon-docs (to show the slots, properties) - may _need_ access to the very source. Fingers crossed.

[0]: https://www.npmjs.com/package/vue-docgen-loader

jakobw:
@wiese Yes, it seems that the addon-docs "prop tables" indeed can't be generated from a component at run time. I checked that this works out of the box with 3a97630 but no longer works with the latest state. What do we do?

jakobw:
With #105 and friends merged, I think this can be closed! :tada:

PR #90, opened 2020-07-17
Storybook addons

jakobw:
This adds the knobs and a11y storybook addons.

### DIFF
--- a/docs/src/dummy.stories.js
+++ b/docs/src/dummy.stories.js
@@ -1,9 +1,10 @@
 import HelloWorld from '@wmde/wikit-vue-components/src/components/HelloWorld';
 import { withKnobs, text } from '@storybook/addon-knobs';
+import { withA11y } from '@storybook/addon-a11y';
 
 export default {
 	title: 'Components/Dummy',
-	decorators: [ withKnobs ],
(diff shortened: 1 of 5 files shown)


docs/src/dummy.stories.js
```diff
@@ -1,15 +1,23 @@
 import HelloWorld from '@wmde/wikit-vue-components/src/components/HelloWorld';
+import { withKnobs, text } from '@storybook/addon-knobs';
+import { withA11y } from '@storybook/addon-a11y';
 
 export default {
 	title: 'Components/Dummy',
+	decorators: [ withKnobs, withA11y ],
```
wiese, line 7: I assume we want this to be true for all (component) stories.

Was the decision not to register those decorators globally, 
[e.g. like was done in bridge](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/Wikibase/+/5f1d7d1/client/data-bridge/.storybook/preview.js), a conscious one? What were the arguments?

wiese:
Trying what it takes to combine this with #105.

wiese:
Replaced by #114 (doing the same thing, but in vue-components instead)

PR #92, opened 2020-07-17
Add badges to README

tzhelyazkova:


README.md
```diff
@@ -1,4 +1,10 @@
-# wikit
+# WiKit
+[![lerna](https://img.shields.io/badge/maintained%20with-lerna-cc00ff.svg)](https://lerna.js.org/)
+[![License](https://img.shields.io/badge/License-BSD%203--Clause-blue.svg)](https://opensource.org/licenses/BSD-3-Clause)
+  
+![github](https://github.com/wmde/wikit/workflows/Build%20and%20Deploy%20documentation/badge.svg)
+![github](https://github.com/wmde/wikit/workflows/Test%20and%20deploy/badge.svg)
```
wiese, line 6: If those are links I guess it makes sense they link somewhere a user could want to go.
What about chromatic & the github page respectively?
Alternatively, we could also show a badge with no link.

README.md
```diff
@@ -1,4 +1,10 @@
-# wikit
+# WiKit
```
wiese, line 1: We decided upon the wikit name already a while ago, but so far largely used it in the repo name, ... 
The way it is written here is probably how it is supposed to be in prose (based on https://github.com/wmde/wikit-css/) but I raised the question in the chat.

Happy to create a follow up to use it consistently (e.g. documentation) if that turns out positive.

wiese: Will do the follow-up once #94 landed.

tzhelyazkova:
Rebased, squashed and merged.

PR #94, opened 2020-07-20
Getting started page should be on top

raja-wmde:
Changed file names according to the new documentation structure that we agreed on.


raja-wmde:
@wiese My part is done now. :) Thanks for updating the folders and reviewing.

wiese:
@raja-wmde Makes sense to me. I did the rename (hopefully) as discussed. Mind checking that I got it right?

wiese:
Rebased and removed the one merge commit

PR #95, opened 2020-07-20
Externalize vue as a dependency

tzhelyazkova:
### Externalize MW-supported packages [T257827](https://phabricator.wikimedia.org/T257827)

I can't think of anything else but Vue for now.

### DIFF
--- a/vue-components/vue.config.js
+++ b/vue-components/vue.config.js
@@ -1,0 +1,13 @@
+const DEV_MODE = process.env.WEBPACK_TARGET === 'dev'
+
+function externals () {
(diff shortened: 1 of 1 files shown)


jakobw:
Vue is externalized for lib builds [by default](https://cli.vuejs.org/guide/build-targets.html#library). I ran into this while working on the storybook PRs. I think we may not need this at all if we go with the default behavior for production builds and `--inline-vue` in dev mode if necessary.

tzhelyazkova:
As suggested, I'll add a few lines about this in "Getting started".

PR #96, opened 2020-07-20
docs: document base CSS decision

wiese:
Bug: https://phabricator.wikimedia.org/T257828

### DIFF
--- a/docs/src/doc/dev/adr/0007-css-base.stories.mdx
+++ b/docs/src/doc/dev/adr/0007-css-base.stories.mdx
@@ -1,0 +1,44 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+
+<Meta title="Documentation|dev/ADRs/7) CSS base" />
+
+# 7) CSS base
+
+Date: 2020-07-20
+
+## Status
+
+Accepted
+
+## Context
+
+We are in the process of implementing a component library. Those components are supposed to become portable manifestations of our Design System decisions which can get integrated into existing applications where needed.
+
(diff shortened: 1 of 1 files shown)


wiese:
This is a bit weak on the question _which_ lib we will use to achieve the desired effect, but I'm not sure we came to a definite verdict.

docs/src/doc/dev/adr/0007-css-base.stories.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+The decision was to employ best practices as seen in other popular open source software, e.g. [vuetify] and [bootstrap-vue], and - as documented there - go with a solution "to correct inconsistencies across browsers and devices while providing slightly more opinionated resets to common HTML elements".
+
+We also decided to trust the task of "mitigating styles inherited from other greedy selectors" to the individual users of our library. _If_ it is needed, then the applications will know best which concrete styles need compensation.
+
+## Consequences
+
+We have to decide for one of the existing technical solutions for this challenge, e.g. [ress], [normalize.css] or [bootstrap reboot].
+
+Component development can happen against a stable base style. We must create a shared "namespace"/selector for our components in order to ensure that we apply the base styles only to the intended namespace under our control.
```
wiese, line 35: This, in my eyes, implies two new sub tasks for the component library
* (hard requirement) employ prefixwrap or comparable - to build a shared namespace for all CSS
* (advisable, but maybe later) employ purgecss or comparable - to eliminate base CSS which has no use in the mark-up of our components

PR #98, opened 2020-07-21
display all width tokens

SaiSan-WMDE:
testing a quick fix


SaiSan-WMDE:
Getting rid of this landfill of failed commits. Will try a more structured approach

PR #100, opened 2020-07-21
optimize size and width tokens

SaiSan-WMDE:


SaiSan-WMDE:
All size dimensions are now captured as globals. Width tokens, like heights and paddings, are now aliases taking those globals as values. The overall logic structure is more consistent.

Storybook seems to be reflecting all changes correctly.

PR #102, opened 2020-07-22
Create a shared style namespace

tzhelyazkova:
### Create a shared CSS "namespace" for our components - [T258481](https://phabricator.wikimedia.org/T258481)

Using [postcss-prefixwrap](https://github.com/dbtedman/postcss-prefixwrap)

### DIFF
--- a/vue-components/postcss.config.js
+++ b/vue-components/postcss.config.js
@@ -1,0 +1,6 @@
+module.exports = {
+	plugins: [
+		require( 'autoprefixer' ),
+		require( 'postcss-prefixwrap' )( '.wikit' ),
+	],
+};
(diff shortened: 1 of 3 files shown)


vue-components/postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+	plugins: [
+		require( 'autoprefixer' ),
```
wiese, line 3: I find it a bit iffy to never directly depend on autoprefixer or postcss-autoprefixer and yet put it to work here, but it is the same way in bridge and termbox and works. Fair.

It also totally makes sense to have it so this patch mentioning is just making explicit what happened implicitly through vue-cli already and adding another plugin next to it.

PR #105, opened 2020-07-23
Dedicated storybook for vue-components & compose it into docs

wiese:
Please see the individual commit messages for the full story.

Bug: https://phabricator.wikimedia.org/T258494


wiese:
The dummy component in the docs storybook now is really obsolete - but we have to find a way to remove it without breaking chromatic (which does not like 0 stories).

wiese:
> Now that the vue components have their own storybook, there is no reason for the doc storybook to support vue.

Agree.

> If we uninstall @storybook/vue then it should be possible to use the <Story> component that comes from the docs-addon in our token .mdx files.

That sounds great, looking forward.

> can be merged and we can try the thing in a follow up. What do you think

Rebasing.

wiese:
Rebased.

Fingers crossed the GH pages deploy works as well as predicted.

PR #106, opened 2020-07-23
Reset styles using ress

jakobw:
Bug: [T258691](https://phabricator.wikimedia.org/T258691)

### DIFF
--- a/vue-components/package-lock.json
+++ b/vue-components/package-lock.json
@@ -22174,12 +22174,6 @@
         }
       }
     },
-    "postcss-prefixwrap": {
-      "version": "1.16.0",
-      "resolved": "https://registry.npmjs.org/postcss-prefixwrap/-/postcss-prefixwrap-1.16.0.tgz",
-      "integrity": "sha512-HQIk1ZYH/8ziVkE6O99aV1ykpRv85onurhpOv5YnbART/s3rSRAAzbwj33tlEQ8wouMJUx2gomKRhmiz5THbmg==",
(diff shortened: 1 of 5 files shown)


jakobw:
@tzhelyazkova @wiese I updated the pull request according to what we discussed. I copied ress manually and modified using the following regex in my IDE via search & replace:
search: `^(  )?(.+?)( \{|,)$`
replace: `$1$2.wikit, .wikit $2$3`
That worked for all but two selectors which I manually modified.

I also removed postcss-prefixwrap so that the built css isn't double-prefixed.

jakobw:
Thanks @wiese, that comment was super helpful. I created #117 (draft) as a possible solution for the removed style scoping. This also does away with the manually copied & modified version of ress. I think that could be a nice solution if I'm not missing anything.

The second problem is annoying, and (I think) stems from the problem that our component library has no single entry point that we could include the style reset in. The sass-loader's `additionalData` obviously gets pulled in per component, which we don't really want. That looks like it should be solvable somehow. Will look into it!

wiese:
> The second problem is annoying

I should have mentioned that I don't see this blocking, as it's clearly not the first time it happens and we so far have always lived with it. But if you have an epiphany it'd of course be a brilliant, starting our fresh component lib without that baggage.

jakobw:
The approach and commit order here changed too often for this PR and its discussion to still make sense, so I'm closing it, pointed #117 at master, and cherry-picked the useful commits into it.

PR #107, opened 2020-07-24
Use css logical properties to support both LTR and RTL languages

tzhelyazkova:
### Decide on an approach to RTL support - [T257829](https://phabricator.wikimedia.org/T257829)
The lib has to support RTL languages.
CSS Logical Properties[0] offers the ability to control the layout through logical, rather than physical, direction and dimension mappings. They are supported in all major browsers but IE11[1]. That's why this patch introduces two postcss plugins - css logical props and the :dir selector[2]

This should be sufficient for devs to write css that supports LTR and RTL languages.

[0] https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties
[1] https://caniuse.com/#search=logical%20properties
[2] https://caniuse.com/#search=%3Adir


wiese:
Rebased (fingers crossed)

vue-components/src/components/HelloWorld.vue
```diff
@@ -38,17 +38,23 @@ export default Vue.extend( {
 
 <style lang="scss">
 .wikit-HelloWorld {
+	inset: logical 0 5px 10px;
+	padding-inline: 20px 40px;
+
 	&__headline {
 		margin: 40px 0 0;
```
wiese, line 45: Following https://github.com/wmde/wikit/pull/112/files#diff-e313bcec67163945b0a78b9059513594R32 I assumed there would be some stylelint rule preventing us from using `margin`/`padding`.
I'm a bit surprised that csstools/use-logical does not do that for us, but https://stylelint.io/user-guide/rules/property-blacklist could arguably be used to implement it on our end (not having to would obviously be superior).

Maybe I just misunderstood what the sentence in the ADR means but looking at the compiled CSS there really are no different LTR/RTL versions for those styles.

vue-components/src/components/HelloWorld.vue
```diff
@@ -38,17 +38,30 @@ export default Vue.extend( {
 
 <style lang="scss">
 .wikit-HelloWorld {
+	inset: logical 0 5px 10px;
+
 	&__headline {
-		margin: 40px 0 0;
+		margin-block-start: 40px;
+		margin-inline-end: 0;
+		margin-block-end: 0;
```
wiese, line 46: **Not really relevant - just for alignment**
```suggestion
		margin-block-start: 40px;
```

As soon we as we have #117 in place this will look a lot less convoluted as the opinionated reset will automatically set the 0 values for us.

vue-components/src/components/HelloWorld.vue
```diff
@@ -38,17 +38,30 @@ export default Vue.extend( {
...
 	}
+
 	&__action-list {
 		list-style-type: none;
-		padding: 0;
+		padding-block-start: 0;
+		padding-inline: 0;
+		padding-block-end: 0;
+		padding-inline-end: 0;
```
wiese, line 54: **Not really relevant - just for alignment**
```suggestion
```

As soon we as we have #117 in place we can omit this altogether as the opinionated reset will automatically set the 0 values for us.

vue-components/src/components/HelloWorld.vue
```diff
@@ -38,17 +38,30 @@ export default Vue.extend( {
...
 	}
+
 	&__action {
 		display: inline-block;
-		margin: 0 10px;
+		margin-block: 0;
+		margin-inline-end: 10px;
+		margin-block-end: 0;
+		margin-inline: 10px;
```
wiese, line 62: From my interpretation a sufficient conversion would have been
```suggestion
		margin-block: 0;
		margin-inline: 10px;
```
(margin-inline does "left" and "right" at once, no need for the margin-inline-end). Do you share this view or am I still misunderstanding something about this new field?

And with #117 it will only be (**Not really relevant - just for alignment**)

```suggestion
		margin-inline: 10px;
```

PR #108, opened 2020-07-24
docs: CSS scoping ADR & apply decision in real life

wiese:
Bug: https://phabricator.wikimedia.org/T258774


docs/src/doc/decisions-and-ADRs/adr/0008-css-scoping.stories.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+We decided to not go with Vue's component-scoped CSS but instead
+* prefix (e.g. through [prefixwrap](https://www.npmjs.com/package/postcss-prefixwrap)) all our selectors with a dedicated class which scopes the style effects to the matching elements
+* apply styling inside of the components through CSS classes unique across components
+
+## Consequences
+
+Components will be created without the `scoped` attribute on the style section of Single File Components.
+
+As this is a component library and there is no shared `App` DOM node but components will be used in isolation, we will need a way (manual or automatic) to ensure that we consistently add the class matching the "prefix" to all component root elements.
```
wiese, line 33: Writing this helped me realize that this may be a problem.

docs/src/doc/decisions-and-ADRs/adr/0008-css-scoping.stories.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+## Status
+
+Accepted
+
+## Context
+
+CSS authors need to take care to not cause undesired global side effects by creating overly greedy selectors (e.g. styling certain tags on the whole page instead of just the desired ones).
+
+Vue offers a built-in solution for this challenge called [component-scoped CSS](https://vuejs.org/v2/guide/comparison.html#Component-Scoped-CSS). It [adds a data attribute](https://vue-loader.vuejs.org/guide/scoped-css.html#scoped-css) to all component's elements and extends the CSS selectors to match them.
```
wiese, line 17: Before reading https://vue-loader.vuejs.org/guide/scoped-css.html#child-component-root-elements I never really thought about the "leak into child components" aspect (only about the way _up_), but I guess with our way of very precise classes this is covered just as well.

docs/src/doc/decisions-and-ADRs/adr/0008-css-scoping.stories.mdx
wiese: Yes it does for the CSS side. But where does the DOM node with the respective class come from to ensure that the CSS selector even matches?

Imaging prefixwrap + ress in action:
CSS:
```
.wikit p { margin-top: 0 }
```

DOM 1:
```
<div class="wikit-FancyComponent">
   <p>something</p>
</div>
```
Nothing resets the top margin of the `<p>`. Could be fixed by adding the `wikit` class to all components' root node (next to the `wikit-FancyComponent` class).

Worse example:


DOM 2:
```
<p class="wikit-FancyComponent">
   something
</p>
```
The root node _is_ actually the `<p>`. Even adding the `wikit` class there, will not make the `.wikit p` selector match.

In our apps, the node with the equivalent of the `wikit` class (e.g. [in bridge](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/Wikibase/+/0de341a/client/data-bridge/src/presentation/App.vue#2)), always comes from the application itself, but for our component library we want to be able to use individual components in an otherwise uncontrolled environment.

Maybe I'm just tired and this will go away over the weekend...

wiese:
The title in 8672f28 will need an update as seen in fb91cf8 once #105 is merged (incl. an update to storybook 6).

wiese:
Oh. Now that we merged #105, the vue-components' storybook need a sass loader in dev, too.

I pointed #110 into this branch as 445ebd2 applies the ADR (in 1e2193f) - so the PR (primarily the first and last commit) only makes sense if we decide to go forward with what is discussed in https://github.com/wmde/wikit/pull/108#discussion_r459982127

wiese:
Fixed the SASS introduction to add the SASS loader to storybook.

docs/src/doc/decisions-and-ADRs/adr/0008-css-scoping.stories.mdx
jakobw: I'm also a little scared about version mismatches between potentially server-side rendered markup and css served by ResourceLoader. That's adventurous enough without scope hashes that are out of our control.

docs/src/doc/decisions-and-ADRs/adr/0008-css-scoping.stories.mdx
jakobw: I agree that this is a problem :/

PR #109, opened 2020-07-24
Install and configure Stylelint

tzhelyazkova:
This PR installs Stylelint to help us lint our style code (css/scss).
The config follows the docs and uses `stylelint-config-standard` with the only change to use indentation of tabs instead 2 spaces (default).
The plugin `stylelint-use-logical` is used to lint for css logical properties, which we introduced in the parent branch of this PR.

### DIFF
--- a/vue-components/package.json
+++ b/vue-components/package.json
@@ -10,8 +10,10 @@
     "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json",
     "test:unit": "vue-cli-service test:unit",
     "test:lint": "vue-cli-service lint --no-fix",
+    "test:lintcss": "stylelint '**/*.vue'",
     "test": "npm-run-all test:*",
     "fix": "vue-cli-service lint",
+    "fix:css": "stylelint '**/*.vue' --fix",
     "build:storybook": "build-storybook",
     "storybook": "start-storybook"
   },
@@ -52,6 +54,9 @@
     "postcss-dir-pseudo-class": "^5.0.0",
     "postcss-logical": "^4.0.2",
     "postcss-prefixwrap": "^1.16.0",
+    "stylelint": "^13.6.1",
+    "stylelint-config-standard": "^20.0.0",
+    "stylelint-use-logical": "^1.1.0",
     "typescript": "~3.9.3",
     "vue-template-compiler": "^2.6.11"
   },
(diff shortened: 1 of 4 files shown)


tzhelyazkova:
> The processor `stylelint-processor-html` allows for stylelint to lint `.vue` files. Unfortunately we lose the `--fix` option.

This is annoying and I can't help but think there must be a solution. Maybe we can address it later, in a separate PR, as this one aims to fulfill a criteria we have in [T257829](https://phabricator.wikimedia.org/T257829)

wiese:
> > The processor `stylelint-processor-html` allows for stylelint to lint `.vue` files

I recently saw this package in tainted reference but could not help the big [deprecation warning](https://www.npmjs.com/package/stylelint-processor-html) - in favor of [@mapbox/stylelint-processor-arbitrary-tags](https://www.npmjs.com/package/@mapbox/stylelint-processor-arbitrary-tags) apparently. Would that work as well?

tzhelyazkova:
> > > The processor `stylelint-processor-html` allows for stylelint to lint `.vue` files
> 
> I recently saw this package in tainted reference but could not help but notice the big [deprecation warning](https://www.npmjs.com/package/stylelint-processor-html) - in favor of [@mapbox/stylelint-processor-arbitrary-tags](https://www.npmjs.com/package/@mapbox/stylelint-processor-arbitrary-tags) apparently. Would that work as well?

Thanks, I didn't notice it's deprecated.
Anyway, it turns out it is not needed after all. I was getting a weird error because of the flag `--syntax scss ` in the stylelint command. Stylelint natively supports vue files.

vue-components/package.json
```diff
@@ -10,8 +10,10 @@
     "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json",
     "test:unit": "vue-cli-service test:unit",
     "test:lint": "vue-cli-service lint --no-fix",
+    "test:lint:css": "stylelint '**/*.vue'",
```
jakobw, line 13: I'm not sure what exactly is going wrong, but it appears that this script does not get picked up by `npm-run-all test:*`. If I remove the second ":" it works fine. Odd.

tzhelyazkova: Yep, npm-run-all does not seem to support the double `:`
I would personally drop it in favor of `&&` but since this is not a concern of this PR, I'll just remove the second `:` for now.

tzhelyazkova:
> Could include [stylelint-config-wikimedia](https://github.com/wikimedia/stylelint-config-wikimedia) but that can also be done in a follow-up.

No objections to that, but this PR aims to solve one thing related to RTL support/css logical properties. So I'd be happy if we add that package separately

PR #111, opened 2020-07-24
docs: add ADR regarding logical CSS directionality

wiese:
Bug: https://phabricator.wikimedia.org/T257829


wiese:
Replaced by #112

PR #114, opened 2020-08-03
vue-components: storybook: update & add addons

wiese:
Add addon-knobs and addon-a11y and configure our dummy story to expose a
knob.


wiese:
(This does not solve the topic of composition knobs yet but will be required either way)

PR #115, opened 2020-08-04
docs: storybook: remove vue-specificity & add addons

wiese:
Works only in conjunction with (i.e. contains) #114; please look at that first.

Gives us usable a11y & knobs in docs which work on the stories from vue-components (see chromatic).

I added a temporary [dummy commit to show the build results](https://github.com/wmde/wikit/runs/943894992#step:7:1) to verify that dc9602e is truly needed (see absence of `stories.json`) - I don't understand why storybook does not automatically do that for us.


docs/package.json
```diff
@@ -28,27 +28,25 @@
...
   },
   "devDependencies": {
     "@babel/core": "^7.10.3",
     "@babel/preset-react": "^7.10.4",
-    "@storybook/addon-docs": "^6.0.0-rc.14",
-    "@storybook/vue": "^6.0.0-rc.14",
+    "@storybook/addon-a11y": "^6.0.0-rc.25",
+    "@storybook/addon-docs": "^6.0.0-rc.25",
+    "@storybook/addon-knobs": "^6.0.0-rc.25",
+    "@storybook/html": "^6.0.0-rc.25",
```
tzhelyazkova, line 42: Why the `html` storybook?

wiese: I did not invest a whole lot of time in getting this to work, but the goal was to get "the most basic" storybook, i.e. without it pulling in framework-specific dependencies which we probably don't want ([c.f.](https://github.com/storybookjs/storybook/issues/3157)).
Clearly, storybook itself is written in react but depending on it here in my eyes would be akin to saying "we want to present components which are implemented in react", which is not the case.

I also tried pulling in only @storybook/core but that did not work.

Happy for all alternative suggestions.

tzhelyazkova: > Clearly, storybook itself is written in react but depending on it here in my eyes would be akin to saying "we want to present components which are implemented in react", which is not the case. 

Technically we are asking it to show React components since we wrote the token presenters using React.

> Happy for all alternative suggestions.

I don't consider this a blocking problem since this implementation works and does its job. I will try @storybook/react on the side and see what explodes. :crossed_fingers:

wiese:
Rebased.

PR #116, opened 2020-08-04
Add class 'wikit' to all component's root nodes

tzhelyazkova:
### Add "wikit" class to all component's root level nodes [T259493](https://phabricator.wikimedia.org/T259493)
Manually add the class `wikit` to each component.
Ensure (by a unit test) that all components contain `wikit` at their root node.
TODO: Add instructions in the docs


vue-components/src/components/HelloWorld.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="wikit-HelloWorld">
+	<div class="wikit wikit-HelloWorld">
```
wiese, line 2:
```suggestion
	<div class="wikit HelloWorld">
```

What do you guys think about dropping the prefix?
I feel with the wikit class present at some point in the tree (and in the component library that tree should not be too deep), the component name should be descriptive enough to give people inspecting the DOM an idea where to start looking for the implementation (which in my eyes was the reason for adding the prefix).

OTOH with a strong prefix we are also shielding us from accidentally inheriting from someone trying to style all `.Button`s...

vue-components/tests/unit/components.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { mount } from '@vue/test-utils';
+import * as allComponents from '@/main';
+
+describe( 'All components', () => {
+	it( 'have the class wikit in their root node', () => {
+		Object.keys( allComponents ).forEach( ( component ) => {
```
wiese, line 6:
```suggestion
		Object.values( allComponents ).forEach( ( component: VueConstructor<Vue> ) => {
```
(with `import Vue, { VueConstructor } from 'vue';`)

would be slightly nicer in my eyes, but it's no big deal.

It's not clear to me why `Object.values()` is frowned upon though (es/no-object-values).

vue-components/src/components/HelloWorld.vue
tzhelyazkova: > OTOH with a strong prefix we are also shielding us from accidentally inheriting from someone trying to style all .Buttons...

Yes, I think that's a valid concern.
I'm in favor of keeping it, even though it looks a bit strange to have `wikit wikit-..`

PR #117, opened 2020-08-04
Style scoping and style reset

jakobw:
* removes the `postcss-prefixwrap` scoping which didn't work as intended due to the lack of a shared root node
* adds the custom postcss plugin which limits the selector scope to `.wikit` elements and elements within `.wikit` elements, effectively applying `.wikit ${selector}, ${selector}.wikit` everywhere, with adjustments for pseudo and attribute selectors
* adds ress via sass-loader's `additionalData`

### DIFF
--- a/vue-components/tests/unit/build/postcssWikitScope.spec.ts
+++ b/vue-components/tests/unit/build/postcssWikitScope.spec.ts
@@ -1,0 +1,40 @@
+const postcss = require( 'postcss' );
+const wikitScope = require( '../../../build/postcssWikitScope' );
+
+async function processWithPlugin( input: string ): Promise<{ css: string }> {
+	return postcss( [ wikitScope ] ).process( input, { from: undefined } );
+}
+
+describe( 'postcssWikitScope', () => {
+
+	it.each( [
+		[
+			'.foo {}',
+			'.wikit .foo, .foo.wikit {}',
+		],
+		[
+			'::before {}',
+			'.wikit ::before, .wikit::before {}',
+		],
+		[
+			'[type="button"] {}',
+			'.wikit [type="button"], .wikit[type="button"] {}',
+		],
+		[
+			'div [type=button] {}',
+			'.wikit div [type=button], div.wikit [type=button] {}',
+		],
+		[
+			'.foo a:hover {}',
+			'.wikit .foo a:hover, .foo.wikit a:hover {}',
+		],
+		[
+			'a:hover {}',
+			'.wikit a:hover, a.wikit:hover {}',
+		],
+	] )( 'applies .wikit scoping', async ( input, expected ) => {
+		const { css } = await processWithPlugin( input );
+		expect( css ).toEqual( expected );
+	} );
+
+} );
--- a/vue-components/build/postcssWikitScope.js
+++ b/vue-components/build/postcssWikitScope.js
@@ -1,0 +1,35 @@
+const postcss = require( 'postcss' );
+const selectorParser = require( 'postcss-selector-parser' );
+
+// copied & modified from https://github.com/pazams/postcss-scopify/blob/master/index.js#L86
+function isRuleScopable( rule ) {
+	if ( rule.parent.type !== 'root' ) {
+		return rule.parent.type === 'atrule' && [ 'media', 'supports', 'document' ].includes( rule.parent.name );
+	} else {
+		return true;
+	}
+}
+
+const addWikitClassToFirstSelectorNode = selectorParser( ( selectors ) => {
+	const selector = selectors.first; // there is always only one selector passed at a time from where it's called
+	const wikitClassNode = selectorParser.className( { value: 'wikit' } );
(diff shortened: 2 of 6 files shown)


wiese (commented):
This appears to be shockingly effective at minimal code.
It means implementing our own build software (orange flag) but as we already discussed it feels like this should be a thing as it's the one place where all the styles have to go through.
Would be great to sanity check this in integration with the rest of the moving pieces (which I have not done yet), but I like it.


A tiny(?) bug

`:after.wikit`,
`:before.wikit`
are not possible AFAIK to the point that they are invalid (while `.wikit :before` is).
I guess a valid equivalent of e.g. `:after.wikit` would be `.wikit:after` (similar like in "the other direction" but without the space).
Should be true for any kind of pseudo element?!


A possible performance (CSS size) gain

`.wikit-HelloWorld__link`
gets turned into
`.wikit .wikit-HelloWorld__link` and
`.wikit-HelloWorld__link.wikit`
While this makes sense for the reset styles, it is redundant for component styles.
Component root nodes would want the second version, child nodes would want the first version.
Looking at the rules I did not find a way of telling those situations apart in the (compiled) CSS sense though.
If we were to strictly follow BEM we could use the absence of "__" as a toggle but that sounds brittle.
On the bright side, gzip is going to eat this away like nothing.


A possible micro optimization

`.wikit html`
will probably not happen IRL

jakobw:
> This appears to be shockingly effective at minimal code.
> It means implementing our own build software (orange flag) but as we already discussed it feels like this should be a thing as it's the one place where all the styles have to go through.
> Would be great to sanity check this in integration with the rest of the moving pieces (which I have not done yet), but I like it.
> 
> A tiny(?) bug
> 
> `:after.wikit`,
> `:before.wikit`
> are not possible AFAIK to the point that they are invalid (while `.wikit :before` is).
> I guess a valid equivalent of e.g. `:after.wikit` would be `.wikit:after` (similar like in "the other direction" but without the space).
> Should be true for any kind of pseudo element?!

Thanks for spotting that. Took me a while to get this sorted out with the help of `postcss-selector-parser` but I think the result works. I applied the same for attribute selectors. Not sure if `[attribute].class` is technically invalid, but the other way around is definitely more common.

> 
> A possible performance (CSS size) gain
> 
> `.wikit-HelloWorld__link`
> gets turned into
> `.wikit .wikit-HelloWorld__link` and
> `.wikit-HelloWorld__link.wikit`
> While this makes sense for the reset styles, it is redundant for component styles.
> Component root nodes would want the second version, child nodes would want the first version.
> Looking at the rules I did not find a way of telling those situations aparat in the (compiled) CSS sense though.
> If we were to strictly follow BEM we could use the absence of "__" as a toggle but that sounds brittle.
> On the bright side, gzip is going to eat this away like nothing.

Yup, I noticed that. Using "__" as an indicator could work. I think that could be a follow-up, though.

> 
> A possible micro optimization
> 
> `.wikit html`
> will probably not happen IRL

Yeah, the `html` and `body` resets are unnecessary to begin with for the purpose of our component library, but adding extra noise to the selectors definitely doesn't help. I think eliminating them in an even earlier step of the build pipeline might be easier than attempting to handle them here.

vue-components/tests/unit/build/postcssWikitScope.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+			'::before {}',
+			'.wikit ::before, .wikit::before {}',
+		],
+		[
+			'[type="button"] {}',
+			'.wikit [type="button"], .wikit[type="button"] {}',
+		],
+		[
+			'div [type=button] {}',
+			'.wikit div [type=button], div .wikit[type=button] {}',
```
wiese, line 25:
```suggestion
			'.wikit div [type=button], div.wikit [type=button] {}',
```

But let's discuss this first, maybe I'm confused.

Isn't it the case that we are expecting (by virtue of #116) the wikit class to be placed on the "root" node of the DOM?

So for the second selector (after the comma) we would not be looking for "wikit nodes which are a button residing inside of some div" (`div .wikit[type=button]`), but instead "button nodes residing in a div which is wikit" (`div.wikit [type=button]`).

Can this be generalized that for all "deep" selectors (with a space), only the part before the first space needs to be either scoped (`.wikit something`) or being the scope itself (`something.wikit`)?

vue-components/tests/unit/build/postcssWikitScope.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+		],
+		[
+			'.foo a:hover {}',
+			'.wikit .foo a:hover, .foo a.wikit:hover {}',
```
wiese, line 29:
```suggestion
			'.wikit .foo a:hover, .foo.wikit a:hover {}',
```

ditto

vue-components/tests/unit/build/postcssWikitScope.spec.ts
jakobw: Nope, turns out I was confused. Adjusted the tests, and changed the code. Thanks!

PR #118, opened 2020-08-05
docs: storybook: mdx only

wiese:
It's surprising to me that `--exit-once-uploaded` in 091858b does not work out, given [its description](https://www.chromatic.com/docs/cli#chromatic-options) but that does seem to be subject to interpretation.

Feel free to edit.

Bug: https://phabricator.wikimedia.org/T259694


wiese:
This is a dated attempt at a - now - solved problem.

PR #120, opened 2020-08-06
vue-components: implement neutral flavor of Button

wiese:
Please review/merge #122 first.

Bug: https://phabricator.wikimedia.org/T256779

### DIFF
--- a/vue-components/src/components/Button.vue
+++ b/vue-components/src/components/Button.vue
@@ -29,20 +29,6 @@ export default Vue.extend( {
 </script>
 
 <style lang="scss">
-$wikit-Button-medium-padding-vertical: $dimension-padding-vertical-medium;
-$wikit-Button-medium-padding-horizontal: $dimension-padding-horizontal-medium;
-
-$wikit-Button-large-padding-vertical: $dimension-padding-vertical-large;
-$wikit-Button-large-padding-horizontal: $dimension-padding-horizontal-large;
-
-$wikit-Button-font-family: $font-family-style-label;
-$wikit-Button-font-weight: $font-weight-bold;
-$wikit-Button-font-size: $font-size-style-label; // unit? px vs em vs rem
-$wikit-Button-line-height: $font-line-height-style-label;
-$wikit-Button-border-width: $border-width-thin;
-$wikit-Button-border-style: $border-style-base;
-$wikit-Button-border-radius: $border-radius-base;
-
 /**
  * Construction site identified:
  * How do we name the component level token?
@@ -50,15 +36,13 @@ $wikit-Button-border-radius: $border-radius-base;
  * Suggestion for naming scheme:
  * wikit-<Component>-[prop]-[state]-css-property
  */
-$wikit-Button-default-cursor: $cursor-pointer;
-$wikit-Button-disabled-cursor: $cursor-default;
 
 .wikit-Button {
 	padding-inline: $wikit-Button-medium-padding-horizontal;
 	padding-block: $wikit-Button-medium-padding-vertical;
 	font-family: $wikit-Button-font-family;
-	font-weight: $font-weight-bold;
-	font-size: $font-size-style-label;
+	font-weight: $wikit-Button-font-weight;
+	font-size: $wikit-Button-font-size; // unit? px vs em vs rem
 	line-height: $wikit-Button-line-height;
 	cursor: $wikit-Button-default-cursor;
 	border-width: $wikit-Button-border-width;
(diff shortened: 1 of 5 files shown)


wiese:
Moved the first 4 (infrastructure) commits into #122

vue-components/src/components/Button.vue
```diff
@@ -0,0 +1,95 @@
...
+ * Construction site identified:
+ * How do we name the component level token?
+ *
+ * Suggestion for naming scheme:
+ * wikit-<Component>-[prop]-[state]-css-property
+ */
+
+.wikit-Button {
+	padding-inline: $wikit-Button-medium-padding-horizontal;
+	padding-block: $wikit-Button-medium-padding-horizontal;
```
wiese, line 42: We were really struggling with telling apart horizontal and vertical...

vue-components/src/components/Button.vue
```diff
@@ -0,0 +1,95 @@
...
+} );
+</script>
+
+<style lang="scss">
+/**
+ * Construction site identified:
+ * How do we name the component level token?
+ *
+ * Suggestion for naming scheme:
+ * wikit-<Component>-[prop]-[state]-css-property
```
wiese, line 37: This should be documented somewhere. ADR is too big but "Developing Components" in the documentation would lend itself.

wiese:
This is now in a state where it can be reviewed, I think.
I would not call any of the flavors finished at this point (not even 'neutral' because of the [size question](https://github.com/wmde/wikit/pull/120/files#diff-b84cff79179fa454f6235575745f085eR45)) but by moving it into the main line we allow for the documented challenges to be worked on in parallel without keeping a long-running feature branch.

vue-components/src/components/Button.vue
SaiSan-WMDE: Interesting! did you agree on a more familiar name? These tokens were previously named "-padding-x & "-padding-y" respectively, but I found that could be confusing too

wiese: My comment was strictly about us doing a mistake when mapping both "inline" (x) and "block" (y) to "horizontal" (instead of "vertical" for "block").

As the tokens we are seeing are component-level tokens and created within the same commit they were never called "x" or "y". Maybe they were called like this in the aliases or globals but this is not what this is about - but both is fine in my eyes.

SaiSan-WMDE:
Hey! It's so cool to see our first component in Storybook 🎉  

I noticed that the button is apparently not displaying its active state. It becomes "focused" instead of active on and while holding click. (I have to say that the button created during the prototyping week was afflicted by the same issue – which wasn't mended due to time constrains.)

jakobw:
> Hey! It's so cool to see our first component in Storybook 
> 
> I noticed that the button is apparently not displaying its active state. It becomes "focused" instead of active on and while holding click. (I have to say that the button created during the prototyping week was afflicted by the same issue – which wasn't mended due to time constraints.)

@SaiSan-WMDE well spotted! It turns out that a clicked button is both `:active` and `:focus`ed, and in our case the latter overrode the former. I think I managed to solve it in the [latest version](https://5efdb3b5f65950002286285d-uvzsstllqr.chromatic.com/?path=/story/vue_button--normal).

SaiSan-WMDE:
Nice! Active state is on and the button is now behaving correctly. Thanks, @jakobw 

I noticed, though that in this last version, the vertical padding seems to be taking the value of the horizontal padding (it was fine before) 👀

jakobw:
@SaiSan-WMDE ugh, sorry about that. Fixed in the [latest version](https://5efdb3b5f65950002286285d-aapgqzxusq.chromatic.com/?path=/story/vue_button--normal).

SaiSan-WMDE:
Hey @jakobw, thanks for the last fix! It's hard for me to be entirely sure (not even after checking the code), but it appears that no transitions (duration) have been applied "between" states, right?

jakobw:
@SaiSan-WMDE Added the transitions!

SaiSan-WMDE:
I run out of itty bitty corrections to bring up! All good 👍

vue-components/src/components/Button.vue
```diff
@@ -0,0 +1,123 @@
...
+ */
+$wikit-Button-default-cursor: $cursor-pointer;
+$wikit-Button-disabled-cursor: $cursor-default;
+
+.wikit-Button {
+	padding-inline: $wikit-Button-medium-padding-horizontal;
+	padding-block: $wikit-Button-medium-padding-vertical;
+	font-family: $wikit-Button-font-family;
+	font-weight: $font-weight-bold;
+	font-size: $font-size-style-label;
```
tzhelyazkova: If we have the component level tokens above ($wikit-Button-font-weight and $wikit-Button-font-size) why not use them?

vue-components/src/components/Button.vue
```diff
@@ -37,7 +37,9 @@ export default Vue.extend( {
  * wikit-<Component>-[prop]-[state]-css-property
  */
 
-.wikit-Button {
+$base: ".wikit-Button";
+
+#{$base} {
```
tzhelyazkova, line 42: :cool:

vue-components/src/components/Button.vue
jakobw: Whoops, that was an accident. Fixed!

PR #121, opened 2020-08-06
Install Nightwatch for e2e tests

tzhelyazkova:
### Add browser testing [T257818](https://phabricator.wikimedia.org/T257818)
[vue-cli](cli.vuejs.org) offers two plugins for e2e tests: [vue-cli-plugin-e2e-cypress](https://cli.vuejs.org/core-plugins/e2e-cypress.html#vue-cli-plugin-e2e-cypress) and [vue-cli-plugin-e2e-nightwatch](https://cli.vuejs.org/core-plugins/e2e-nightwatch.html#vue-cli-plugin-e2e-nightwatch). They recommend the nightwatch one for cross browser needs.
[Nightwatch](https://nightwatchjs.org/) is based on Selenium and works with Saucelabs.

TODO:
- [x] Extend [vue-cli's nightwatch config](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-e2e-nightwatch/nightwatch.config.js) to work with saucelabs
- [x] Run the e2e tests in GitHub Actions
- [ ] Run the e2e tests in Docker
- [x] Make `launch_url` dynamic. For local runs it should be `localhost:STORYBOOK_PORT`, and for CI runs it should be the choromatic url generated on the branch.


vue-components/nightwatch.config.js
```diff
@@ -1,8 +1,65 @@
...
 	test_settings: {
 		default: {
-			launch_url: 'http://localhost:6005/',
+			launch_url: 'https://wmde.github.io/wikit/vue-components/',
+		},
+		sauceLabs: {
+			selenium_host: `ondemand.${process.env.REGION === 'eu' ? 'eu-central-1.' : ''}saucelabs.com`,
+			selenium_port: 80,
+			username: process.env.SAUCE_USERNAME,
+			access_key: process.env.SAUCE_ACCESS_KEY,
```
tzhelyazkova, line 22: In order to run the tests on saucelabs one has to configure a username and access key locally as described in
https://wiki.saucelabs.com/display/DOCS/Best+Practice%3A+Use+Environment+Variables+for+Authentication+Credentials
The access key and username can be found at https://app.saucelabs.com/user-settings

vue-components/nightwatch.config.js
```diff
@@ -1,8 +1,65 @@
...
+
+	test_workers: {
+		enabled: true,
+		workers: 'auto',
+	},
+
 	test_settings: {
 		default: {
-			launch_url: 'http://localhost:6005/',
+			launch_url: 'https://wmde.github.io/wikit/vue-components/',
```
tzhelyazkova, line 16: Temp hard-coded to the production storybook for testing purposes.
In the final solution this will ideally be dynamic and take the URL chromatic generates per branch.

tzhelyazkova:
Currently the e2e tests run against the storybook on github pages.
The next step will be to change the `launch_url` based on the environment - local or CI. This will be addressed in a separate PR, otherwise this one will become too big and hard to review.

.github/workflows/test-and-demo.yaml
```diff
@@ -27,3 +27,13 @@ jobs:
       run: npm run build:storybook
     - name: Deploy Storybook
       run: npm run chromatic
+    - name: Extract branch name
```
jakobw, line 30: Strange that this is necessary. Apparently github provides the branch name for actions that run on pull requests, but not on push.

package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "postinstall": "lerna bootstrap",
     "test": "lerna run test",
+    "test:e2e": "lerna run e2e:saucelabs --scope @wmde/wikit-vue-components",
```
jakobw, line 12: What made you go with the e2e label instead of browser test? In most code bases I know e2e tests are the kind that send http requests to the application webserver and inspect the response, but without a browser.

vue-components/nightwatch.config.js
```diff
@@ -0,0 +1,68 @@
...
+		sauceIE: {
+			extends: 'sauceLabs',
+			desiredCapabilities: {
+				browserName: 'internet explorer',
+				platform: 'Windows 10',
+				version: 'latest',
+			},
+		},
+
+		sauceEdge: {
```
jakobw, line 59: tee-hee, `sauceEdge` :hotdog:

package.json
tzhelyazkova: The problem I encountered was that in CI we run test:* before the storybook commands, but for the cross browser tests we need a built and deployed storybook, so the e2e needs to run last. 
Removing `test:` was a quick solution. I suppose we can have it as `test:e2e` but then we need to change the lerna test to run all but the e2e.

.github/workflows/test-and-demo.yaml
tzhelyazkova: Yep, I learned this the hard way :x

vue-components/nightwatch.config.js
tzhelyazkova: lol haven't noticed that :laughing:

tzhelyazkova:
> It looks like we don't need any of the files in `tests/e2e/custom-assertions/` and `tests/e2e/custom-commands/`. I think we should get rid of those to reduce some noise.

Done :heavy_check_mark:

package.json
jakobw: Ah, I wasn't asking about `e2e` vs `test:e2e`, but rather why `e2e`/`test:e2e` and not `test:browser`? In termbox for example `tests/edge-to-edge` contains non-browser e2e tests that make http requests against the server and expect certain responses, whereas ~~`tests/browser`~~ `tests/selenium` contain the actual browser tests. I think I saw a similar structure in a FUN (?) code base as well.

Not a big deal, though. It can be renamed later, if anyone trips over it.

tzhelyazkova: Oooh, I didn't actively name it `e2e`. It came from the [@vue/cli-plugin-e2e-nightwatch](https://www.npmjs.com/package/@vue/cli-plugin-e2e-nightwatch) plugin.

PR #122, opened 2020-08-07
vue-components: prerequisites

wiese:
Adds some more prerequisites for the development of components and their tests.

Bug: https://phabricator.wikimedia.org/T257812

### DIFF
--- a/vue-components/src/styles/main.scss
+++ b/vue-components/src/styles/main.scss
@@ -1,0 +1,2 @@
+@import '~ress';
+@import '~@wmde/wikit-tokens/dist/variables';
--- a/vue-components/.eslintrc.js
+++ b/vue-components/.eslintrc.js
@@ -1,7 +1,6 @@
 module.exports = {
 	extends: [
 		'wikimedia',
-		'wikimedia/client',
 		'wikimedia/language/rules-es2017', // the not-* parts are obsolete after transpiling and polyfills
 		'plugin:vue/strongly-recommended',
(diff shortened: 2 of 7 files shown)


vue-components/src/styles/main.scss
```diff
@@ -0,0 +1,2 @@
+@import '~ress';
+@import '~@wmde/wikit-tokens/dist/variables';
```
jakobw, line 2: I'm expecting this to change in the future. While the variables are supposed to be included in every components' scss, we actually only want ress to be included once. Good for now, though.

wiese: You mentioned this in person already but I'm still not sure I understand.
ress is de-duplicated even if there are multiple components (if that is the concern), so is this just about the coarse way the resulting CSS is ordered/commented which I mentioned in https://github.com/wmde/wikit/pull/106#pullrequestreview-460672832?

jakobw: > ress is de-duplicated even if there are multiple components
Its CSS itself is de-duplicated because of some smart postcss plugin in our build step is doing a good job at it. The license comment will be included in the built CSS once for each component, which is silly. It's true that aside from the comment there isn't a lot of practical differences but to me it would make sense for ress to be included only once, because it globally resets styles for all components. The SCSS variables in turn are needed by each component's SCSS code and have no effect on the build output, so it makes sense for them to be loaded once per component.

This changes a little depending on how we actually divide/ship our CSS - per component, or bundled for the whole library?

PR #123, opened 2020-08-07
vue-components: remove HelloWorld dummy component

wiese:
Based on #120

The now existing Button component uses all features which HelloWorld
implemented as a placeholder so far - it has done its job.

Bug: https://phabricator.wikimedia.org/T256779


jakobw:
Not merging yet to avoid adding noise to #120

vue-components/src/main.ts
```diff
@@ -1,7 +1,5 @@
-import HelloWorld from './components/HelloWorld.vue';
 import Button from './components/Button.vue';
 
 export {
 	Button,
-	HelloWorld,
```
tzhelyazkova, line 4: The dummy browser tests in #121 use `<HelloWorld>`.
We should think about which PR goes when into master so we don't create a big mess of failing CI etc :smile:

jakobw: I put #127 in front of it so this is no longer a problem =)

PR #124, opened 2020-08-07
vue-components: some infrastructure improvements

wiese:
Based on #123


vue-components/.storybook/.eslintrc.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
+	overrides: [
+		{
+			files: [ 'preview.js' ],
```
jakobw, line 4: I'm again a little confused by the blurred lines between run time and build time in storybook. If I'm not mistaken, `preview.(j|t)s` is actually executed at run time, so I'd rather remove this `.eslintrc.js` entirely, and just rename `preview.js` -> `preview.ts` and not bother specifying additional rules for run time `.js` files.

PR #128, opened 2020-08-11
vue-components: Add primaryProgressive Button flavor

tzhelyazkova:
### Create Button Component [T256779](https://phabricator.wikimedia.org/T256779)

This PR adds the primary progressive flavor to the Button component.

### DIFF
--- a/tokens/properties/components/Button.json
+++ b/tokens/properties/components/Button.json
@@ -132,17 +132,6 @@
                 "transition-duration": {
                     "value": "{transition.interaction.button.focus.duration.value}"
                 }
-            },
-            "disabled": {
-                "color": {
-                    "value": "{font.color.inverted.value}"
-                },
-                "border-color": {
-                    "value": "{border.color.base.disabled.value}"
-                },
-                "background-color": {
-                    "value": "{background.color.base.disabled.value}"
-                }
             }
         },
         "active": {
@@ -181,6 +170,15 @@
         "disabled": {
             "cursor": {
                 "value": "{cursor.default.value}"
+            },
+            "color": {
+                "value": "{font.color.inverted.value}"
+            },
+            "border-color": {
+                "value": "{border.color.base.disabled.value}"
+            },
+            "background-color": {
+                "value": "{background.color.base.disabled.value}"
             }
         }
     }
--- a/tokens/properties/components/Button.json
+++ b/tokens/properties/components/Button.json
@@ -87,6 +87,64 @@
                 }
             }
         },
+        "progressive": {
+            "color": {
+                "value": "{font.color.inverted.value}"
+            },
+            "border-color": {
+                "value": "{border.color.progressive.default.value}"
+            },
+            "background-color": {
+                "value": "{background.color.progressive.default.value}"
+            },
+            "hover": {
+                "border-color": {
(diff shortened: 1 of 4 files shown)


tzhelyazkova:
@SaiSan-WMDE please have a look at the progressive button story in this branch's Storybook.

SaiSan-WMDE:
Glad to see that selector working! 🎉 Some comments:

On the flavor (type) selector:

 - I believe the word "Type" is probably more conventional than Flavor and would therefore recommend using the former.
- The type name should be simple "progressive" rather than primaryProgressive

On the component itself:

- The focus state of the button should display the following bg color by defaul: background-color-progressive-focus, #36; instead, it seems to be displaying the hover bg color persistently.

- Is it possible that the button transition is still rough on blur? It should mirror the hover transition: I'm thinking that either this is because of a missing timing function or duration spec. Sorry for bringing this up again, it was easier to perceive this thanks to the colored bg. Happy to clarify this last point!

tzhelyazkova:
> On the flavor (type) selector:
> * I believe the word "Type" is probably more conventional than Flavor and would therefore recommend using the former.
> * The name of this type should be simply "progressive" rather than primaryProgressive

Done :heavy_check_mark: 

> On the component itself:
>  * The focus state of the button should display the following bg color: background-color-progressive-focus, #36; instead, it seems to be displaying the hover bg color persistently.

Done :heavy_check_mark: 

> * Is it possible that the button transition is still rough on blur? It should mirror the hover transition: I'm thinking that either this is because of a missing timing function or duration spec. Sorry for bringing this up again, it was easier to perceive this thanks to the colored bg. Happy to clarify this last point!

Hmm, I don't quite understand what exactly is wrong. All tokens from the Figma spec have been defined in the component.
Which transition is wrong :hover, :active or :focus?

SaiSan-WMDE:
Since we decided to tackle the "unhover" transition issue in the future, I'd say this button type is good to go from a UX standpoint.

tokens/properties/components/Button.json
```diff
@@ -87,6 +87,64 @@
...
+                    "value": "{transition.timing-function.ease.value}"
+                },
+                "transition-property": {
+                    "value": "{transition.interaction.button-primary.focus.property.value}"
+                },
+                "transition-duration": {
+                    "value": "{transition.interaction.button.focus.duration.value}"
+                }
+            },
+            "disabled": {
```
jakobw, line 136: As far as I remember the disabled state is meant to look the same for all button ~~flavors~~ types, so its component-level tokens should not be nested within the type, but separately a level under `wikit-Button`. This is already reflected in the `Button.vue` css in that `&:disabled` is separated from the type selectors. It just wasn't fully implemented yet.

tzhelyazkova: There are specific disabled styles for the progressive button defined in Figma.
@SaiSan-WMDE are the button types supposed to look the same when they are disabled?

tzhelyazkova:
rebased

tokens/properties/components/Button.json
SaiSan-WMDE: Yes, all three button types look exactly the same when disabled (same applies to other button variants like quiet). I'm sorry to have caused this misunderstanding through the specs.

PR #134, opened 2020-08-12
Add missing transition durations

SaiSan-WMDE:


SaiSan-WMDE:
We decided to stick to specifying the transition duration only once in order to avoid redundancy.

PR #140, opened 2020-08-13
Apply generic button transitions instead of state-specific ones

jakobw:
- [x] adds generic tokens for button transition properties and duration
- [x] applies the new generic tokens instead of the state-specific ones
- [x] removes the (now unused) state-specific button transition tokens

### DIFF
--- a/tokens/properties/alias.json
+++ b/tokens/properties/alias.json
@@ -438,45 +438,6 @@
                 },
                 "duration": {
                     "value": "{transition.duration.fast.value}"
-                },
-                "hover": {
-                    "property": {
-                        "value": "{transition.property.background-color.value}, {transition.property.border-color.value}"
-                    },
-                    "duration": {
-                        "value": "{transition.duration.fast.value}"
-                    }
-                },
-                "active": {
-                    "property": {
-                        "value": "{transition.property.background-color.value}, {transition.property.border-color.value}"
-                    },
-                    "duration": {
-                        "value": "{transition.duration.fast.value}"
-                    }
-                },
-                "focus": {
-                    "property": {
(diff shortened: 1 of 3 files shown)


jakobw:
@SaiSan-WMDE it would be brilliant if you could look at the two commits that change the tokens (ded2a270eff2071049c8f839462a5bc327e83693 & 76b0d42b370b64f7e4f0359ed3b3b080a0992b40) and verify that it does what it's supposed to [here](https://5efdb3b5f65950002286285d-fmfaqqugeu.chromatic.com/).

tokens/properties/alias.json
```diff
@@ -433,44 +433,11 @@
...
             "button": {
-                "hover": {
-                    "property": {
-                        "value": "{transition.property.background-color.value}, {transition.property.border-color.value}"
-                    },
-                    "duration": {
-                        "value": "{transition.duration.fast.value}"
-                    }
+                "property": {
+                    "value": "{transition.property.background-color.value}, {transition.property.background-color.value}, {transition.property.border-color.value}, {transition.property.box-shadow.value}"
```
SaiSan-WMDE, line 437: The text color of neutral buttons is lightened on hover (see deleted line 461), so the token {transition.property.color.value} should also be added here. 
```suggestion
                    "value": "{transition.property.background-color.value}, {transition.property.border-color.value}, {transition.property.box-shadow.value}, {transition.property.color.value} "
```

jakobw: Yup, the duplicated `background-color` was totally my failed attempt at adding the `color` property. I blame the heat. Thanks for spotting that! Fixed!

tzhelyazkova:
rebased

PR #141, opened 2020-08-13
Move Chromatic to run on the vue-components package

tzhelyazkova:
### Explore how component changes affect chromatic [T257826](https://phabricator.wikimedia.org/T257826) 

After enabling the UI Review and Visual Tests options on Chromatic, it became clear that they don't work on our composed Storybook, whose main stories are the design system tokens.
After discussing this with UX, we agreed that it's more important to leverage the features Chromatic offers on our vue components and not on our design tokens.

This PR moves chromatic from the `docs` package to the `vue-components` package.
:heavy_plus_sign: This means the per-branch generated Chromatic URLs will show only the vue components Storybook. It will enable us to run visual tests and use the UI Review feature.
:heavy_plus_sign: This also means the browser tests no longer need to make a difference between a composed storybook and a local vue components storybook.
:heavy_minus_sign: The downside is that any token changes cannot be previewed on Chromatic anymore. This can only happen locally.

### DIFF
--- a/vue-components/tests/e2e/custom-commands/openComponentStory.js
+++ b/vue-components/tests/e2e/custom-commands/openComponentStory.js
@@ -3,13 +3,9 @@ module.exports.command = function openComponentStory( component ) {
 		.click( `a[href="#${component}"]` )
 		.element(
 			'id',
-			'storybook-ref-vue',
+			'storybook-preview-iframe',
 			( { value } ) => {
-				if ( value && !value.error ) {
-					this.frame( value ); // storybook composition
-				} else {
-					this.frame( 'storybook-preview-iframe' ); // vue-components storybook, no composition
-				}
+				this.frame( value );
 			},
 		);
 };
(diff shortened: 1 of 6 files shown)


vue-components/tests/e2e/custom-commands/openComponentStory.js
```diff
@@ -3,13 +3,9 @@ module.exports.command = function openComponentStory( component ) {
 		.click( `a[href="#${component}"]` )
 		.element(
```
jakobw, line 4: Yay for deleting code!! This can become even smoler!
```js
	return this.waitForElementVisible( '#root' )
		.click( `a[href="#${component}"]` )
		.frame( 'storybook-preview-iframe' );
```

tzhelyazkova: I just remembered what the problem was with the short version. Apparently Geckodriver has a problem with iframes and the tests on firefox fail if we don't invoke the iframe in a callback. 
https://github.com/nightwatchjs/nightwatch/issues/2030

jakobw: > Geckodriver has a problem with iframes

Can't blame it, iframes are weird. Alrighty, then this is fine as it is.

PR #149, opened 2020-08-14
tokens: add watcher for rebuilding during development

jakobw:
Run `docker-compose run --rm node lerna run watch --stream` during development to see the effect of changes to the token source files (almost) immediately in both the vue-components and the docs storybook without having to manually run `build:tokens`.


tzhelyazkova:
rebased

PR #152, opened 2020-08-18
docs: Update dev docs

tzhelyazkova:
### Document what devs need for the implementation of a component [T256791](https://phabricator.wikimedia.org/T256791)

### DIFF
--- a/docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
+++ b/docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
@@ -5,9 +5,26 @@ import { Meta } from '@storybook/addon-docs/blocks';
 
 # Developing Components
 
+Due to the decentralized nature of the design system, developers embedded in different feature teams will be accountable for creating new components to use in the feature they are building.
+
+The following is a set of instructions on different topics that might come up during the process of developing a new component.
+
+**Step by step workflow:**
+
+0. Gather relevant global and/or alias tokens.
+0. Create component-level tokens as described below.
+0. Create the vue component and write unit tests.
+0. Create component stories in the vue-components storybook.
+0. Create browser tests if needed (more info below).
+0. Open a PR and look at the UI Tests in Chromatic to check the results from the visual regression tests.
+0. Ask UX for a review.
+0. Ask a fellow developer for a technical review.
+0. Make a new release after the PR is merged to master.
+
+
 ## Working with Component-level Design Tokens
 
-A general overview of how we use design tokens in our design system can be found in the "Using design tokens" section of the [design documentation](?path=/story/documentation-how-to-create-reusable-components-designing-components--page). While global and alias tokens are created by system designers, the creation of component-level tokens is part of the component implementation and thus responsibility of developers.
+A general overview of how we use design tokens in our design system can be found in the "Using design tokens" section of the [design documentation](?path=/story/documentation-how-to-create-reusable-components-designing-components--page). While global and alias tokens are created by system designers, the creation of component-level tokens is part of the component implementation and thus the responsibility of developers.
 
 All component-level tokens must reference alias and global tokens listed in the design specification. This way, they act as a check for the completeness of the design specification.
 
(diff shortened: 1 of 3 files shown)


docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
```diff
@@ -5,14 +5,36 @@ import { Meta } from '@storybook/addon-docs/blocks';
...
 
 All component-level tokens must reference alias and global tokens listed in the design specification. This way, they act as a check for the completeness of the design specification.
 
 The name of a component-level token file should match the name of the component. For example, a component implemented in `Button.vue` should have a `Button.json` containing its tokens.
 
 The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[prop-][state-]<css-property>` - increasing specificity from left to right. A token containing the background-color (CSS property) value of a hovered (state) progressive (prop) button (component name) should be called `wikit-Button-progressive-hover-background-color`. Default prop values and states can be omitted from the name. A border-radius token that applies to all button types in all states can simply be named `wikit-Button-border-radius`.
 
-releasing components
+## Working with browser tests
```
SaiSan-WMDE, line 22: I find the content perfectly clear and well-written. The only suggestion I can think of is to maybe create a higher level section to group the two testing sections (e.g. "Testing components")?

docs/src/doc/decisions-and-ADRs/01_overview.stories.mdx
```diff
@@ -16,7 +16,12 @@ There is also some additional - very tech heavy - documentation contained in sub
...
-The code of Wikit is organized into multiple packages, each tending to its own distinct task:
+The code of WiKit is organized into multiple packages, each tending to its own distinct task:
 
 * docs - the docs you are looking at, this includes general documentation and demoing the other packages
 * tokens - the Design Tokens which Wikit, as a Design system, is based on
+* vue-components - the component library that implements the Design System
+
+### Vue components
+
+The library assumes it is going to be used in a Vue project, therefor Vue is externalized.
```
itamargiv, line 27: Nitpick: therefore

itamargiv, line 27: Another Nitpick: I'm not fully convinced this needs a full section, if the only purpose is to mention that Vue is externalized. We should cosider if this is redundant or maybe just add it as an (*) asterisk.

docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
```diff
@@ -5,14 +5,38 @@ import { Meta } from '@storybook/addon-docs/blocks';
 
 # Developing Components
 
+Due to the decentralized nature of the design system, developers embedded in different feature teams will be accountable for creating new components to use in the feature they are building.
```
itamargiv, line 8: This whole page is really great! thank you for the documentation. The only thing I would add here is maybe a section describing the step by step workflow for a component as a quick reference and overview. Something like:

0. Gather relevant alias tokens
0. Create component tokens
0. Write component code (and tests)
0. Create e2e tests (regression, selenium)

etc...

tzhelyazkova: That's a good idea, will do.

docs/src/doc/decisions-and-ADRs/01_overview.stories.mdx
tzhelyazkova: This already existed in another part of the docs, I just moved it to the appropriate section.
We wrote it, because usually when creating a Vue app we can externalize Vue in the config (because it's available in MediaWiki core) and the code documented itself, see [data-bridge](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/Wikibase/+/refs/heads/master/client/data-bridge/vue.config.js#7).
But here [`vue-cli` does that for us ](https://cli.vuejs.org/guide/build-targets.html#library)when we register the project as a library, so technically it's not documented anywhere.

docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
SaiSan-WMDE: That sounds good! Outlining the workflow in both the design and tech sections in "How to create reusable components" will make writing the Overview easier (we could _simply_ collect all the steps with a short description in there).

docs/src/doc/decisions-and-ADRs/01_overview.stories.mdx
SaiSan-WMDE: How is the content of these two sections (Packages and Vue components) related to the ADRs (e.g. are they relevant extracts, or relevant information that hasn't been capture in the ADR format)? Should we clarify that here?

At the same time, this feels like Getting started material 🤔

docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
itamargiv: perfect! looking great, thank you :)

docs/src/doc/decisions-and-ADRs/01_overview.stories.mdx
```diff
@@ -6,7 +6,7 @@ import { Meta } from '@storybook/addon-docs/blocks';
 
 Documentation on important decisions, non-tech and tech-related in the form of ADRs.
 
-## Wikit dev documentation
+## WiKit dev documentation
```
SaiSan-WMDE, line 9: Do you think this content should have its own section, at the same level as the _ADRs_ one? 🤔

PR #153, opened 2020-08-18
Create release workflow

tzhelyazkova:
This is the first step from [T260116](https://phabricator.wikimedia.org/T260116)
This workflow will be triggered when a new release is created on the GitHub repo.
The workflow will publish both the `vue-components` and the`tokens` packages with the same version.


itamargiv (approved):
:sunglasses: Cool :dancing_men:

PR #154, opened 2020-08-18
Add release workflow explanation to dev docs

tzhelyazkova:
This PR is part of [T260116](https://phabricator.wikimedia.org/T260116)


docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
```diff
@@ -34,7 +34,16 @@ The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[p
 
 ## Releasing a new component
 
-TODO
+Once your changes are merged and you want to release a new version of the components library you need to take the following steps:
+ - First create and push a new tag in compliance with [semantic versioning standards](https://semver.org/), e.g. `v1.2.3`
```
itamargiv, line 38: npm has a pretty handy tool for that, I wonder if we should just instruct developers to use that  or at least link to the docs (if that was the original intention) https://docs.npmjs.com/cli/version

itamargiv (commented):
Perfect, I just have one quesomment (which is a question and a comment rolled up in one...)

docs/src/doc/create-reusable-components/03_developing_components.stories.mdx
tzhelyazkova: The github workflow runs the tests before it publishes to npm, so that step is literally about the git command to create a tag. I guess we can add a link to this tool as well :)

itamargiv (approved):
Thank you :D

PR #160, opened 2020-08-19
Reorder button story to show all types at once

tzhelyazkova:
Remove the knobs and display button types in the same story.
This increases visual tests coverage significantly.

### DIFF
--- a/vue-components/stories/Button.stories.ts
+++ b/vue-components/stories/Button.stories.ts
@@ -1,5 +1,4 @@
 import Button from '@/components/Button';
-import { boolean, text, select } from '@storybook/addon-knobs';
 import { Component } from 'vue';
 
 export default {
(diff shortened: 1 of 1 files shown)


vue-components/stories/Button.stories.ts
```diff
@@ -10,24 +9,15 @@ export default {
...
-					'neutral',
-				),
-			},
-		},
-		template: '<Button v-html="content" :disabled="disabled" :type="buttonType" />',
+		// The normal button types are all in the same story to achieve high % of visual tests coverage.
+		// Do not use knobs to change the type unless you actively decide that is better than having test coverage.
+		template: `
+			<div>
+				<Button type="neutral">Normal</Button>
```
SaiSan-WMDE, line 16:
```suggestion
				<Button type="neutral">Neutral</Button>
```
This way, all button labels communicate their corresponding typology (flavor) for now. As mentioned in https://phabricator.wikimedia.org/T260797 the normal neutral button may have to be renamed in the future, since it's a candidate to become our official "secondary" or "default" button.

vue-components/stories/Button.stories.ts
```diff
@@ -10,24 +9,15 @@ export default {
...
-				),
-			},
-		},
-		template: '<Button v-html="content" :disabled="disabled" :type="buttonType" />',
+		// The normal button types are all in the same story to achieve high % of visual tests coverage.
+		// Do not use knobs to change the type unless you actively decide that is better than having test coverage.
+		template: `
+			<div>
+				<Button type="neutral">Neutral</Button>
+				<Button type="primaryProgressive">Progressive</Button>
```
SaiSan-WMDE, line 17:
```suggestion
				<Button type="primaryProgressive">Primary progressive</Button>
```
After aligning on this with Raja, we find that it would be valuable to specify the "use case" of both primary buttons in their labels.

vue-components/stories/Button.stories.ts
```diff
@@ -10,24 +9,15 @@ export default {
...
+				<Button type="primaryDestructive">Destructive</Button>
```
SaiSan-WMDE, line 18:
```suggestion
				<Button type="primaryDestructive">Primary destructive</Button>
```
ídem: After aligning on this with Raja, we find that it would be valuable to specify the "use case" of both primary buttons in their labels.

vue-components/stories/Button.stories.ts
```diff
@@ -10,24 +9,15 @@ export default {
...
-			buttonType: {
-				default: select(
-					'Type',
-					[ 'neutral', 'primaryProgressive', 'primaryDestructive' ],
-					'neutral',
-				),
-			},
-		},
-		template: '<Button v-html="content" :disabled="disabled" :type="buttonType" />',
+		// The normal button types are all in the same story to achieve high % of visual tests coverage.
```
itamargiv, line 12: Thanks for the comment! I was actually wondering why we were moving away from knobs :+1:

itamargiv (approved):

PR #163, opened 2020-08-20
Swap doc and design tokens order in Storybook

tzhelyazkova:
It's better to land on the Getting Started page than the Color tokens when you open the design system Storybook.
Once everyone is well acquainted with the design system we can rethink the order of the sections again.


itamargiv (commented):
Looking good. Just the conflict left...

itamargiv (approved):
Ready to go!

PR #181, opened 2020-08-27
docs: Add RTL icons section to ADR 9

tzhelyazkova:


docs/src/00-doc/05-decisions-and-ADRs/adr/0009-rtl-support.stories.mdx
```diff
@@ -24,8 +24,25 @@ We identified two challenges:
...
 
+### RTL support for icons
+Neither of the tools and solutions discussed above concern themselves with icons.
+
+#### Symmetrical icons
+Symmetrical icons don’t need to be flipped between LTR and RTL layouts.
+
+#### Non-symmetrical icons
+Non-symmetrical icons can be flipped (with css transform, e.g. [data-bridge button](https://gerrit.wikimedia.org/r/plugins/gitiles/mediawiki/extensions/Wikibase/+/refs/heads/master/client/data-bridge/src/presentation/components/EventEmittingButton.vue#289)) for RTL scripts.  
+They can also be provided with two separate `svg`s - one for LTR and one for RTL. The correct `svg` should be loaded based on the `dir` attribute.
```
SaiSan-WMDE, line 35: Even after doing some research, I'm not sure of which one of the two approaches would perform better, but – as metioned in Mattermost – the system iconography will contain RTL versions of the non-symmetrical icons that need adjustment, so the "separate svg" approach would be covered.

SaiSan-WMDE (approved):
Even after doing some research, I'm not sure of which one of the two approaches (axis flip or specific svg load) would perform better, but – as metioned in Mattermost – the system iconography will contain (and showcase in Storybook) RTL versions of the non-symmetrical icons that need adjustment, so the "separate svg" approach would be covered if we decide to go for it.

It'd be great to have a unified way to tackle RTL iconography. Until then, this is a great explanation.

itamargiv (approved):

PR #191, opened 2020-09-22
Use commit hash as chromatic url

manicki:
Extract and use the commit hash as the [chromatic permalink](https://www.chromatic.com/docs/permalinks#build-your-own-permalink) for saucelabs to access. This prevents issues with branch names and enables us to name our branches in any way we please.

Related To: [T262270](https://phabricator.wikimedia.org/T262270)

### DIFF
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -32,7 +32,7 @@ jobs:
       run: npm run chromatic
     - name: Extract branch name
       shell: bash
-      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/} | sed 's/[^A-Za-Z0-9_]/-/g')"
       id: extract_branch
     - name: Run e2e cross browser tests
       env:
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -32,11 +32,11 @@ jobs:
       run: npm run chromatic
     - name: Extract branch name
       shell: bash
-      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/} | sed 's/[^A-Za-Z0-9_]/-/g')"
+      run: echo "##[set-output name=branch;]$(echo \"${GITHUB_REF#refs/heads/}\" | md5sum"
       id: extract_branch
     - name: Run e2e cross browser tests
       env:
         SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
         SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


manicki:
changed it to "escape" everything else than digits, hyphen and Latin letters. Not sure how much sense this makes though. @itamargiv suggests hashing branch name instead

.github/workflows/test-and-demo.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       run: npm run chromatic
     - name: Extract branch name
       shell: bash
-      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+      run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/} | sed 's/[^A-Za-Z-]/-/g')"
```
itamargiv, line 35: Looks good so far (I think underscores are also possible, and caps are not allowed (according to the [docs](https://wmde.github.io/wikit/?path=/story/documentation-contributing--page#branches), though I'm not sure why caps or digits are not allowed)

itamargiv: BTW, If you want to validate this before merging, you want to add this particular file to the globbing pattern at the begining of this workflow, so that CI runs when this is changed.

manicki: right, I meant to added but failed. I also meant to include digits, done in another version.

Re: validating. I actually was trying to push a noop code change on top of those to trigger github action. No success so far, mostly because of failures with amending the sed commit

manicki:
so I cannot use sed apparently: https://github.com/wmde/wikit/runs/1208426458

.github/workflows/test-and-demo.yaml
itamargiv: well, the workflow itself isn't triggering though, I think it might be because this workflow only runs when you push to the `vue-components` subfolder. I think if you want to see if it works IRL then you need to add `.gihub/workflows/test-and-demo.yml` to the `paths` list in line 11

itamargiv:
Which line are you referring to?

manicki:
Smart. The same commit could be potentially subject to multiple builds, e.g. first on merging a PR, and then when tagging a new release? Do we foresee any problems due to this? I understand that might be a marginal edge case any way. I simply know too little on this chromatic thing

itamargiv:
> "Do we foresee any problems due to this?"
Not that I could think of, it appears as though chromatic simply gets the commit and branch info, runs the build script to output the sotorybook in a temp folder and then uploads the files to their container.Sso in this case, I think the project will just be rebuilt again.

tzhelyazkova (commented):
Looks good to me! Def better than using the branch name.
Can we rebase the dependabot PR to be on top of this so we can test on saucelabs?

itamargiv:
Good question, trying now

itamargiv:
@tzhelyazkova `dependabot/...` is now green :frog:

tzhelyazkova (commented):
Awesome! ✨

tzhelyazkova (approved):

PR #193, opened 2020-09-24
vue-components: create TextInput component

jakobw:
# TextInput component :sparkles: 

- [x] initial styling
- [x] width prop
- [x] responsiveness
- [x] label
- [x] placeholder
- [x] disabled
- [x] states (hover, active/focus, ...)
- [x] add warning/error icons

Phabricator task: https://phabricator.wikimedia.org/T263734
Figma spec: https://www.figma.com/file/gi3JYwSUQYhKS0mHwK25cQwi/WiKit?node-id=3975%3A1

Most of the code here was written in pair programming by @guergana and @jakobw.

PR #194, opened 2020-09-29
Improve component-level token naming docs

jakobw:
At the time the original naming conventions were documented, the Button
was the only existing component which only exists of a single element.
While working on the TextInput component in #193 we found this addition
to the naming scheme to work well for us.

### DIFF
--- a/docs/src/00-doc/03-create-reusable-components/03_developing_components.stories.mdx
+++ b/docs/src/00-doc/03-create-reusable-components/03_developing_components.stories.mdx
@@ -37,7 +37,17 @@ All component-level tokens must reference alias and global tokens listed in the 
 
 The name of a component-level token file should match the name of the component. For example, a component implemented in `Button.vue` should have a `Button.json` containing its tokens.
 
-The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[prop-][state-]<css-property>` - increasing specificity from left to right. A token containing the background-color (CSS property) value of a hovered (state) progressive (prop) button (component name) should be called `wikit-Button-progressive-hover-background-color`. Default prop values and states can be omitted from the name. A border-radius token that applies to all button types in all states can simply be named `wikit-Button-border-radius`.
+The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[prop-][element-][state-]<css-property>` - increasing specificity from left to right. The order depends on the context however; states should follow the element on which the interaction occurred.
+
+**For example**: 
+
+A "hover" on the "menu-item" impacting the "menu-item" style yields 
+
+<code>wikit-Dropdown-progressive-<strong>menu-item-hover</strong>-background-color</code>
+  
+While a "hover" on the "Dropdown" component impacting the "menu-item" style yields
+
+<code>wikit-<strong>Dropdown-progressive-hover</strong>-menu-item-background-color</code>
 
 ## Testing Components
 


docs/src/00-doc/03-create-reusable-components/03_developing_components.stories.mdx
```diff
@@ -37,7 +37,7 @@ All component-level tokens must reference alias and global tokens listed in the
...
+The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[element-][prop-][state-]<css-property>` - increasing specificity from left to right. A token containing the background-color (CSS property) value of a hovered (state) progressive (prop) button (component name) should be called `wikit-Button-progressive-hover-background-color`. Default prop values and states can be omitted from the name. A border-radius token that applies to all button types in all states can simply be named `wikit-Button-border-radius`. The `element` part of the name is relevant for more complex components consisting of multiple pieces (= elements), e.g. a TextInput with a label may have a `wikit-TextInput-label-font-family` token.
```
Ladsgroup, line 40: I'd say let's make it clear that in this example "label" is the element (or go with an example that html tag is obvious, like "div" but it doesn't make much sense in this context).

The other concern is that this paragraph is getting really big. I think this should be broken apart but that can be done later.

jakobw: I'm not sure I understand the suggestion. It's specifically "label" not as in `<label>` but as in "the label of the text input". Tokens are intended to be technology-agnostic, so this isn't talking about HTML in any way. I'm using the word element more in the sense of BEM-element, not DOM element. I'd be happy to change this if you see a way to make it clearer.

wiese: FWIW I had to check the change out locally to be able to reason about the diff with [diff-so-fancy](https://github.com/so-fancy/diff-so-fancy#screenshot).

Maybe part of addressing Amir's concern about brevity could be to not add another "more complex components" example but change the base example to that directly (and mention that parts are optional, if that is not clear already).

I like the change in general and it's in line with my mental model of how this would further develop.

jakobw: Thanks, @wiese. Coming up with a token name consisting of all possible component name parts turned out to be a useful exercise. We don't have any existing case for this (not even in WIP code) but it made me realize that in most cases I can think of the prop should come before the element. I changed this accordingly and added a made up example. Let me know if that makes sense!

wiese: > made me realize that in most cases I can think of the prop should come before the element

Oh, because that way the `element` (e.g. "menu-item") can have its `css-property` (e.g. "background-color") styled differently for different `[prop-]` (e.g. "progressive"). That makes sense.
Why does the same logic not apply to `state` (e.g. "hover"), so that `element` only comes after `state`, e.g. 

`wikit-Dropdown-progressive-hover-menu-item-background-color` instead of 
`wikit-Dropdown-progressive-menu-item-hover-background-color` though?
I suppose the reason lies in the statement "increasing specificity from left to right". In the example we are talking about (assumption here, correct me please!) we are talking about a "hover" on the "menu-item", not about a "hover" on the "Dropdown" component impacting the "menu-item" style. This however also means that the order is not fully fixed but instead depends on the behavior we want to achieve. Does that make sense at all? Sorry for the noise.

jakobw: Thanks @wiese, that thought crossed my mind as well and it absolutely makes sense to add this. Updated per our mattermost conversation! :+1:

docs/src/00-doc/03-create-reusable-components/03_developing_components.stories.mdx
```diff
@@ -37,7 +37,7 @@ All component-level tokens must reference alias and global tokens listed in the
...
 
-The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[prop-][state-]<css-property>` - increasing specificity from left to right. A token containing the background-color (CSS property) value of a hovered (state) progressive (prop) button (component name) should be called `wikit-Button-progressive-hover-background-color`. Default prop values and states can be omitted from the name. A border-radius token that applies to all button types in all states can simply be named `wikit-Button-border-radius`.
+The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[prop-][element-][state-]<css-property>` - increasing specificity from left to right. The order depends on the context however; states should follow the element on which the interaction occurred. Example: a "hover" on the "menu-item" impacting the "menu-item" style yields `wikit-Dropdown-progressive-menu-item-hover-background-color` while a "hover" on the "Dropdown" component impacting the "menu-item" style yields `wikit-Dropdown-progressive-hover-menu-item-background-color`.
```
itamargiv, line 40:
```suggestion
The recommended naming scheme for individual tokens is `wikit-<ComponentName>-[prop-][element-][state-]<css-property>` - increasing specificity from left to right. The order depends on the context however; states should follow the element on which the interaction occurred. 

For example: 

A "hover" on the "menu-item" impacting the "menu-item" style yields 
<code>wikit-Dropdown-progressive-<strong>menu-item-hover</strong>-background-color</code>
  
While a "hover" on the "Dropdown" component impacting the "menu-item" style yields
<code>wikit-Dropdown-<strong>progressive-hover</strong>-menu-item-background-color</code>.
```

jakobw: Thanks, @itamargiv ! I agree that the description is a bit hard to follow. Unfortunately I lack the necessary skills to illustrate the example, but I like the changes you made to at least split apart the long sentence and highlight the important differences!

itamargiv (approved):

PR #195, opened 2020-09-29
Add spacing tokens

SaiSan-WMDE:
Tokens used to space elements in compositions


tokens/properties/alias.json
```diff
@@ -487,6 +490,23 @@
                     "value": "{dimension.dynamic.size.56.value}"
                 }
             }
+        },
+        "spacing": {
```
guergana, line 494: maybe we should change the "padding" property to "spacing" or "layout" instead. to avoid confusion when using "padding" tokens in margins.

guergana (changes requested):

tokens/properties/alias.json
SaiSan-WMDE: Spacing tokens (which may after all be called "layout") were the ones devised to be used to define margins, actually. Or maybe I'm not interpreting your comment correctly, sorry!

guergana (approved):

PR #198, opened 2020-10-01
correct dynamic size value

SaiSan-WMDE:
Tiny PR to correct a human error made when specifying the value of the token "dimension-dynamic-size-125", used to define the large size of icons (1.250 em, 20px).


jakobw (approved):
Yay bigger icons!

PR #201, opened 2020-10-02
TextInput: emit input event on user input

jakobw:
This is necessary in order to get the user input out of that component,
and bind it e.g. via v-model:
https://vuejs.org/v2/guide/components.html#Using-v-model-on-Components


guergana (approved):
Reviewed and tested. Looks good!

PR #204, opened 2020-10-02
docs for version release using lerna version

guergana:
+ added in README
+ added in storybook

### DIFF
--- a/docs/src/00-doc/04_release_management.stories.mdx
+++ b/docs/src/00-doc/04_release_management.stories.mdx
@@ -15,6 +15,10 @@ How to release a new verision is documented in the section [Developing component
 
 A version of the Design System and component library is released when the component and tokens are done and "ready" and the application wanting to use them makes use of the version in which they are contained.
 
+To release a new version run the command `lerna version [major | minor | patch | premajor | preminor | prepatch | prerelease]`. This will take of updating the version number, committing and pushing the changes.
+
+[lerna version docs](https://github.com/lerna/lerna/tree/master/commands/version). 
+
 ## Releasing changes to an existing component or tokens
 
 Design System changes are supposed to influence application styles, therefore a Design System change is not finished until it's propagated to all consuming applications.
--- a/docs/src/00-doc/04_release_management.stories.mdx
+++ b/docs/src/00-doc/04_release_management.stories.mdx
@@ -27,7 +27,7 @@ It is also their responsibility to inform other relevant roles like UX and PM (i
 ## Creating a release 
 
 Once your changes are merged and you want to release a new version of the components library you need to take the following steps:
- - To release a new version run the command `lerna version [major | minor | patch | premajor | preminor | prepatch | prerelease]`. This will take of updating the version number, committing and pushing the changes.
+ - To release a new version run the command `lerna version [major | minor | patch | premajor | preminor | prepatch | prerelease]`. This will take care of updating the version number, committing and pushing the changes.
 [lerna version docs](https://github.com/lerna/lerna/tree/master/commands/version). Wait for the PR to be merged.
  - After it's merged create and push a new tag, named after the version number with a "v" prefix, e.g. `v1.2.3`. Alternatively you can use the npm provided tool [npm-version](https://docs.npmjs.com/cli/version)
  - Then go to the GitHub repository and create a new release out of that tag, describing what changed, like in the image below.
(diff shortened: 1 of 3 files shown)


docs/src/00-doc/04_release_management.stories.mdx
```diff
@@ -15,6 +15,10 @@ How to release a new verision is documented in the section [Developing component
 
 A version of the Design System and component library is released when the component and tokens are done and "ready" and the application wanting to use them makes use of the version in which they are contained.
 
+To release a new version run the command `lerna version [major | minor | patch | premajor | preminor | prepatch | prerelease]`. This will take of updating the version number, committing and pushing the changes.
```
itamargiv, line 18: Documentation for releasing already exists in the DS docs in: https://wmde.github.io/wikit/?path=/story/documentation-how-to-create-reusable-components-developing-components--page#releasing-a-new-component

If something is required to be updated it should probably go there. Unless you are suggesting to move the instructions, in which case, the should be removed from there and added here.

itamargiv (changes requested):

docs/src/00-doc/04_release_management.stories.mdx
jakobw: Ooh, we didn't see this existing bit of documentation. I wouldn't have expected the docs there, given most of that document is about working with the component library, while releasing concerns all (=both tokens and component lib) parts of the design system. I'd suggest to move the release instructions to this document here, because it's more centrally located.

guergana: So the new paragraphs that were added to storybook should go in this section (README) instead? Just asking for confirmation.

itamargiv: I think what @jakobw is saying is that the content from [this document](https://github.com/wmde/wikit/blob/cf58b904ed4b0b6b37f4df7bee3e5cdaf6a27225/docs/src/00-doc/03-create-reusable-components/03_developing_components.stories.mdx#releasing-a-new-component) should go in this document.

Either or, I don't hink this information should be in the README file so that we don't need to make sure that the information is synchronised between two places. We can link to here from the readme file.

guergana: Yes, but you are saying the opposite. I can move them either way but you have to agree.

jakobw: @guergana I don't think @itamargiv and I are disagreeing. I agree that we shouldn't duplicate the same information in the README.md in the root dir of the repo. And as far as I can tell Itamar wasn't disagreeing that here would be a good place to keep the release instructions that currently reside in the "Developing Components" doc.

guergana: Nevermind! Sorry, got confused.

docs/src/00-doc/04_release_management.stories.mdx
```diff
@@ -22,4 +22,19 @@ Design System changes are supposed to influence application styles, therefore a
...
 
 Deciding if a change to the Design System is "breaking" is the responsibility of the developer(s) making the change.  
-It is also their responsibility to inform other relevant roles like UX and PM (if the change is being made at their request) of that, so the additional effort of updating all consuming applications can be taken into consideration.
\ No newline at end of file
+It is also their responsibility to inform other relevant roles like UX and PM (if the change is being made at their request) of that, so the additional effort of updating all consuming applications can be taken into consideration.
+
+## Creating a release 
+
+Once your changes are merged and you want to release a new version of the components library you need to take the following steps:
+ - To release a new version run the command `lerna version [major | minor | patch | premajor | preminor | prepatch | prerelease]`. This will take of updating the version number, committing and pushing the changes.
```
jakobw, line 30:
```suggestion
 - To release a new version run the command `lerna version [major | minor | patch | premajor | preminor | prepatch | prerelease]`. This will take care of updating the version number, committing and pushing the changes.
```

jakobw (commented):
Missing a word, otherwise looks good to me now! :+1:

docs/src/00-doc/04_release_management.stories.mdx
guergana: done

jakobw (approved):
Thanks! LGTM :sparkles:

itamargiv (approved):
Danke dir!

PR #206, opened 2020-10-06
use phabricator as a bug tracker for all packages

wiese:
The title should maybe read "for all published packages and the root
project" instead but it's cumbersome.

A short research did not yield any information if just putting it in the
root package.json would somehow have lerna take care of distributing it
to the packages but I wanted to be sure the information is part of the
published artifacts.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,5 +21,8 @@
   "devDependencies": {
     "lerna": "^3.22.1",
     "npm-run-all": "^4.1.5"
+  },
+  "bugs": {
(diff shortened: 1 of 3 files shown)


wiese:
As a consequence, "issues" should be disabled here.

SaiSan-WMDE (approved):
Agree to using Phabricator as the central place to document/track issues. Thanks!

tzhelyazkova (approved):

Ladsgroup (approved):

PR #207, opened 2020-10-07
TextInput: fix incorrect typing

jakobw:
This caused a build error in the Query Builder code base when loading the TextInput.vue source file.

Interestingly I'm not getting any errors running `docker-compose run --rm node lerna run build --scope @wmde/wikit-vue-components` even before fixing them, but the problems are definitely there. I can't see any major differences between the two (Query Builder and vue-components/ here) build setups. I'd be curious to hear if anyone has an idea why the type checking isn't as strict here.

Phabricator: https://phabricator.wikimedia.org/T264784

### DIFF
--- a/vue-components/src/components/TextInput.vue
+++ b/vue-components/src/components/TextInput.vue
@@ -33,9 +33,11 @@ export default Vue.extend( {
 	props: {
 		error: {
 			type: Object,
-			validator( error: object ): boolean {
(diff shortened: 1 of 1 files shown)


guergana (approved):
Looks good!

wiese:
I created https://phabricator.wikimedia.org/T264891

Until we discuss the idea in further detail I am not surprised by cross-repository use of TypeScript files causing problems.

PR #209, opened 2020-10-09
vue-components: extract Input sub component from TextInput

jakobw:
We are about to start implementing the [Lookup](https://phabricator.wikimedia.org/T264089) component which in it's resting state is indistinguishable from a TextInput component. Since the parts (label, input, feedback message) that the two  components share are supposed to look and behave identical, we extract them into smaller internal components that are  reusable.

Bug: [T265005](https://phabricator.wikimedia.org/T265005)


vue-components/src/components/TextInput.vue
```diff
@@ -174,7 +125,7 @@ $base: ".wikit-TextInput";
 	}
 
 	& #{$base}__label {
-		color: $wikit-TextInput-input-color;
+		color: $wikit-TextInput-label-font-color;
```
Ladsgroup, line 128: This has changed but I can't find out why.

Ladsgroup:
LGTM, one question.

vue-components/src/components/TextInput.vue
jakobw: This should have used the `$wikit-TextInput-label-font-color` to begin with, and only looked correct by chance. The fact that we're using the wrong token here surfaced when I extracted the Input element and its styles, and removed all `$wikit-TextInput-input-...` tokens, which shouldn't have been referenced anywhere anymore at that point.

I can extract it into a separate commit if it looks confusing.

itamargiv:
I'm trying understand the use case for this sub component. AFAICT, this is just encapsulating certain styles with a regular text input, where is this going to be used apart from this component?

jakobw:
> I'm trying understand the use case for this sub component. AFAICT, this is just encapsulating certain styles with a regular text input, where is this going to be used apart from this component?

@itamargiv I tried to answer this question in the PR description. It's also outlined in more detail in the description of [the linked task](https://phabricator.wikimedia.org/T265005). We want to keep the existing TextInput component as small as possible while being able to reuse its parts for the [Lookup component](https://phabricator.wikimedia.org/T264089).

itamargiv:
But wouldn't this effectively create coupling between Lookup and TextInput, in a way that is not necessarily desirable? Also, at least on a personal level I would find it somewhat confusing to have components that are not documented in the storybook.

vue-components/src/components/TextInput.vue
Ladsgroup: no, just had a question about it, it's too small to be extracted. Thanks!

vue-components/src/components/Input.vue
```diff
@@ -0,0 +1,106 @@
...
+			},
+			default: null,
+		},
+	},
+
+	computed: {
+		classes(): string[] {
+			const classes = [ 'wikit-Input' ];
+
+			if ( this.feedbackType ) {
```
bereket-WMDE, line 33: does this ever resolve to false ?
AFAIK, typescript objects always resolve to true, even if they're empty.
are we checking here if feedbackType is empty or undefined ?

jakobw: `feedbackType` is a prop that is `null` by default (if not specified) or one of the strings specified in the prop validator. So yes, it can be falsy!

jakobw:
@itamargiv, @SaiSan-WMDE, @guergana and I discussed this separately in a call, and decided that extracting this `Input` component is the way to go after all, but that it's important to document this, and not hide this as an implementation detail. We considered separating the `TextInput` into two components - the combination of input and label, and the FeedbackMessage component, but concluded that the two should stay contained in the same component.

I updated this PR to
* add the new `Input` component to the storybook
* add a note to the `TextInput` docs that it's using `Input` internally. (I briefly looked into generating this documentation automatically, but couldn't find anything)

jakobw:
Updated to group the two Input stories together. This is ready for chromatic review, @SaiSan-WMDE !

SaiSan-WMDE:
Hello! All good. Although for some reason I didn't expect to find `wikit-Input--warning` and `--error` in the Basic input page, but I guess these are reusable components too rather than states, right?

jakobw:
> Hello! All good. Although for some reason I didn't expect to find `wikit-Input--warning` and `--error` in the Basic input page, but I guess these are reusable components too rather than states, right?

@SaiSan-WMDE `wikit-Input--warning` and `wikit-Input--error` are modifications of the same component, yup!

guergana (approved):
Looks good!

PR #210, opened 2020-10-09
vue-components: extract ValidationMessage sub component from TextInput

jakobw:
Similar to #209 but for the `FeedbackMessage` component.

Bug: [T265005](https://phabricator.wikimedia.org/T265005)

### DIFF
--- a/tokens/properties/components/TextInput.json
+++ b/tokens/properties/components/TextInput.json
@@ -31,38 +31,6 @@
 			"padding-block-end": {
 				"value": "{dimension.spacing.xsmall.value}"
 			}
-		},
-		"error-message": {
-			"padding-block-start": {
-				"value": "{dimension.spacing.xsmall.value}"
-			},
-			"font-family": {
-				"value": "{font.family.style.body.value}"
-			},
-			"font-size": {
-				"value": "{font.size.style.body.value}"
-			},
-			"font-weight": {
-				"value": "{font.weight.style.body.value}"
-			},
-			"line-height": {
-				"value": "{font.line-height.small.value}"
-			},
-			"error": {
-				"color": {
-					"value": "{font.color.error.value}"
-				}
-			},
-			"warning": {
-				"color": {
-					"value": "{font.color.warning.value}"
-				}
-			},
-			"icon": {
-				"margin-right": {
-					"value": "{dimension.spacing.xsmall.value}"
-				}
-			}
 		}
 	}
 }
--- a/tokens/properties/components/ValidationMessage.json
+++ b/tokens/properties/components/ValidationMessage.json
@@ -1,0 +1,34 @@
+{
+    "wikit-ValidationMessage": {
+        "padding-block-start": {
+            "value": "{dimension.spacing.xsmall.value}"
+        },
+        "font-family": {
+            "value": "{font.family.style.body.value}"
+        },
+        "font-size": {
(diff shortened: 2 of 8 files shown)


vue-components/src/components/FeedbackMessage.vue
```diff
@@ -0,0 +1,66 @@
...
+		/>
+		{{ message }}
+	</div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import Icon from './Icon.vue';
+
+export default Vue.extend( {
```
Ladsgroup, line 17: You had a similar message for Input component saying it's internal and should not be used outside. Maybe we can have that in here too?

Ladsgroup:
One nitpick and good to go.

vue-components/src/components/FeedbackMessage.vue
jakobw: Good point, thanks. Done!

Ladsgroup (approved):

SaiSan-WMDE:
Leaving a heads-up here that, since the inline feedbackMessage is its own component now, it should have its own page in Storybook, close to the "block messages" (coming soon). Suggested structure:

Message
- Block
- Inline

Would it be worth reconsidering the naming of the "feedbackMessage" component to "inlineMessage", in order to differentaite both variants (they both provide feedback)? This terminology is also taken straight from the WMF Style guide's [Messages page](https://design.wikimedia.org/style-guide/components/messages.html).

jakobw:
Added to storybook, and renamed to ValidationMessage after discussing with @SaiSan-WMDE. Ready for chromatic review!

jakobw:
Hi @SaiSan-WMDE thanks for the comments! Replying inline:


> * I don't think we'll need to nest any pages under Basic and TextInput. Would it be possible to remove the "All" level?

_\*dies inside a little\*_

I expected this to come up and unfortunately I don't know how we can easily get rid of this extra nesting level. If you look closely you'll see that all component stories follow this format of _Component Name_/_some attribute_ (Button/normal, Icon/all, ...). 

The [docs](https://storybook.js.org/docs/vue/api/csf) say "Every component story file consists of a required default export and one or more named exports.", with the default export being tied to the component (top-level navigation item), and the named export being this nesting level that we want to get rid of.

While typing this reply I asked in the DS channel and the other devs also don't see a way to do it with our current setup. I created https://phabricator.wikimedia.org/T265454 and hope we can keep it as is for now :/

> * Should the text of ValidationMessage display styles? I unfortunately think there're some alignment issues too (҂◡_◡)

OMG yes. Sorry. Fixed.

> * I assumed that ValidationMessage would be nested within a "Message" section that will also contain the BlockMessage component. Would that make sense?

Yup, done!

> * I didn't expect to find `wikit-Input--warning` and `--error` in the Basic input page, but I guess these component states are reusable too?

Yep!

SaiSan-WMDE:
> While typing this reply I asked in the DS channel and the other devs also don't see a way to do it with our current setup. I created https://phabricator.wikimedia.org/T265454 and hope we can keep it as is for now :/

Hey @jakobw! Didn't mean to make you die inside 😭 I thought that this could be either solved in one second or require to change the whole page set up. Thanks so much for creating that ticket! We definitely can leave it as is 👍 

Thanks for all the fixes & changes 🙇‍♀️  All looks 💯 now!

PR #216, opened 2020-10-14
v1.1.2

itamargiv:
Updating version per release. Though I'm not sure this is the right thing (used `lerna version` as instructed in the Release documentation), do we need to bump the version of all the packages on each change?


tzhelyazkova (approved):

jakobw (approved):
:rocket:

PR #218, opened 2020-10-14
vue-components: plain storybook extract command

wiese:
At the time of writing, storybook 6 was not released yet and running the
`extract` command on a storybook 5 (the npx default) caused an error.

Not sure if pinning to the now-current storybook version would be
better, but it's not how the storybook people document it[1], and will
cause future update effort (or inconsistency) while it is unlikely the
command will ever change without us having to touch the line anyways.

[1]: https://storybook.js.org/docs/vue/workflows/storybook-composition#improve-your-storybook-composition

### DIFF
--- a/vue-components/package-lock.json
+++ b/vue-components/package-lock.json
@@ -7156,9 +7156,9 @@
 			}
 		},
 		"bl": {
-			"version": "1.2.2",
-			"resolved": "https://registry.npmjs.org/bl/-/bl-1.2.2.tgz",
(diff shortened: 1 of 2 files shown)


wiese:
Found https://github.com/storybookjs/storybook/issues/12204, relief in the fact that we are not the only ones concerned that a config already known to storybook in machine-readable format has to extracted from it using a webbrowser, and reconciliation in hearing that the sb team [has taken note of this problem](https://github.com/storybookjs/storybook/issues/12204#issuecomment-679031641) and it may be resolved at some point.

jakobw (approved):
Nice!

PR #219, opened 2020-10-14
Implement menu part of the Lookup component

guergana:
Bug: T265008


tokens/properties/components/LookupMenu.json
```diff
@@ -0,0 +1,88 @@
...
+				"value": "{transition.interaction.menu.duration.value}"
+			},
+			"transition-property" : {
+				"value": "{transition.interaction.menu.property.value}"
+			},
+			"transition-timing-function" : {
+				"value": "{transition.timing-function.ease.value}"
+			}
+		},
+		"label" : {
```
jakobw, line 53: Shouldn't this be nested inside "item"? It's the label of a menu item, not the label of the menu.

tokens/properties/components/LookupMenu.json
```diff
@@ -0,0 +1,88 @@
...
+				"value": "{font.weight.bold.value}"
+			},
+			"line-height": {
+				"value": "{font.line-height.style.label.value}"
+			},
+			"font-color": {
+				"value": "{font.color.base.value}"
+			}
+		},
+		"description" : {
```
jakobw, line 70: same as above

vue-components/src/components/Lookup.vue
```diff
@@ -0,0 +1,122 @@
...
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import ValidationMessage from './ValidationMessage.vue';
+import Input from './Input.vue';
+import LookupMenu from './LookupMenu.vue';
+
+/**
+ * Text input fields are form elements that let users input and edit values in the form of text.
```
jakobw, line 28: Copy-pasta from `TextInput`. Needs its own description.

vue-components/src/components/Lookup.vue
```diff
@@ -0,0 +1,122 @@
...
+	data() {
+		return {
+			focused: false,
+			inputValue: '',
+		};
+	},
+	props: {
+		menuItems: {
+			type: Array,
+			default: (): [] => [],
```
jakobw, line 43: `[]` literally means empty array, which is technically correct for the default?! I'd have expected this to eventually declare it to be a list of menu items, though.

tokens/properties/components/LookupMenu.json
guergana: yes

tokens/properties/components/LookupMenu.json
guergana: agree

vue-components/src/components/Lookup.vue
guergana: worked on

vue-components/src/components/Lookup.vue
guergana: we discussed it, we think it is fine for now.

vue-components/src/components/Lookup.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+	<label :class="[ 'wikit', 'wikit-Lookup', `wikit-Lookup--${width}` ]">
+		<div class="wikit-Lookup__label">{{ label }}</div>
+		<Input
+			@input.native="inputValue = $event.target.value"
+			@focus.native="focused = true"
```
itamargiv, line 6: I might be missing something here, but is it really necessary to reimplement the native focus and blur with a controlled `focus` prop? I don't see it being used elsewhere within the component behaviour.

vue-components/src/components/LookupMenu.vue
```diff
@@ -0,0 +1,88 @@
...
+	props: {
+		menuItems: {
+			type: Array,
+			default: (): [] => [],
+		},
+	},
+	watch: {
+		menuItems(): void {
+			const menuItems = this.$el.querySelectorAll( '.wikit-LookupMenu__item' );
+			if ( menuItems.length > 6 ) {
```
itamargiv, line 31: I see why 6 but it could be considered a magic number, is it possible to add a comment explaining it here (that the menu should be scrollable with more than 6 items long, I assume)?

vue-components/src/components/LookupMenu.vue
```diff
@@ -0,0 +1,88 @@
...
+	},
+} );
+</script>
+
+<style lang="scss">
+$base: '.wikit-LookupMenu';
+
+#{$base} {
+	min-width: 100%;
+	width: 200%;
```
itamargiv, line 45: I am a bit confused by this, as these values are not encoded in design tokens, though I'm not sure if they should be?

itamargiv (changes requested):
Looks great, thank you. Requesting changes because of the `label` `div` confusion, ther rest are just questions - comments. 

Also did we decide not to add browser tests to the components of this hike? Is it something we are reserving for later? as I see none were added for any of the components of this hike.

micgro42:
> Also did we decide not to add browser tests to the components of this hike? Is it something we are reserving for later? as I see none were added for any of the components of this hike.

There should definitely be browser tests. Though as I understand it, it was decided to implement them when the full intended functionality of the component is available. That is tracked in an extra sub-ticket: https://phabricator.wikimedia.org/T265014

vue-components/src/components/LookupMenu.vue
SaiSan-WMDE: I had the same doubt while specifying the component. I guess that using alias tokens to define the width of the menu would be consistent, and convenient in case we make this a reusable element (comparable to the input).

tokens/properties/components/LookupMenu.json
```diff
@@ -0,0 +1,91 @@
...
+			},
+			"color" : {
+				"value": "{border.color.base.default.value}"
+			}
+		},
+		"box-shadow": {
+			"value": "{box-shadow.elevation.menu.value}"
+		},
+		"item" : {
+			"padding-vertical": {
```
micgro42, line 24: This should probably be `spaceing-vertical` or something similar

guergana: they are referenced as padding-vertical and padding-horizontal in the design files. don't know what to make of this.

micgro42: No, I don't think that is has to be the same as the base token, either.
Rather that it should not needlessly preclude the specific implementation, and apparently there is some work going on removing some implementation detail from earlier token names. OTOH, if that term (i.e. `padding`) is a good way to talk about it from a design perspective, then it's probably OK.

vue-components/src/components/Lookup.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+	<label :class="[ 'wikit', 'wikit-Lookup', `wikit-Lookup--${width}` ]">
+		<div class="wikit-Lookup__label">{{ label }}</div>
```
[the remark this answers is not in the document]
guergana: yes. we need to do something similar to this here: https://github.com/wmde/wikit/pull/228/files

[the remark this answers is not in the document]
jakobw: Agree! Rebased on top of #228 and done!

jakobw:
Added some slightly nicer to look at examples for the story.

I think this is ready for a first look, @SaiSan-WMDE! Note that this doesn't contain any of the expected behavior beyond expanding and collapsing the Lookup menu when interacting with it. Keyboard navigation, changing the text in the input field upon selecting, and filtering will come later.

vue-components/src/components/LookupMenu.vue
```diff
@@ -0,0 +1,98 @@
...
+	props: {
+		menuItems: {
+			type: Array,
+			default: (): [] => [],
+		},
+	},
+	watch: {
+		menuItems(): void {
+			const menuItems = this.$el.querySelectorAll( '.wikit-LookupMenu__item' );
+			// the height automatically adjusts for up to 6 elements, then shows a scrollbar
```
itamargiv, line 38: Shouldn't the comment say 7 (0 based index :P )?

itamargiv (commented):
Looking good, thanks for the changes, one small comment

vue-components/src/components/LookupMenu.vue
jakobw: It's correct, I think!?

It checks for `if ( menuItems.length > maxNumberOfElementsDisplayed )` i.e. it does nothing special (= grows to the content height) for up to 6, and if there are more then it cuts off at `menuItems[ maxNumberOfElementsDisplayed ].offsetTop` which is the top of the 7th element. So there are 6 displayed, right?

guergana: I also think it is correct with 6.

itamargiv: looking at the storybook I see seven list items.

jakobw: Ah, you're right! Fixed, thanks :+1:

tokens/properties/components/LookupMenu.json
guergana: I asked @SaiSan-WMDE and we thought that since this is a conceptual issue, we can probably leave it like this and discuss this topic with the design team and change it retroactively when a consensus is achieved.

itamargiv (approved):
\o/ Is done! Thank you!

guergana:
> \o/ Is done! Thank you!
> 
> 

I must be dreaming!

jakobw (approved):
Rebased and squashed the commits with the latest update. LGTM :rocket:

PR #211, opened 2020-10-12
Add prepare script fo vue-components

itamargiv:
This PR aims to resolve an issue with the `dist` folder being excluded from our npm package. To validate and reproduce locally, please follow these instructions: https://phabricator.wikimedia.org/T264891#6536974

Once this PR is merged, a release needs to be made to confirm the issue was indeed fixed and fix the bug this PR addresses.

Bug: [T264891](https://phabricator.wikimedia.org/T264891)

PR #222, opened 2020-10-15
vue-components: fix dist entry point

jakobw:
This allows consuming applications to load components via
`import { Button } from '@wmde/wikit-vue-components';` which is what our
docs already claim to work. The `main` entry that is currently there
(dist/index.js) doesn't exist.

(Sorry, I should've noticed that while reviewing #211.)

I tried this change out in the query builder repo, and it works ~~, but complains about missing type declarations, which I think can be added in a separate change~~.

Added the missing puzzle piece for enabling imports from a TS code base in #223.

### DIFF
--- a/vue-components/package.json
+++ b/vue-components/package.json
@@ -26,7 +26,7 @@
     "chromatic": "npx chromatic build --project-token=3evi132fpys --build-script-name=build:storybook",
     "prepare": "npm run build"
   },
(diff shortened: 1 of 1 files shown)


itamargiv (approved):
LGTM, thanks for catching this, shouldv'e noticed that as well :grimacing:

PR #223, opened 2020-10-16
vue-components: export .d.ts files

jakobw:
Exporting type declaration files with the component library allows us to import its bundled `dist/wikit-vue-components.common.js` file from a TypeScript code base without producing compile-time errors. For now, the types for exported components don't contain any information beyond declaring them as Vue components, i.e. no information about props or slots.

The test files are removed from the tsconfig `include` entry to avoid exporting type declaration files for them. This shouldn't have 
 any effect when running the tests.

### DIFF
--- a/vue-components/package.json
+++ b/vue-components/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "serve": "vue-cli-service serve",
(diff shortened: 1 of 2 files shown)


Ladsgroup (approved):
LGTM

itamargiv (approved):
LGTM, well caught :bug: thank you.

PR #224, opened 2020-10-16
Input: re-emit input event to allow parents to listen for `@input`

jakobw:
This also allows us to bind a value to the Input component via `v-model` which we tripped over while working on #219.


itamargiv (approved):
LGTM, thanks! :sunglasses:

PR #225, opened 2020-10-16
Add Message component

Ladsgroup:
Based on the Figma spec outlined in the ticket

Bug: T265581


vue-components/src/components/Message.vue
```diff
@@ -0,0 +1,91 @@
+<template>
+	<div :class="[ 'wikit', 'wikit-Message', `wikit-Message--${type}` ]">
+		<span class="wikit-Message__content">
+			<Icon
+				class="wikit-Message__icon"
+				:type="type === 'error' ? 'error' : 'alert'"
```
guergana, line 6: Looks like we need to add more icon types / names to the Icon component for this component

guergana (changes requested):
Looks pretty good so far. Just a couple details.

vue-components/src/components/Message.vue
micgro42: Yes, but there is a separate task for that.

vue-components/src/components/Message.vue
```diff
@@ -0,0 +1,91 @@
...
+				:color="type"
+				size="medium"
```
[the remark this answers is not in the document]
micgro42: Fixed. Alignment looks now OK in Chrome, but still misaligned in Firefox. I'll have another look there.

micgro42:
I added more commits to this PR for the sake of code review. In the end, this PR should only be a single commit.

vue-components/src/components/Message.vue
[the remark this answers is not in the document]
micgro42: The problem seems to be that Firefox aligns the font differently within the line. For a `line-height` of `24px` and a `font-size` of 16px, meaning a capital letter has a size of `12px`, then Firefox aligns that as follows:

4px free space
12px Capital letter "A"
8px free space

image


Chrome (and presumably Safari) aligns it differently:
6px free space
12px Capital letter "A"
6px free space
image

I currently have no idea how to fix this.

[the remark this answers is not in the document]
micgro42: Ah, found the reason. Firefox did render a different font than Chrome. By default, Firefox renders for me `Nimbus Sans` whereas Chrome renders `Liberation Sans`. If I manually force the font in Firefox to be `Liberation Sans` then it aligns as expected:

image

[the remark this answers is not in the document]
micgro42: It does look ok if the Roboto font is installed: 
image

vue-components/src/components/Message.vue
guergana: ok :+1:

vue-components/src/components/Message.vue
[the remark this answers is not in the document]
guergana: Well, I guess this can be looked in more detail in case it gives problems later on. But for now I think it looks pretty good.

guergana (dismissed):
Looks good to me. :+1:

micgro42:
Remaining Todos:
* replace message prop with slot to allow for markup like bold, italic or links
* either make fullwidth the default or add tests and responsive behavior for default width

vue-components/src/components/Message.vue
guergana: after merging the other task we should rebase and reflect the updated values here and in storybook

guergana: Nevermind! There is a ticket for that that I hadn't seen. https://phabricator.wikimedia.org/T265583 This will be addressed in another PR.

guergana (approved):
Looks good!

vue-components/stories/Message.stories.ts
```diff
@@ -0,0 +1,60 @@
...
+				<p><Message type="notice">A neutral message which inform about general higher priority notices to the user. Stay neutral!</Message></p>
+				<p><Message type="success">A success message which means that the user did everything right and should know about it. Yay!</Message></p>
+				<p><Message fullwidth="true" type="success">A success message which means that the user did everything right and should know about it. Yay!</Message></p>
+			</div>
+		`,
+	};
+}
+
+
+export function controllableMessage(args): Component {
```
itamargiv, line 26: Currently, due to a bug I'm trying to resolve [T265039](https://phabricator.wikimedia.org/T265039), type checking isn't working. However, I think we need to add a type here.

```suggestion
export function controllableMessage( args: object ): Component {
```

itamargiv (changes requested):
Looking good, there is one particular note about the sass calculation, that I'd like to get confirmation from @SaiSan-WMDE about.

vue-components/stories/Message.stories.ts
micgro42: But shouldn't that have failed at _some point_ then? Even the job that did build the storybook for Chromatic is green?

But I can update it with its actual type.

itamargiv: like I said, it currently doesn't fail due to a bug I am working on.

micgro42: Ah sorry, I've misread that ticket as saying that it fails only in the build process.

vue-components/src/components/Message.vue
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	&--success {
+		border-color: $wikit-Message-success-border-color;
+		background: $wikit-Message-success-background-color;
+	}
+
+	&--default-width {
+		/* to keep the desired width dimension of x-large when the borders are added */
+		width: ($wikit-Message-default-width - (2 * $wikit-Message-border-width));
```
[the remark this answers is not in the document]
SaiSan-WMDE: I'm now realizing that indeed we should have used `box-sizing: border-box` to achieve this. That's how the text input width was approached.

vue-components/stories/Message.stories.ts
micgro42: I think that should now be fixed :)

vue-components/src/components/Message.vue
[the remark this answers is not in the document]
micgro42: I think that should now be fixed :)

itamargiv (approved):
:100: Above and beyond. Thank you for the great work.

Error dialog box that reads "operation completed successfully"

PR #227, opened 2020-10-19
Add bump-version script

itamargiv:
Adds a script to manage versioning when making new releases. This prevents lerna from pushing incorrect tags automatically, and gives developers more control on what is committed with a version bump change.


jakobw (approved):
Wooo, this `--no-git-tag-version` is definitely a useful discovery! Thanks, looks good to me! :+1:

tzhelyazkova (approved):
Nice!

PR #228, opened 2020-10-20
TextInput: use <label> for the label

jakobw:
Wrapping the label around the input element became questionable as soo
as we also included the validation message in the TextInput component.
We originally went with the label as the root element of the component
to work around having to generate an id for it.

According to the WCAG having the `for`-attribute is also beneficial for
screenreaders:
https://www.w3.org/TR/2008/WD-WCAG20-TECHS-20080430/H44.html

### DIFF
--- a/vue-components/src/components/TextInput.vue
+++ b/vue-components/src/components/TextInput.vue
@@ -1,7 +1,8 @@
 <template>
-	<label :class="[ 'wikit', 'wikit-TextInput', `wikit-TextInput--${width}` ]">
-		<div class="wikit-TextInput__label">{{ label }}</div>
+	<div :class="[ 'wikit', 'wikit-TextInput', `wikit-TextInput--${width}` ]">
+		<label class="wikit-TextInput__label" :for="id">{{ label }}</label>
 		<Input
+			:id="id"
 			:value="value"
 			@input="emitInputEvent"
 			:feedback-type="feedbackType"
@@ -13,7 +14,7 @@
 			:type="error.type"
 			:message="error.message"
 		/>
-	</label>
+	</div>
 </template>
 
 <script lang="ts">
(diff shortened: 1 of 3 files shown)


vue-components/src/components/TextInput.vue
```diff
@@ -64,6 +65,13 @@ export default Vue.extend( {
 		},
 	},
 
+	data() {
+		return {
+			// https://github.com/vuejs/vue/issues/5886
+			id: `wikit-TextInput-${Math.floor( Math.random() * 1000000 )}`,
```
itamargiv, line 71: As I already foresee us generating an Id for the lookup input as well (and potentially radios and checkboxes in the future), maybe we should abstract this into a helper?

itamargiv (commented):
Thank you for reading my mind on the accessibility topic. It's become so engrained in me to prefer the `for` attribute over wrapping inputs in labels that I completely forgot that the latter was even a valid option. Left one comment, but non blocking.

vue-components/src/components/TextInput.vue
jakobw: Yup, makes perfect sense to me given we already want to use it in the Lookup component! Done!

vue-components/src/components/util/generateUid.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Generates an ID that can be used within components, e.g. for binding an input id to a label's `for` attribute
+ *
+ * @param {string} componentName
```
jakobw, line 4: Our eslint config forced me to add `@param` and `@return` with type annotations here. This doesn't make much sense to me since we're using TS. Thoughts on disabling `jsdoc/require-param` and `jsdoc/require-returns`?

itamargiv: I didn't even know we had those enabled... I'm good either / or

guergana (approved):
Looks good to me. :+1:  Maybe after [resolving this question ](https://github.com/wmde/wikit/pull/228#discussion_r508434873)

PR #229, opened 2020-10-21
Lookup behavior

jakobw:
This implements the lookup behavior by exposing the Lookup's internal input value through a `.sync`able prop (with a corresponding event), and making the selected item bindable via `v-model`. Naming aside, whether `v-model` binds the internal `<input>` value or the selected item seems to be the choice that has to be made when developing a Lookup component. The latter approach made more sense to me, as it seems more intuitive that the value of such a component is what is selected, not what is typed into the search field.

Bug: [T265012](https://phabricator.wikimedia.org/T265012)

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -3,8 +3,8 @@
 		<label class="wikit-Lookup__label" :for="inputId">{{ label }}</label>
 		<Input
 			:id="inputId"
-			v-model="inputValue"
-			@input="hasItemSelected = false"
+			:value="searchInput"
+			@input="onInput"
 			@focus.native="focused = true"
 			@blur.native="focused = false"
 			:feedback-type="feedbackType"
@@ -15,7 +15,7 @@
 			class="wikit-Lookup__menu"
 			:menu-items="menuItems"
 			v-if="showsMenu"
-			@selected="hasItemSelected = true"
+			@select="onSelect"
 		/>
 		<ValidationMessage
 			v-if="error"
@@ -31,6 +31,7 @@ import ValidationMessage from './ValidationMessage.vue';
 import Input from './Input.vue';
 import LookupMenu from './LookupMenu.vue';
 import generateUid from '@/components/util/generateUid';
+import { MenuItem } from '@/components/MenuItem';
 
 /**
  * The lookup component is a text input field that provides matching selectable suggestions as a user types into it.
@@ -44,11 +45,13 @@ export default Vue.extend( {
 		return {
 			hasItemSelected: false,
 			focused: false,
-			inputValue: '',
 			inputId: generateUid( 'wikit-Lookup' ),
 		};
 	},
 	props: {
+		/**
+		 * Array of objects that will be displayed in the lookup menu. Must contain a `label` and a `description` field.
+		 */
 		menuItems: {
 			type: Array,
 			default: (): [] => [],
@@ -75,9 +78,15 @@ export default Vue.extend( {
 			type: String,
 			default: '',
 		},
+		/**
+		 * The selected menu item, can be of type `MenuItem` or `null`.
+		 *
+		 * The data usually comes from the consumer's `v-model` annotation on the Lookup component.
+		 * Currently this prop is not used and does nothing.
+		 */
 		value: {
-			type: String,
-			default: '',
+			type: Object,
+			default: null,
 		},
 		width: {
 			type: String,
@@ -85,6 +94,41 @@ export default Vue.extend( {
 				return [ 'small', 'medium', 'large', 'full-width' ].includes( value );
 			},
 			default: 'medium',
+		},
+		/**
+		 * Sets the value of the Lookup component's inner `<input>` element. This prop can be used with the `.sync`
+		 * modifier. When bound to a field in the consuming component's data object, it can be used within a watcher or
+		 * computed property to dynamically update the Lookup's `menuItems` prop.
+		 */
+		searchInput: {
+			type: String,
+			default: '',
+		},
+	},
+
+	methods: {
+		onInput( value: string ): void {
+			this.hasItemSelected = false;
+
+			// the following comment generates the event's description for the docs tab in storybook
+			/**
+			 * Enables the `searchInput` prop to be used with the `.sync` modifier. It's used to transport the value of
+			 * the Lookup component's inner `<input>` element to the parent component.
+			 */
+			this.$emit( 'update:searchInput', value );
+		},
+
+		onSelect( menuItem: MenuItem ): void {
+			this.hasItemSelected = true;
+
+			// the following comment generates the event's description for the docs tab in storybook
+			/**
+			 * This even is emitted whenever an item is selected on the Lookup. The event payload contains the whole
+			 * MenuItem object.
+			 */
+			this.$emit( 'input', menuItem );
+			this.$el.querySelector( 'input' ).blur();
+			this.$emit( 'update:searchInput', menuItem.label );
 		},
 	},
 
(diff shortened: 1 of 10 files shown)


vue-components/stories/Lookup.stories.ts
```diff
@@ -1,53 +1,81 @@
...
-					},
-				],
+				search: '',
+				selectedItem: null,
 			};
 		},
+
+		computed: {
+			menuItems(): MenuItem[] {
+				return vegetableItems.filter( ( item ) => item.label.includes( this.search ) );
```
itamargiv, line 60: I wonder (as I assume we will be getting the items from an external api) whether we should also include loading indication behaviour here. @SaiSan-WMDE did you all discuss this in the QB hike before?

vue-components/src/components/Lookup.vue
```diff
@@ -86,6 +92,41 @@ export default Vue.extend( {
...
+
+		onSelect( menuItem: MenuItem ): void {
+			this.hasItemSelected = true;
+
+			// the following comment generates the event's description for the docs tab in storybook
+			/**
+			 * This even is emitted whenever an item is selected on the Lookup. The event payload contains the whole
+			 * MenuItem object.
+			 */
+			this.$emit( 'input', menuItem );
```
itamargiv, line 126: Fantastic choice! :sunglasses: we shouldn't care whether the implementer of this components has a `value` key or not....

vue-components/src/components/Lookup.vue
```diff
@@ -86,6 +92,41 @@ export default Vue.extend( {
...
+		 * computed property to dynamically update the Lookup's `menuItems` prop.
+		 */
+		searchInput: {
+			type: String,
+			default: '',
+		},
+	},
+
+	methods: {
+		onInput( value: string ): void {
```
itamargiv, line 107: see comment about loading indication behaviour, vuetify does this with a [`loading` prop](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/mixins/loadable/index.ts), example: https://vuetifyjs.com/en/components/autocompletes/#cryptocurrency-selector (not sure if that's the way to go though, it's really a design question).

vue-components/stories/Lookup.stories.ts
itamargiv: Another point to think about is whether the intent was that there would be filtering capabilities after the retrieved items have been cached ( i.e. filter on an already existing list), and if that filtering behaviour doesn't actually belong within the lookup component itself. (I'm sure you probably discussed this before, but I'm not fully caught up on this)

itamargiv (commented):
Great work @jakobw especially like how we handle the select event. I do have some ponderings about the functionality of this component in terms of filtering and handling asynchronicity.

jakobw:
<sup>(throwing everything into one comment because the answers to the two main points are related. hope that's ok)</sup>

Thanks, @itamargiv for the great comments :pray: 
I agree that we need to think about adding a loading indicator as soon as we load data asynchronously. What exactly this should look like isn't specified anywhere yet as far as I know but @SaiSan-WMDE already has it in the back of her mind if I remember correctly.

As to how the component knows that it's loading, I think this approach here is in line with that of most the other component libraries' lookup equivalents ([buefy](https://buefy.org/documentation/autocomplete/#async-with-custom-template), [react material ui](https://material-ui.com/components/autocomplete/#asynchronous-requests), [vuetify](https://vuetifyjs.com/en/components/autocompletes/#cryptocurrency-selector) all have some sort of `loading` prop) which leaves the responsibility of how items are loaded entirely up to the code using the Lookup. With the code already being opinionated about this, I think the loading indicator can be added in a similar way at a later point (but soon!) with a non-breaking change.

I'm curious to hear more about building filtering and caching capabilities into the Lookup. Can this be consolidated with the idea of having the parent component take care of data fetching? Can you give a concrete example for what you'd expect the component to behave like in such a case?

SaiSan-WMDE:
> I agree that we need to think about adding a loading indicator as soon as we load data asynchronously. What exactly this should look like isn't specified anywhere yet as far as I know but @SaiSan-WMDE already has it in the back of her mind if I remember correctly.

That's exactly correct, Jakob! Specifying the loading state for dropdown menus is on my to-do list. I have to say, though, that I'm a bit blocked and unsure regarding how and how much of the loader component can/should be specified using tokens (i.e. what to do with all those animation properties?). We should, in principle, use the "three (bouncing) dots" loader in the small (12px) size. There's an [existing implementation](https://codepen.io/Volker_E/pen/yqNXMe) of it, created by Volker.

Maybe checking the existing component and specs together could be helpful @itamargiv, @jakobw 🙏

itamargiv:
> I'm curious to hear more about building filtering and caching capabilities into the Lookup. Can this be consolidated with the idea of having the parent component take care of data fetching? Can you give a concrete example for what you'd expect the component to behave like in such a case?

Well first of all, I might have linked to the wrong example before: https://vuetifyjs.com/en/components/autocompletes/#asynchronous-items

As you can see, the component itself employs a filtration mechanism to the menu items passed as props, so that filtration operates "offline" as well. this filtration can be overridden by the [`filter` prop](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VAutocomplete/VAutocomplete.ts#L40)

As for caching, this is an interesting idea I saw in vuetify and elsewhere, where the component itself keeps items after they are fetched. In vuetify it is enabled by a `cache-items` boolean prop that is inherited from the [`<v-select>` component](https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/components/VSelect/VSelect.ts#L258-L272).

jakobw:
Get it, thanks! The `filter` prop looks to me like a convenience feature catering to a very specific use case, but I agree that we would want to build something like `cache-items` into the component to avoid repeatedly implementing it in consuming code when loading data over the network (or some other expensive operation). I'm not necessarily seeing it as a required feature for the initial version of the Lookup component, though.

itamargiv:
> The `filter` prop looks to me like a convenience feature catering to a very specific use case,

I'm not suggesting that we add the prop (or anything else for that matter), I was just pondering whether that was the intended use case, as the base functionality I know for autocomplete components like this is that it is intended for filtering from a pre-existing list first (and the asynchronous completion usecase is considered the "specific use case"). But if you say this component doesn't need to have any filtering functionality at all, and it is meant to be exclusively used with filtering functionality that is externalized to it, then sure, I agree.

vue-components/src/components/Lookup.vue
```diff
@@ -75,9 +78,12 @@ export default Vue.extend( {
 			type: String,
 			default: '',
 		},
+		/**
+		 * The selected menu item, can be of type `MenuItem` or `null`.
+		 */
 		value: {
```
micgro42, line 84: Why is this a prop? Also, it seems to be unused?

guergana: Agree, doesn't seem to be used. It is most likely a leftover from using the TextInput component as a boilerplate.

micgro42:
I rebased the branch on top of master and added the extended comment to the `value` prop.

vue-components/src/components/Lookup.vue
guergana: "Currently this prop is not used and does nothing." :( Hahah, poor prop. 

Looks good. Will approve.

guergana (approved):
Looks good. I think we went through all the topics @itamargiv had questions about.

guergana:
Just realized the [keyboard navigation PR](https://github.com/wmde/wikit/pull/241) that is based on this one is not merged.

micgro42:
Let's merge this as it is. As I learned, PRs that are targeting this branch should then be automatically retargeted to master by GitHub. And this branch is already quite large.

PR #230, opened 2020-10-21
Add icons for success/checkmark and notice/info

micgro42:
Bug: [T265582](https://phabricator.wikimedia.org/T265582)

### DIFF
--- a/vue-components/stories/Icon.stories.ts
+++ b/vue-components/stories/Icon.stories.ts
@@ -7,7 +7,7 @@ export default {
 	title: 'Icon',
 };
 
-export function all(): Component {
+export function allTypes(): Component {
 	return {
 		data(): object {
 			return { iconTypes };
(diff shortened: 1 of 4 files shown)


vue-components/stories/Icon.stories.ts
```diff
@@ -13,10 +15,15 @@ export function all(): Component {
...
+		props: {
+			colour: {
+				default: select('Color', iconColors, iconColors[0]),
+			}
+		},
 		template: `
 			<div>
 				<div v-for="type in iconTypes" style="float: left; padding: 20px 50px; color: #555">
-					<Icon :type="type" />
+					<Icon :type="type" :color="colour" />
```
itamargiv, line 26: Nitpick: I find this a bit of a dissonance, to switch from American to British spelling like that.

guergana (approved):
looks great!

vue-components/stories/Icon.stories.ts
micgro42: Yeah, either rename it or use the controls feature instead knobs as this is the first component in this storybook that actually uses knobs.

micgro42: Looking at the comments in the Button story, we probably want neither Knobs nor Controls here. I'll adjust this PR tomorrow so that it shows all variants at the same time (or maybe in multiple stories).

But if Controls are interesting for other Storybooks: That branch shows how to make them happen for the Icon story: https://github.com/wmde/wikit/tree/storybookControls

itamargiv: :+1: Thanks for POC, this is super helpful!

micgro42: There are now multiple stories for the Icon component. One for each prop (type, color, size).

itamargiv (approved):
LGTM, thank you!

PR #231, opened 2020-10-22
Use latest version of wikimedi-ui-base

SaiSan-WMDE:
Makes WiKit to use the latest version of wikimedia-ui-base, [0.17.0](https://www.npmjs.com/package/wikimedia-ui-base). The values of global tokens have been replaced by cherry-picked variables whenever possible.

Bug: https://phabricator.wikimedia.org/T263722

### DIFF
--- a/docs/src/00-doc/05-decisions-and-ADRs/adr/0012-remove-widths.stories.mdx
+++ b/docs/src/00-doc/05-decisions-and-ADRs/adr/0012-remove-widths.stories.mdx
@@ -1,0 +1,38 @@
+import { Meta } from '@storybook/addon-docs/blocks';
+
+<Meta title="Documentation/Decisions and ADRs/ADRs/0012) Remove predefined widths" />
+
+# 12) Remove predefined component widths
+
+Date: 2020-11-10
+
+## Status
+
+Proposed
+
+## Context
+
+Originally, block-level system components – those which size is not defined by their content (e.g. input field) – were designed and implemented following a set of predefined default widths based on the system's spacing unit (the most common size being 256px, for example). 
+These components were also designed and implemented to take a "full-width" value as a prop, intended to make them responsive and adjustable to their context via their containers.
+
+While these predefined widths were purely provided as orientation, we inferred that they might be unused after all due to the variability of contexts in which components can be implemented. 
+Making components full-width by default would simplify things: making them more versatile, responsive and adjustable to the layout of their application. 
+
+## Considered actions
+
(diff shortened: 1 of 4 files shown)


wiese:
This is seriously cool seeing it come from a designer!

wiese:
I see a problem, not specific to this patch, with respect to our verification capabilities for changes like this at the moment (fingers crossed I just don't have the full picture).

* at some point towards the end of the initial DS hike (#141) we changed our chromatic builds to involve the vue-components only (not the token presenters in docs)
* a bit later (#174) we changed the workflow to now only build chromatic for changes in the components' source code

This leaves us in a situation where for token changes 
* there is no preview for them - neither in token presenters in docs (I think that's what https://phabricator.wikimedia.org/T261650 tracks), nor in practice in vue-components
* visual regression tests are not run

Taking this change as motivation to overcome this situation would be a good idea in my eyes, otherwise we are relying on informal verification ("tried it on my machine"), not leveraging the great capabilities we identified as helpful already.

SaiSan-WMDE:
> * there is no preview for them - neither in token presenters in docs (I think that's what https://phabricator.wikimedia.org/T261650 tracks), nor in practice in vue-components
> * visual regression tests are not run
> 
> Taking this change as motivation to overcome this situation would be a good idea in my eyes, otherwise we are relying on informal verification ("tried it on my machine"), not leveraging the great capabilities we identified as helpful already.

That is an accurate depiction of the current situation, and I agree. So far it has been easy to check and trace the outcome of each one of the token changes locally, but the more components are added, the riskier it feels not expanding those capabilities to run end-to-end tests. We'll have this comment in mind when prioritizing that DS task (T261650) with @itamargiv. I consider fair to block this patch.

guergana:
> > * there is no preview for them - neither in token presenters in docs (I think that's what https://phabricator.wikimedia.org/T261650 tracks), nor in practice in vue-components
> > * visual regression tests are not run
> > 
> > Taking this change as motivation to overcome this situation would be a good idea in my eyes, otherwise we are relying on informal verification ("tried it on my machine"), not leveraging the great capabilities we identified as helpful already.
> 
> That is an accurate depiction of the current situation, and I agree. So far it has been easy to check and trace the outcome of each one of the token changes locally, but the more components are added, the riskier it feels to not expand those capabilities to run end-to-end tests. We'll have this comment in mind when prioritizing that DS task (T261650) with @itamargiv. I consider fair to block this patch.

Should we put in the draft state then???

PR #232, opened 2020-10-23
Lookup: add "no match found" behavior

jakobw:
We implemented the "no match found" text with a slot as opposed to a
prop anticipating future requirements to also show markup and/or links
when no results were found.

Bug: [T265010](https://phabricator.wikimedia.org/T265010)

### DIFF
--- a/vue-components/src/components/LookupMenu.vue
+++ b/vue-components/src/components/LookupMenu.vue
@@ -69,6 +69,7 @@ $base: '.wikit-LookupMenu';
 	border: $wikit-LookupMenu-border-width $wikit-LookupMenu-border-style $wikit-LookupMenu-border-color;
 	box-shadow: $wikit-LookupMenu-box-shadow;
 	overflow-y: auto;
+	box-sizing: border-box;
 
 	&__item {
 		padding-block: $wikit-LookupMenu-item-padding-vertical;
--- a/vue-components/src/components/LookupMenu.vue
+++ b/vue-components/src/components/LookupMenu.vue
@@ -15,6 +15,9 @@
 				{{ menuItem.description }}
 			</div>
 		</div>
+		<div v-if="menuItems.length === 0" class="wikit-LookupMenu__no-results">
+			<slot name="no-results" />
+		</div>
 	</div>
 </template>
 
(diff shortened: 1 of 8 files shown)


vue-components/src/components/LookupMenu.vue
```diff
@@ -15,6 +15,9 @@
 				{{ menuItem.description }}
 			</div>
 		</div>
+		<div v-if="menuItems.length === 0" class="wikit-LookupMenu__no-results">
```
itamargiv, line 18: So from this I gather that the default state of the component shows items (even before an initial lookup started), right? Otherwise it would show this slot as the message initially.

itamargiv (commented):
Great work, awesome test coverage :100: I have one question about desired behaviour for an initially empty list. See below:

vue-components/src/components/LookupMenu.vue
micgro42: Mh, the menu should probably only show after we received the first response to our auto-complete query.

micgro42: (Though that's something that has to be implemented in the Lookup component, not in the LookupMenu)

vue-components/tests/unit/components/LookupMenu.spec.ts
```diff
@@ -20,4 +20,34 @@ describe( 'LookupMenu', () => {
...
+			},
+			slots: {
+				'no-results': noResultsFoundText,
+			},
+		} );
+
+		expect( wrapper.text() ).toBe( noResultsFoundText );
+	} );
+
+	it( 'doest not show the "no match found" text if there are menu items', () => {
```
micgro42, line 38: typo:

```suggestion
	it( 'does not show the "no match found" text if there are menu items', () => {
```

vue-components/src/components/LookupMenu.vue
micgro42: Mh, thinking more about this: the prop `menuItems` on `Lookup` should probably default to `null` instead of an empty list to be able to differentiate between no auto-complete response yet and an empty response.

jakobw: As far as I can see this problem goes away as soon as we have way to indicate the "loading" state. What do you think about building that into the Lookup in a forward-compatible way instead of modelling this initial lookup state?

micgro42: I didn't see anything about a loading state in the design, but if we were to add such a loading state, that could resolve this issue as well. Though we have to be careful about race conditions when implementing it.

That being said, I consider all this to be material for incremental improvements. The current code would be fine for me for the initial version of the component.

Not sure what @itamargiv thinks about this :)

vue-components/tests/unit/components/LookupMenu.spec.ts
jakobw: Fixed, thanks!

jakobw:
Updated to fix the typo and added `Co-authored-by` to mention @guergana in the commits.

itamargiv (approved):

PR #233, opened 2020-10-23
Use correct icons for message types

micgro42:
The extra icons were added in #230 (cb1bcb200757564c4ee87a5c115603fa4e3377c7) for [T265582](https://phabricator.wikimedia.org/T265582), so they can now be used in this Message component.

Bug: [T265583](https://phabricator.wikimedia.org/T265583)

PR #235, opened 2020-10-23
Use TypeScript enums for icon types

micgro42:
An enum seems to be the appropriate data structure for enumerating the valid types of an Icon.

This PR sits on top of #233 to better showcase how those Enums would be used, but it is primarily intended for discussion.

What do you all think? Does using the `enum` data-structure more (like in [Bridge](https://github.com/wikimedia/mediawiki-extensions-Wikibase/tree/master/client/data-bridge/src/definitions)) make sense to you?

### DIFF
--- a/vue-components/src/components/Icon.vue
+++ b/vue-components/src/components/Icon.vue
@@ -7,7 +7,7 @@
 			viewBox="0 0 16 16"
 			fill="none"
 			xmlns="http://www.w3.org/2000/svg"
-			v-if="type === 'error'"
+			v-if="type === IconTypes.ERROR"
 		>
 			<path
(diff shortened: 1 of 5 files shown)


itamargiv:
To me yes, definitely,  even if only to reduce the amount of string touchpoints in the code.

guergana:
> To me yes, definitely, even if only to reduce the amount of string touchpoints in the code.

Agree with @itamargiv

micgro42:
Rebased it on current master. Let's hope all the typescript is actually valid :crossed_fingers:

guergana (approved):
Looks good!!

PR #236, opened 2020-10-23
Lookup browser test

guergana:
created browser tests that check if

+ the menu appears on user input
+ no matches found text is displayed
+ selection indicator text is visible after selecting menu-item

What's not included:

+ test of the keyboard navigation. This could be added once as a separate ticket once the keyboard navigation task is completed.


jakobw (approved):
These are some good lookin' tests :100: 

Not sure if any of the code actually changed after the tests failed the last two times. If not, make sure to only merge this if they consistently pass now. Maybe re-run them a couple of times before merging. Few things are worse than flaky browser tests :D

PR #238, opened 2020-10-26
docs: list/visualize component-level tokens

wiese:
Gets the tokens from the tokens' module export, filters it to
component level and groups them.

The resulting stories are a function of the existing components (i.e.
dynamic) which makes things much harder than one would think it needs to
be. Find a way to dynamically create navigation entries, use our custom
components in the mark up.

Bug: https://phabricator.wikimedia.org/T260425


itamargiv:
Closing this PR for now, as this issue is not currently a high priority. Will reopen once we get to tackle this task mentioned in the description, if still relevant.

PR #241, opened 2020-10-26
Keyboard navigation for selecting an item

bereket-WMDE:
Bug: T265884

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -5,18 +5,27 @@
 			:id="inputId"
 			:value="searchInput"
 			@input="onInput"
-			@focus.native="focused = true"
-			@blur.native="focused = false"
+			@focus.native="onFocus"
+			@blur.native="showMenu = false"
 			:feedback-type="feedbackType"
 			:placeholder="placeholder"
 			:disabled="disabled"
+			@keydown.up.native.prevent="onArrowUp"
+			@keydown.down.native.prevent="onArrowDown"
+			@keyup.enter.native="onEnter(); hasActiveElement = false"
+			@keydown.tab.native="onTab"
+			@keydown.enter.native.prevent="hasActiveElement = true"
+			@keyup.esc.native="onEsc"
+			autocomplete="off"
 		/>
 		<LookupMenu
 			class="wikit-Lookup__menu"
 			:menu-items="menuItems"
-			v-if="showsMenu"
+			v-if="showMenu"
 			@select="onSelect"
 			@scroll="onScroll"
+			:selected-item-index="selectedItemIndex"
+			:has-active-element="hasActiveElement"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
@@ -31,7 +40,8 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
+import Vue, { PropType } from 'vue';
+import isEqual from 'lodash.isequal';
 import ValidationMessage from './ValidationMessage.vue';
 import Input from './Input.vue';
 import LookupMenu from './LookupMenu.vue';
@@ -48,11 +58,12 @@ export default Vue.extend( {
 	name: 'Lookup',
 	data() {
 		return {
-			hasItemSelected: false,
-			focused: false,
+			showMenu: false,
 			inputId: generateUid( 'wikit-Lookup' ),
 			scrollIndexStart: null as ( number | null ),
 			scrollIndexEnd: null as ( number | null ),
+			selectedItemIndex: -1,
+			hasActiveElement: false,
 		};
 	},
 	props: {
@@ -60,7 +71,7 @@ export default Vue.extend( {
 		 * Array of objects that will be displayed in the lookup menu. Must contain a `label` and a `description` field.
 		 */
 		menuItems: {
-			type: Array,
+			type: Array as PropType<MenuItem[]>,
 			default: (): [] => [],
 		},
 		error: {
@@ -89,7 +100,6 @@ export default Vue.extend( {
 		 * The selected menu item, can be of type `MenuItem` or `null`.
 		 *
 		 * The data usually comes from the consumer's `v-model` annotation on the Lookup component.
-		 * Currently this prop is not used and does nothing.
 		 */
 		value: {
 			type: Object,
@@ -114,8 +124,12 @@ export default Vue.extend( {
 	},
 
 	methods: {
+		canShowMenu( currentSearchInput: string ): boolean {
+			return currentSearchInput.length > 0;
+		},
 		onInput( value: string ): void {
-			this.hasItemSelected = false;
+			this.selectedItemIndex = -1;
+			this.showMenu = this.canShowMenu( value );
 
 			// the following comment generates the event's description for the docs tab in storybook
 			/**
@@ -123,22 +137,54 @@ export default Vue.extend( {
 			 * the Lookup component's inner `<input>` element to the parent component.
 			 */
 			this.$emit( 'update:searchInput', value );
+			this.$emit( 'input', null );
 		},
 
 		onSelect( menuItem: MenuItem ): void {
-			this.hasItemSelected = true;
+			this.selectedItemIndex = -1;
+			this.showMenu = false;
 
 			// the following comment generates the event's description for the docs tab in storybook
 			/**
-			 * This even is emitted whenever an item is selected on the Lookup. The event payload contains the whole
-			 * MenuItem object.
+			 * This event is emitted whenever an item is selected on the Lookup. The event payload contains the whole
+			 * MenuItem object. The payload is null when no item is selected or the item is deselected.
 			 */
 			this.$emit( 'input', menuItem );
 			this.$emit( 'update:searchInput', menuItem.label );
+		},
 
-			// We know that there is one input here because it is part of this component
-			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-			this.$el.querySelector( 'input' )!.blur();
+		onEnter(): void {
+			if ( this.selectedItemIndex !== -1 ) {
+				this.onSelect( this.menuItems[ this.selectedItemIndex ] );
+			}
+		},
+
+		onArrowUp(): void {
+			this.selectedItemIndex = Math.max( 0, this.selectedItemIndex - 1 );
+		},
+		onArrowDown(): void {
+			this.selectedItemIndex = Math.min( this.menuItems.length - 1, this.selectedItemIndex + 1 );
+		},
+		onTab(): void {
+			if ( this.selectedItemIndex !== -1 ) {
+				this.onSelect( this.menuItems[ this.selectedItemIndex ] );
+			}
+		},
+		onFocus(): void {
+			if ( this.canShowMenu( this.searchInput ) ) {
+				this.showMenu = true;
+			}
+
+			if ( this.value !== null && this.menuItems.length > 0 ) {
+				this.selectedItemIndex = this.menuItems.findIndex(
+					( menuItem ) => { return isEqual( menuItem, this.value ); },
+					this,
+				);
+			}
+		},
+		onEsc(): void {
+			this.showMenu = false;
+			this.selectedItemIndex = -1;
 		},
 		onScroll( firstIndex: number, lastIndex: number ): void {
 			if ( firstIndex !== this.scrollIndexStart || lastIndex !== this.scrollIndexEnd ) {
@@ -159,10 +205,6 @@ export default Vue.extend( {
 	computed: {
 		feedbackType(): string | null {
 			return this.error && this.error.type || null;
-		},
-
-		showsMenu(): boolean {
-			return !this.hasItemSelected && this.focused && this.searchInput.length > 0;
 		},
 	},
 
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -12,9 +12,8 @@
 			:disabled="disabled"
 			@keydown.up.native.prevent="onArrowUp"
 			@keydown.down.native.prevent="onArrowDown"
-			@keyup.enter.native="onEnter(); hasActiveElement = false"
+			@keyup.enter.native="onEnter"
 			@keydown.tab.native="onTab"
-			@keydown.enter.native.prevent="hasActiveElement = true"
 			@keyup.esc.native="onEsc"
 			autocomplete="off"
 		/>
@@ -25,7 +24,6 @@
 			@select="onSelect"
 			@scroll="onScroll"
 			:selected-item-index="selectedItemIndex"
-			:has-active-element="hasActiveElement"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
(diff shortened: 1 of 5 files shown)


bereket-WMDE:
still need to write tests

Incabell:
Hey @bereket-WMDE i had a look and these are the things that I noticed that don't match the specifications in figma. If you have any questions, please let me know.

* when pressing tab without "hovering" above any option, nothing happens. instead the focus should move on to the next element
* after selecting one of the options, i can not use tab anymore. 
* when moving off focus without having selected a option but having input a text, the component does not change to its error state.
* after having selected a option and having moved off focus, when navigating back to the component via tab or mouse, the menu should open up again, showing the selected option as selected.

* i noticed another behavior that we haven't specified and i want to talk to @SaiSan-WMDE about first. But when typing an input, then navigating to an option, but instead of selecting it one presses ESC so the menu closes but the focus on the component stays, and then pressing tab, the previously hovered item gets selected. my intuition tells me we should not select smth that was hovered at some point but them abandoned by an ESC, but i want to double check with sarai.

micgro42:
>     * after selecting one of the options, i can not use tab anymore.

This is very confusing to me. Why do we want to disable navigating with the tab key?

SaiSan-WMDE:
> * i noticed another behavior that we haven't specified and i want to talk to @SaiSan-WMDE about first. But when typing an input, then navigating to an option, but instead of selecting it one presses ESC so the menu closes but the focus on the component stays, and then pressing tab, the previously hovered item gets selected. my intuition tells me we should not select smth that was hovered at some point but them abandoned by an ESC, but i want to double check with sarai.

I agree and would consider this behavior (the last hovered option of an escaped menu being selected on tab) as unwanted.

Incabell:
> > @Incabell wrote:
> > * after selecting one of the options, i can not use tab anymore.
> 
> This is very confusing to me. Why do we want to disable navigating with the tab key?

@micgro42 i have chosen a very bad way to communicate what i wanted to say here, i'm sorry.
What I was doing is stating the things i found wrong during testing. I should have instead mentioned what I expect.

So what I meant when I wrote "after selecting one of the options, i can not use tab anymore." is that this is an unwanted behavior i encountered and that instead I would expect to be able to continue using tab after having selected one of the options, or actually at all times.

Hope that clears things up.

micgro42:
Thank you @Incabell, that clears it up :)

vue-components/tests/unit/components/Lookup.spec.ts
```diff
@@ -135,15 +148,101 @@ describe( 'Lookup', () => {
...
+
+	it( 'emits `update:searchInput` when the internal input element receives input', () => {
+		const wrapper = mount( Lookup );
+		const userInput = 'potato';
+		wrapper.find( 'input' ).setValue( userInput );
+
+		expect( ( wrapper.emitted( 'update:searchInput' ) )![ 0 ] ).toEqual( [ userInput ] );
+	} );
+
+	it( 'checks if the menu closes on Esc key', () => {
```
micgro42, line 197: A note about the naming of tests here. It should describe what the system under test does, not the test itself. E.g.:

```suggestion
	it( 'closes the menu on Esc key', () => {
```

bereket-WMDE:
@Incabell 
>     * when moving off focus without having selected a option but having input a text, the component does not change to its error state.

this one is out of the scope of this ticket. there is a separate ticket to handle all errors

vue-components/src/components/Lookup.vue
```diff
@@ -47,17 +58,20 @@ export default Vue.extend( {
...
+			hasActiveElement: false,
 		};
 	},
 	props: {
 		/**
 		 * Array of objects that will be displayed in the lookup menu. Must contain a `label` and a `description` field.
 		 */
 		menuItems: {
-			type: Array,
+			type: Array as PropType<MenuItem[]>,
```
micgro42, line 74: Nice!

micgro42 (changes requested):
Nice work!

Mainly a test for the up- and down- keys is missing and I have some minor confused questions

micgro42 (dismissed):

vue-components/src/components/LookupMenu.vue
```diff
@@ -4,9 +4,13 @@
...
 			:key="index"
 			v-for="(menuItem, index) in menuItems"
+			:class="{
+				'wikit-LookupMenu__item--selected': index === selectedItemIndex,
+				'wikit-LookupMenu__item--active': index === selectedItemIndex && hasActiveElement
+			}"
 			@click="$emit( 'select', menuItem )"
 			@mousedown.prevent
-			tabindex="0"
+			ref="lookup"
```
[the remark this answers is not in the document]
micgro42: Doesn't seem to be used.

vue-components/src/components/Lookup.vue
```diff
@@ -123,19 +139,57 @@ export default Vue.extend( {
...
+			if ( this.value !== null && this.menuItems.length > 0 ) {
+				this.selectedItemIndex = this.menuItems.findIndex(
+					( menuItem ) => { return isEqual( menuItem, this.value ); },
+					this,
+				);
+			}
+		},
+		onEsc(): void {
+			this.escKeyPressed = true;
+			this.selectedItemIndex = -1;
```
itamargiv, line 192: For some reason (unsure why yet), the selected item is not reset on <kbd>Esc</kbd>. See screencast: https://drive.google.com/file/d/1zly4scjSPBPIfGK_XhOq9Uw0ZAiEs31_/view?usp=sharing

itamargiv (changes requested):
Looking good :sunglasses:, I have a few comments though (Sorry for dragging this longer) and a couple of potential changes to make.

vue-components/src/components/Lookup.vue
```diff
@@ -123,19 +139,57 @@ export default Vue.extend( {
...
 			this.$emit( 'update:searchInput', menuItem.label );
 
-			// We know that there is one input here because it is part of this component
-			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-			this.$el.querySelector( 'input' )!.blur();
+			// If a menu item is selected using the enter key, keep focus on input component
+			if ( !this.enterKeyPressed ) {
+				// We know that there is one input here because it is part of this component
+				// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+				this.$el.querySelector( 'input' )!.blur();
```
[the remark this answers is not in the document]
micgro42: Mh, good point. However, this was not created in this PR. Would it be ok to fix it in a separate PR?

vue-components/src/components/Lookup.vue
```diff
@@ -123,19 +139,57 @@ export default Vue.extend( {
...
+			}
+		},
+
+		onEnter(): void {
+			if ( this.selectedItemIndex !== -1 ) {
+				this.enterKeyPressed = true;
```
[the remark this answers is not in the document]
micgro42: In `onFoucus()` and `onInput(...)` :)

vue-components/src/components/Lookup.vue
micgro42: Found the reason. The problem is not here, the current code works correctly. The problem is that deselecting a previously selected item by typing again doesn't emit an empty input event to let the consumer know that the item was deselected. Will fix.

vue-components/src/components/Lookup.vue
```diff
@@ -145,7 +199,7 @@ export default Vue.extend( {
 		},
 
 		showsMenu(): boolean {
-			return !this.hasItemSelected && this.focused && this.searchInput.length > 0;
+			return !this.escKeyPressed && !this.enterKeyPressed && this.focused && this.searchInput.length > 0;
```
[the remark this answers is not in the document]
micgro42: After some initial skepticism, I tried it out and I like it! Have a look at the current implementation.

I could not yet get rid of `this.enterKeyPressed` as the blur behavior in `onSelect` differs depending on whether the item was selected with Enter or click. This might need some more thinking.

vue-components/src/components/Lookup.vue
micgro42: Fixed. Should work as expected now.

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
micgro42: It is set to false in `onInput` and `onFocus`

vue-components/src/components/Lookup.vue
```diff
@@ -5,17 +5,28 @@
...
 			:value="searchInput"
 			@input="onInput"
-			@focus.native="focused = true"
+			@focus.native="onFocus"
 			@blur.native="focused = false"
 			:feedback-type="feedbackType"
 			:placeholder="placeholder"
 			:disabled="disabled"
+			@keyup.up.native="onArrowUp"
+			@keyup.down.native="onArrowDown"
```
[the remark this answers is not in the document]
micgro42: Done.

micgro42:
@itamargiv I think I addressed all your points. Could you have another look?

Though I feel I did notice one strange behavior: When I reload the storybook story and select an item with tab for the first time, the focus seems to jump away and back immediately so that the Lookup is select again and single result is shown. It doesn't happen on any further use of the tab key until the story is manually reloaded in the browser. Maybe something with the story is strange? Or maybe it is just my setup.

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
itamargiv: Sure, go for it, thanks!

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
itamargiv: Hmm, I would question the reasoning behind the different modes of behaviour, in all other implementations of an autocomplete I know there's no difference between enter and mouse click. Furthermore, the fact that the element blurs after mouse click interrupts a flow where a user uses both keyboard and mouse navigation as the visual focus signifier is lost on click. @SaiSan-WMDE what do you think?

vue-components/src/components/Lookup.vue
itamargiv: Perfect, thank you!

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
itamargiv: Thank you!

itamargiv:
> Maybe something with the story is strange? Or maybe it is just my setup.

That is odd. Also I wasn't able to reproduce, could you please create a screencast? (It's <kbd>Ctrl</kbd> + <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>R</kbd> in ubuntu)

vue-components/src/components/Lookup.vue
```diff
@@ -111,42 +122,85 @@ export default Vue.extend( {
...
 			 * Enables the `searchInput` prop to be used with the `.sync` modifier. It's used to transport the value of
 			 * the Lookup component's inner `<input>` element to the parent component.
 			 */
 			this.$emit( 'update:searchInput', value );
+			this.$emit( 'input', null );
 		},
 
 		onSelect( menuItem: MenuItem ): void {
-			this.hasItemSelected = true;
+			this.hasActiveElement = true;
```
itamargiv, line 143: While looking at the component together with @SaiSan-WMDE we noticed that once the selection is made, and we navigate away from the component, this `hasActiveElement` does not revert back to `false` which causes the items of the menu to appear active when hovered.

According to @SaiSan-WMDE The active style should only appear when the mouse button is down.

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
SaiSan-WMDE: Hi! I would encourage us to make the component behave the same way whenever a menu item is selected either on mouse down or enter: the selection is made and the focus remains on the input element.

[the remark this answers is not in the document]
micgro42: Mh, I encoded that in https://phabricator.wikimedia.org/T265884#6584975 and I did it because I read it somewhere, but I can't find it in Figma now o.O

I'll change it and then we can drop `this.enterKeyPressed` completely :raised_hands:

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
micgro42: I think this is now obsolete as that entire condition was removed and the input is no longer blurred

micgro42:
Aside from a rebase, it looks now fine to me. But I'll wait a bit in case @itamargiv wants to take another look :)

itamargiv:
Please resolve the conflicts, and the I can have another look

micgro42:
> Please resolve the conflicts, and the I can have another look

The conflicts are resolved. Sorry for not catching this right away

vue-components/src/components/LookupMenu.vue
```diff
@@ -107,12 +118,14 @@ $base: '.wikit-LookupMenu';
...
-		&:hover {
+		&:hover,
+		&--selected {
 			background-color: $wikit-LookupMenu-item-hover-background-color;
 			cursor: $wikit-LookupMenu-item-hover-cursor;
 		}
 
-		&:active {
+		&:active,
+		&--active {
```
itamargiv, line 128: So I think we had a misunderstanding here, I think the blue color we had before should still be included but only on mouse click (`:active`, currently it isn't working), also if all this `hasActiveElement` property does is try to mimic an active mouse mode, then I'm pretty sure we do not need it. I don't think I ever saw an implementation of `:active` styles mimicked through the keyboard enter. But maybe that's what @SaiSan-WMDE intended?

itamargiv (changes requested):
Sorry, it seems as though that the intended styles for :active` are not currently applied (The text does not turn blue on mouse down)

vue-components/src/components/LookupMenu.vue
SaiSan-WMDE: `:active` is working fine for me (checking in Chrome). I've seen implementations of `:active`styles that are persistent on keyboard enter and thought that was the default way to go, but unfortunately I failed to notice that OOUI is not one of them, so I guess we don't need the `hasActiveElement` property in case it's only trying to emulate the active state. Sorry about that

micgro42: Ah, sorry. Then I was misled by the statements that the component should behave (and to me that implied look) the same whether selecting an item on click or on Enter.

itamargiv: Oddly enough, on Firefox it works on Enter down but not on Mouse down :man_shrugging: Works as expected on Firefox. And don't get me wrong, if the intention was to have the mouse down behaviour on enter, then by all means, keep it. I don't think we need to have a 1:1 facsimile of OOUI, I really was just unclear on the design intent.

bereket-WMDE: Based on @SaiSan-WMDE comment. I just removed the whole  hasActiveElement thing. Now the active style is only applied up mouse click. there will be no 'active' style when a user selects using Enter.
@itamargiv There is a separate ticket to address the browser issues that you're currently experiencing. 
see here: https://phabricator.wikimedia.org/T265917

that will be my next task as soon as we're done with this one.

itamargiv: Thank you @bereket-WMDE Totally missed it.

itamargiv (approved):
Finally, it is done! Great work @bereket-WMDE and @micgro42!

Person wildly typing on keyboard

PR #242, opened 2020-10-27
Update @vue/cli-plugin-typescript to v4.5

itamargiv:
After an arduous investigation, it seems that a bad combination of `@storybook/vue` v6+ and `@vue/cli-plugin-typescript` v4+ caused type checking to be skipped or disabled (filed a [bug report in Storybook](https://github.com/storybookjs/storybook/issues/12921)): https://phabricator.wikimedia.org/T265039#6580648.

As we already use Storybook 6 features such as controls, I've opted to update to  v4.5 of `@vue/cli-plugin-typescript` in hopes that it doesn't have additional effects apart from restoring our typechecking.

Bug: [T265039](https://phabricator.wikimedia.org/T265039)


vue-components/package.json
```diff
@@ -48,12 +48,12 @@
     "@types/jest": "^24.0.19",
     "@typescript-eslint/eslint-plugin": "^2.33.0",
     "@typescript-eslint/parser": "^2.33.0",
-    "@vue/cli-plugin-babel": "~4.4.0",
+    "@vue/cli-plugin-babel": "~4.5.0",
```
micgro42, line 51: What is the motivation behind pinning some of those to their minor version (`~`) and some to their major version (`^`)?

micgro42 (commented):
Thank you for tracking this down!

I've just one question about the versioning. I notice that this wasn't introduced with this patch, so this question should not be considered blocking.

vue-components/package.json
itamargiv: The pinning to minor done by default when you run `vue create` (with whatever vue cli version you are using). I assume they do this to avoid peer dependencies falling out of sync, as it seems the plugins and vue cli are all cut from the same source and versioned together.

But now that you mentioned it, I see I accidentally pinned `@vue/cli-plugin-typescript` to major (`^`) instead of minor (`~`), thanks!

micgro42: Good point. But given that point, would it make sense to raise the `@vue/cli-plugin-e2e-nightwatch` dependency to `4.5.x` as well?

itamargiv: Great question, and yes, I will change the entry to 4.5 and up. I'm leaving the pin to major (`^`) As it seems to have been introduced by the installation of this package itself: when you run `vue add e2e-nightwatch` (after boiler-plating your project without it) the package.json entry is added as pinned to major (I agree that this is odd, but I assume the vue cli team have their reasons).

micgro42 (approved):
The code changes make sense to me.

Also, I tried it locally and it does fail with type errors present. :heavy_check_mark: 

What is strange though, is that it fails during `npm install`, but it _passes_ during `npm run test`. Shouldn't it be the other way around?

itamargiv:
> What is strange though, is that it fails during `npm install`, but it _passes_ during `npm run test`. Shouldn't it be the other way around?

I think the key here is that type checking occurs on `build` which is not necessary for `test` (by default) but does occur on `install` (since we set the `prepare` script to run `build` as it's needed in order to ensure everything in the dist directory is included in the npm package).

PR #244, opened 2020-10-27
create dropdown component tokens

SaiSan-WMDE:
This PR contains the component-level tokens that will be used to define the visual properties of our future Dropdown.

### DIFF
--- a/tokens/properties/components/Dropdown.json
+++ b/tokens/properties/components/Dropdown.json
@@ -1,8 +1,5 @@
 {
     "wikit-Dropdown": {
-        "default-width": {
-            "value": "{dimension.width.full.value}"
-        },
         "border-color": {
             "value": "{border.color.base.default.value}"
         },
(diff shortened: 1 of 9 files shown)


micgro42:
Changed the base so that I can base my work on the Dropdown component on top of this PR

tokens/properties/components/Dropdown.json
```diff
@@ -0,0 +1,221 @@
...
+        "border-width": {
+            "value": "{border.width.thin.value}"
+        },
+        "border-radius": {
+            "value": "{border.radius.base.value}"
+        },
+        "background-color": {
+            "value": "{background.color.neutral.default.value}"
+        },
+        "font-family": {
```
micgro42, line 21: While implementing, I noticed that it makes probably more sense to group these font-familiy and line-height styles into a "selectedOption" subsection to better contrast them with the placeholder styles.

tokens/properties/components/Dropdown.json
```diff
@@ -0,0 +1,221 @@
+{
+    "wikit-Dropdown": {
+        "default-width": {
```
micgro42, line 3: I think this can now be dropped with #259 merged

tokens/properties/components/Dropdown.json
```diff
@@ -0,0 +1,221 @@
...
+                "value": "{transition.interaction.field.duration.value}"
+            },
+            "property": {
+                "value": "{transition.interaction.field.property.value}"
+            },
+            "timing-function": {
+                "value": "{transition.timing-function.ease.value}"
+            }
+        },
+        "menu": {
```
micgro42, line 123: Since we are not using a native select after all, but are reusing the LookupMenu, these `menu` tokens are probably no longer needed. Sorry for the extra work :pensive:

tokens/properties/components/Dropdown.json
SaiSan-WMDE: Makes sense! For a moment, I wanted to suggest grouping both (placeholder and selected option) under the same object and call it "dropdown text" or similar, since they share exactly the same styles. What stopped me was 1. the will to keep a certain parallelism between field elements, 2. the thought that maybe their styles could be updated and differ in the future (not so sure about this, though).

tokens/properties/components/Dropdown.json
SaiSan-WMDE: thanks for all the heads-up! this file really needed a review 😄

micgro42:
As far as I can see, these tokens are now complete. Would it be alright if I squashed all these commits into one and then added another commit with the actual component? That would make it easier to chain them after the prerequisite pull requests.

micgro42:
I think this is ready for review and (squash-)merge

itamargiv (dismissed):
LGTM, thank you!

itamargiv (approved):
Looking good, thank you!

PR #246, opened 2020-10-28
Disable false positive "could be null" linting error

micgro42:
I decided against using `this.$el.querySelector( 'input' )?.blur();` as this error is actually a case of a false positive and I wanted to make that obvious and explicit.


guergana:
Looks good! It does solve the error.

guergana (approved):
LGTM!

PR #247, opened 2020-10-28
Add stories for the internal LookupMenu component

micgro42:
This will be used for design verification and visual regression testing.
That seems in particular important for possible upcoming refactoring for
this component.

Behavior has to still be evaluated with the actual component.

### DIFF
--- a/vue-components/stories/Lookup.stories.ts
+++ b/vue-components/stories/Lookup.stories.ts
@@ -4,7 +4,7 @@ import { MenuItem } from '@/components/MenuItem';
 
 export default {
 	component: Lookup,
-	title: 'Lookup',
+	title: '/Lookup/Lookup',
 };
 
 const vegetableItems = [
--- a/vue-components/stories/LookupMenu.stories.ts
+++ b/vue-components/stories/LookupMenu.stories.ts
@@ -1,0 +1,84 @@
+import LookupMenu from '@/components/LookupMenu';
+import { Component } from 'vue';
+import { MenuItem } from '@/components/MenuItem';
+
+export default {
+	component: LookupMenu,
+	title: '/Lookup/LookupMenu',
+};
+
+const vegetableItems = [
+	{
+		label: 'potato',
+		description: 'root vegetable',
+		value: 'Q10998',
+	},
+	{
+		label: 'carrot',
+		description: 'root vegetable, usually orange in color',
+		value: 'Q81',
+	},
+	{
+		label: 'zucchini',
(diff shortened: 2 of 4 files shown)


SaiSan-WMDE:
Thanks for this, @micgro42! I think it does make sense for the menu to be its own component, specially in case reusability is an option. Just a heads-up, though, that a different approach to the classification of subcomponents is being followed here: the LookupMenu is included under an Internal Subcomponents section, while the basic input subcomponent is grouped under a specific component folder. 

I personally think that the LookupMenu subcomponent should live in a Lookup folder with its "parent". The same would apply to all subcomponents. But I'd like to hear everyone's preferences regarding classification, so please feel free to comment if you have one 🙏🏻

micgro42:
Mh, I agree that we should be consistent with how structure those stories. I'll move the story into a `Lookup` directory.

I feel at some point we might want to have another layer of directories in order to make it obvious what is an internal component and what is not (c.f. ValidationMessage), but we can do that later when we have a larger number of components and the need is more present.

SaiSan-WMDE:
Thanks for the changes, @micgro42! Looking good 😎

itamargiv (approved):
Thank you for adding this story, are we completely sure that we will reuse this component?

micgro42:
> Thank you for adding this story, are we completely sure that we will reuse this component?

_Sure_ that we are going to reuse it? No. But probably we'll refactor it and split it up. Maybe we need to somehow modify it for [T266856](https://phabricator.wikimedia.org/T266856).

We realized that it is not possible to style the Dropdown options as extensive as we want to. So they'll need something very similar and that is going include some refactoring that could introduce bugs that might change how this component loks. Therefore, I think it is good to have visual regression tests for it.

SaiSan-WMDE (approved):

guergana:
> > Thank you for adding this story, are we completely sure that we will reuse this component?
> 
> _Sure_ that we are going to reuse it? No. But probably we'll refactor it and split it up. Maybe we need to somehow modify it for [T266856](https://phabricator.wikimedia.org/T266856).
> 
> We realized that it is not possible to style the Dropdown options as extensive as we want to. So they'll need something very similar and that is going include some refactoring that could introduce bugs that might change how this component loks. Therefore, I think it is good to have visual regression tests for it.
> 
> What are your concerns about this?

I think the concern comes from some previous discussions were we decided we would extract this component if it was required in the future as it was required. If it needed too much customization for the dropdown component we would not reuse it to avoid making this component overly complicated, because even if it shares some similarities with the dropdown component, it is also quite different.

PR #248, opened 2020-10-28
define input bg color

SaiSan-WMDE:
Quick fix to apply a white background color to the Input component and prevent transparency. The issue remained unnoticed during component verfication.


micgro42:
Looks good, but I think the changes to `package-lock.json` are not supposed to be in here :)

itamargiv (commented):
Purrfect, thank you!

Painting wall over reclining dog

itamargiv (approved):

PR #249, opened 2020-10-28
Add story for Lookup with error state and warning state

guergana:
This PR has two commits:

+ fix a display error of the menuItem to be above the elements below.
+ Error and Warning states added to main Lookup component Story

T266595

### DIFF
--- a/vue-components/src/components/LookupMenu.vue
+++ b/vue-components/src/components/LookupMenu.vue
@@ -74,6 +74,7 @@ $base: '.wikit-LookupMenu';
 	box-shadow: $wikit-LookupMenu-box-shadow;
 	overflow-y: auto;
 	box-sizing: border-box;
+	z-index: 1;
 
 	&__item {
 		padding-block: $wikit-LookupMenu-item-padding-vertical;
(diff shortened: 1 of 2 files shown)


micgro42 (approved):
I like it. Thank you!

SaiSan-WMDE:
Thanks, Guergana! 💯  Approving the UI tests since everything looks fine, but just to confirm: I'm assuming that it's not necessary (or probably not possible?) to make the error and warning stories display any lookup behaviour (showing the menu) because focusing on any of the validation states actually makes the errors/warnings to go away (they are replaced by the default/focus state of the component). Is that right?

guergana:
> Thanks, @guergana! 100 Approving the UI tests since everything looks fine, but just to confirm: I'm assuming that it's not necessary (or probably not possible?) to make the error and warning stories display any lookup behaviour (showing the menu) because focusing on any of the validation states actually makes the errors/warnings to go away (they are replaced by the default/focus state of the component). The same would apply to TextInput. Is that right?

As far as I understand, yes, what you say is right.

PR #250, opened 2020-10-30
add white-space property to button component

SaiSan-WMDE:
This makes sure our buttons only contain single-line text, which is recommended from a design standpoint.

### DIFF
--- a/vue-components/src/components/Button.vue
+++ b/vue-components/src/components/Button.vue
@@ -54,6 +54,7 @@ $base: '.wikit-Button';
 	transition-duration: $wikit-Button-transition-duration;
 	transition-timing-function: $wikit-Button-transition-timing-function;
 	transition-property: $wikit-Button-transition-property;
+	white-space: nowrap;
 
 	// TODO use breakpoint mixin?
 	@media (max-width: $width-breakpoint-mobile) {


Ladsgroup (approved):
Thanks!

micgro42:
Mh, what is happening when there is more text than what fits into the button, but the button cannot grow in width?

SaiSan-WMDE:
> Mh, what is happening when there is more text than what fits into the button, but the button cannot grow in width?

This would most likely indicate that either the copy (e.g. cta text is too long, translations were not taken into account) or the design need to be fixed. I guess we could always implement this as a prop to cover edge cases, but I worry that we would then be enabling  an "anti-pattern". What do you think? 
I forgot to mention I'm following [WMF indications] here.(https://github.com/wikimedia/wvui/blob/master/src/components/button/Button.vue) here.

micgro42:
The background to my question is that in HTML, text usually doesn't break line out of its own volition. Even if you copy and past text with line-breaks into HTML, it will usually still display as a single line if it has enough space to do so.

With `white-space: nowrap;` you are now adding a declaration that takes effect only in the circumstance that the text does not have enough space to be fully displayed in a single line. Instead of breaking the line it is now forced to overflow the container instead.

I can see the merit in following the WMF Design in principle. Unfortunately, neither their commit message nor their pull request explain why they consider that rule necessary and their comment above it only restates what the rule does and not _why_ it is there.

I found this page helpful to understand how this property works: https://developer.mozilla.org/en-US/docs/Web/CSS/white-space

SaiSan-WMDE:
> With that declaration, instead of breaking the line it is now forced to overflow the container instead.

I  don't think that the button text can really overflow its container, right?

> The background to my question is that in HTML, text usually doesn't break line out of its own volition. Even if you copy and past text with line-breaks into HTML, it will usually still display as a single line if it has enough space to do so.

I believe that the purpose of using `white-space: nowrap` is to prevent the button text from wrapping unexpectedly under some circumstances (like when using flex). I truly though that this was a very common, recommended practice. Seeing it applied by the WMF and [Material design](https://github.com/material-components/material-components-web/issues/4894). But of course, I'm being a bit monkey see monkey do here, and I'm unable to gasp all the consequences 🤔  So I appreciate the push back

micgro42:
> > With that declaration, instead of breaking the line it is now forced to overflow the container instead.
> 
> I don't think that the button text can really overflow its container, right?

You can make it overflow if you give it a `max-width` or such. But I agree that this might be a rather contrived scenario.

> > The background to my question is that in HTML, text usually doesn't break line out of its own volition. Even if you copy and past text with line-breaks into HTML, it will usually still display as a single line if it has enough space to do so.
> 
> I believe that the purpose of using `white-space: nowrap` is to prevent the button text from wrapping unexpectedly under some circumstances (like when using flex). I truly though that this was a very common, recommended practice. Seeing it applied by the WMF and [Material design](https://github.com/material-components/material-components-web/issues/4894).

Mh, that is a good point. I looked around a bit and found somebody having the exact same issue in thoughtbot/bitters#281. The response points to a codepen that showcases the issue that you describe. It seems to occur when the button is paired in a non-wrapping flex-box with another element that also needs to break lines to fit into the available space.

> But of course, I'm being a bit monkey see monkey do here, and I'm unable to gasp all the consequences thinking So I appreciate the push back

This is perfectly fine and true of everybody, we are all always learning. Myself explicitly included as this conversation shows. :relieved:

PR #251, opened 2020-10-30
make block-level components full-width by default

SaiSan-WMDE:
This sets the width prop of the existing block-level components (message, input –  textInput & Lookup) to "full-width" by default. 

Width props have been removed: Implementers would be able to apply flexible (using specific values) or standard (using tokens) widths to block-level components via their wrapping elements.

Story: https://phabricator.wikimedia.org/T265371


vue-components/src/components/Message.vue
```diff
@@ -36,9 +36,12 @@ export default Vue.extend( {
 				return [ 'warning', 'error', 'notice', 'success' ].includes( value );
 			},
 		},
-		fullwidth: {
-			type: Boolean,
-			default: false,
+		width: {
+			type: String,
```
SaiSan-WMDE, line 40: nope. makes sense to keep this a boolean 😅

SaiSan-WMDE:
will start over following different approach

PR #252, opened 2020-11-02
Bump version to 1.2.0-alpha.0

micgro42:
This pre-release allows using the both Message and the Lookup component in other work-in-progress projects while still finishing up the last aspects of the Lookup component.

The Message should already be done.

https://github.com/wmde/wikit/compare/v1.1.2..alphaLookupAndMessageRelease

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@wmde/wikit-docs",
-	"version": "1.1.2",
+	"version": "1.2.0-alpha.0",
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@wmde/wikit-docs",
   "private": true,
-  "version": "1.1.2",
+  "version": "1.2.0-alpha.0",
   "description": "Storybook for illustrating design tokens and components",
   "keywords": [
     "wikit",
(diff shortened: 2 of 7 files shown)


Ladsgroup (approved):
LGTM

itamargiv:
Looks good to me, but we have an issue, as we might not be able to follow the current specified workflow (it turns out npm no longer "hides" pre-release tags since [v3.4.1](https://github.com/npm/npm/releases/tag/v3.4.1). Which means we might need to publish this pre-release from a local computer as I currently don't have time this week to work on defining a new pre release workflow, documented in task [T267094](https://phabricator.wikimedia.org/T267094).

itamargiv (changes requested):
Nothing to change. Blocking this to make sure the comment above is acknowledged.

micgro42:
> Nothing to change. Blocking this to make sure the comment above is acknowledged.

Sorry I was in a lot of meetings and I'm still working on understanding the details here.

With "workflow" you mean an actual GitHub Workflow or an internal WMDE process?

itamargiv:
> With "workflow" you mean an actual GitHub Workflow or an internal WMDE process?

I meant both, but yeah that's too ambiguous.. What I suggest is tat we team up this week and make a one off pre-realease with a dist tag from my local machine, if having this component released is as pressing as it sounds, and then it will give me time to properly think of an organized pre-release workflow (both meanings) for the future. What do you think?

micgro42:
Ah. I think I slightly misread the intention on the phab ticket and already started drafting one of the ADRs as I perceived them to be intended as a prerequisite for this release.

Doing this one-off pre-release this week together sounds good to me :+1:

itamargiv (approved):

PR #253, opened 2020-11-03
Lookup event emitter on scroll

guergana:
+ an event is emitted from Lookup when user scrolls in the LookupMenu that contains the first and last indexes of visible elements
+ added a browser test

### DIFF
--- a/vue-components/package.json
+++ b/vue-components/package.json
@@ -35,6 +35,7 @@
   "dependencies": {
     "@wmde/wikit-tokens": "^1.2.0-alpha.0",
     "core-js": "^3.6.5",
+    "lodash.debounce": "^4.0.8",
     "ress": "^3.0.0",
     "vue": "^2.6.12"
   },
@@ -47,6 +48,7 @@
     "@storybook/addon-viewport": "^6.0.28",
     "@storybook/vue": "^6.0.28",
     "@types/jest": "^24.0.19",
+    "@types/lodash": "^4.14.165",
     "@typescript-eslint/eslint-plugin": "^2.33.0",
     "@typescript-eslint/parser": "^2.33.0",
     "@vue/cli-plugin-babel": "~4.5.0",
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -16,6 +16,7 @@
 			:menu-items="menuItems"
 			v-if="showsMenu"
 			@select="onSelect"
+			@scroll="onScroll"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
@@ -50,6 +51,8 @@ export default Vue.extend( {
 			hasItemSelected: false,
 			focused: false,
 			inputId: generateUid( 'wikit-Lookup' ),
+			scrollIndexStart: null as ( number | null ),
+			scrollIndexEnd: null as ( number | null ),
 		};
 	},
 	props: {
@@ -136,6 +139,20 @@ export default Vue.extend( {
 			// We know that there is one input here because it is part of this component
 			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 			this.$el.querySelector( 'input' )!.blur();
+		},
+		onScroll( firstIndex: number, lastIndex: number ): void {
+			if ( firstIndex !== this.scrollIndexStart || lastIndex !== this.scrollIndexEnd ) {
+				/**
+				 * This event is emitted whenever the first or last index of the
+				 * visible menuItems changes. If the user scrolls but the indexes remain
+				 * unchanged the event won't fire.
+				 *
+				 */
+				this.$emit( 'scroll', firstIndex, lastIndex );
+				this.scrollIndexStart = firstIndex;
+				this.scrollIndexEnd = lastIndex;
+			}
+
 		},
 	},
 
--- a/vue-components/src/components/LookupMenu.vue
+++ b/vue-components/src/components/LookupMenu.vue
@@ -1,5 +1,9 @@
 <template>
-	<div :class="[ 'wikit', 'wikit-LookupMenu' ]">
+	<div
+		:class="[ 'wikit', 'wikit-LookupMenu' ]"
+		@scroll.passive="onScroll"
+		ref="lookup-menu"
+	>
 		<div
 			class="wikit-LookupMenu__item"
 			:key="index"
@@ -7,6 +11,7 @@
 			@click="$emit( 'select', menuItem )"
 			@mousedown.prevent
 			tabindex="0"
+			ref="menu-items"
 		>
 			<div class="wikit-LookupMenu__item__label">
 				{{ menuItem.label }}
@@ -23,6 +28,7 @@
 
 <script lang="ts">
 import Vue from 'vue';
+import debounce from 'lodash/debounce';
 
 /**
  * This is an internal component which used by the Lookup component.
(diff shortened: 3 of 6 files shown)


guergana:
I am aware that the scroll event might need to be throttled. I am not sure about what is the best approach for doing this in vue. Should I use a setTimeout, is 300 a good delay? Making the event passive already improves the performance.

itamargiv:
> I am aware that the scroll event might need to be throttled.

That's great thinking! IMO maybe a debounce would work even better for this kind of task (for anyone interested in more context, see: https://redd.one/blog/debounce-vs-throttle).

> I am not sure about what is the best approach for doing this in vue.

Whichever you chose, it would probably be best to make the chosen function reusable, or even rely on an external lib like lodash (I'm more of a less deps better approach, but it's really a personal choice in this case, as long as the library is maintained and well supported) see [this stack overflow thread](https://stackoverflow.com/questions/42199956/how-to-implement-debounce-in-vue2) for deets.

itamargiv (changes requested):
Blocking until implementation of throttle / debounce.

micgro42:
I think we used a debounce in data bridge. I'll look it up

micgro42:
Here we are using lodash's debounce: 

https://github.com/wikimedia/wikibase-vuejs-components/blob/master/src/components/ResizingTextField.vue

guergana:
Will use debounce then. Thanks for the input @micgro42 and @itamargiv

vue-components/package.json
```diff
@@ -33,8 +33,10 @@
   ],
   "types": "dist/main.d.ts",
   "dependencies": {
+    "@types/lodash": "^4.14.164",
```
micgro42, line 36: These should probably be in the dev dependencies

vue-components/src/components/LookupMenu.vue
```diff
@@ -47,8 +50,26 @@ export default Vue.extend( {
 				rootElem.style.maxHeight = menuHeight + 'px';
 			}
 		},
-	},
+		onScroll: debounce( function ( this: Vue ) {
```
itamargiv, line 53: Looking good so far! I'm not quite sure why we would be declaring a parameter called `this` in this function? I don't think the lodash `debounce`  is smart enough to repass `this` as the first arg (or rather it is smart enough to understand the `this` context on it's own): https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L10400-L10423.

vue-components/src/components/Lookup.vue
```diff
@@ -137,6 +140,15 @@ export default Vue.extend( {
...
 		},
+		onScroll( firstIndex: number, lastIndex: number ): void {
+
+			if ( firstIndex !== this.scrollIndexStart || lastIndex !== this.scrollIndexEnd ) {
+				this.$emit( 'scroll', firstIndex, lastIndex );
+				this.scrollIndexStart = firstIndex;
+				this.scrollIndexEnd = lastIndex;
+			}
+
+		},
```
itamargiv, line 151: Clever bubbling, I like :champagne:

vue-components/stories/Lookup.stories.ts
```diff
@@ -43,6 +43,26 @@ const vegetableItems = [
 		description: 'vegetable',
 		value: 'Q150463',
 	},
+	{
+		label: 'cassava root',
```
itamargiv, line 47: Yums, this is making me hungry... :drooling_face:

vue-components/src/components/Lookup.vue
```diff
@@ -137,6 +140,15 @@ export default Vue.extend( {
 			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 			this.$el.querySelector( 'input' )!.blur();
 		},
+		onScroll( firstIndex: number, lastIndex: number ): void {
+
+			if ( firstIndex !== this.scrollIndexStart || lastIndex !== this.scrollIndexEnd ) {
```
itamargiv, line 145: I think we should add a comment here to describe this event in storybook. See [lines 121-124](https://github.com/wmde/wikit/blob/16466afe3f0794cfef195a4d8d2633f31adde997/vue-components/src/components/Lookup.vue#L120-L124)

itamargiv (changes requested):
A few minor changes and we might be good to go. Looking great otherwise!

vue-components/src/components/LookupMenu.vue
jakobw: As far as I understand it, `this` as it is written here is not actually a function parameter, but a means of letting TypeScript know that you're sure it's bound to a Vue object when it's called.

https://www.typescriptlang.org/docs/handbook/functions.html#this-parameters

itamargiv: That is well confusing... but hey, typescript ;) guess I never noticed it before as I use arrows pretty excessively...

tzhelyazkova: If the sole purpose of passing `this` is to use the `$emit` function (see my other comment below), one can do that with [vm](https://vuejs.org/v2/api/#vm-emit).
I think it's less awkward to pass `vm` to `debounce` as it's clearer what it is (also searchable in the vue docs).

vue-components/src/components/LookupMenu.vue
```diff
@@ -47,8 +50,26 @@ export default Vue.extend( {
...
+			const rootElem = this.$el as HTMLElement;
+			const menuItems = rootElem.querySelectorAll( '.wikit-LookupMenu__item' ) as NodeListOf<HTMLElement>;
```
tzhelyazkova, line 55: The "Vue way" of doing this is to use [ref](https://vuejs.org/v2/api/#ref).
An added bonus of using `ref` is future-proofing this component for the requirement to be server side rendered.

tzhelyazkova (changes requested):
Requesting changes because of the `ref` vs `querySelectorAll` comment.

vue-components/src/components/LookupMenu.vue
guergana: could you be more clear? as in code example. thanks. :)

tzhelyazkova: After revisiting the docs, we can't use vm. For some reason I thought `vm` always exists and is the same as the `this` of the vue instance, but that is wrong. So scratch that idea.

vue-components/package.json
guergana: true!

vue-components/stories/Lookup.stories.ts
guergana: heheh

vue-components/src/components/Lookup.vue
guergana: Forgot about this one. Thanks for the heads up.

vue-components/src/components/LookupMenu.vue
guergana: Great catch, we will need some refactoring in other components then 😅

tzhelyazkova (changes requested):
Requesting changes, because of the `ref` business. I think the PR is almost there :+1:

vue-components/src/components/LookupMenu.vue
```diff
@@ -47,8 +51,26 @@ export default Vue.extend( {
 				rootElem.style.maxHeight = menuHeight + 'px';
 			}
 		},
-	},
+		onScroll: debounce( function ( this: Vue ) {
+			const rootElem = this.$refs[ 'lookup-menu' ] as HTMLElement;
+			const menuItems = rootElem.querySelectorAll( '.wikit-LookupMenu__item' ) as NodeListOf<HTMLElement>;
```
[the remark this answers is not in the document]
guergana: This gives this error: Conversion of type 'Vue | Element | Vue[] | Element[]' to type 'NodeListOf<HTMLElement>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Property 'item' is missing in type 'Element[]' but required in type 'NodeListOf<HTMLElement>'.Vetur(2352)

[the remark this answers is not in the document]
guergana: It seems to work when changing the type.

itamargiv (approved):
Looking like it's going well, I will approve on my side, but agree with @tzhelyazkova that we should use refs where possible.

tzhelyazkova (approved):
Approving since the `ref` thing was addressed.

vue-components/src/components/Lookup.vue
```diff
@@ -137,6 +140,20 @@ export default Vue.extend( {
...
 		},
+		onScroll( firstIndex: number, lastIndex: number ): void {
+			/**
+			 * This event is emitted whenever the first or last index of the
+			 * visible menuItems changes. If the user scrolls but the indexes remain
+			 * unchanged the event won't fire.
+			 *
+			 */
+			if ( firstIndex !== this.scrollIndexStart || lastIndex !== this.scrollIndexEnd ) {
+				this.$emit( 'scroll', firstIndex, lastIndex );
```
itamargiv, line 151: Ooops, I approved too soon, I think this needs to be right above the `$emit` to appear in the docs.

Screenshot from 2020-11-09 17-04-46

```suggestion
			if ( firstIndex !== this.scrollIndexStart || lastIndex !== this.scrollIndexEnd ) {
			      /**
			       * This event is emitted whenever the first or last index of the
			       * visible menuItems changes. If the user scrolls but the indexes remain
			       * unchanged the event won't fire.
			       *
			       */
				this.$emit( 'scroll', firstIndex, lastIndex );
```

itamargiv (changes requested):
:( Sorry, it really is good to go, it's really just that small docs thing I approved too soon

itamargiv (approved):

itamargiv:
:confetti_ball: Thank you, great work!! :confetti_ball: 
Mr. T holding a Torah Scroll up and exclaiming: "Let's Scroll!"

PR #256, opened 2020-11-05
fix contrast of warning text

SaiSan-WMDE:
this makes inline warning messages bulletproof in terms of contrast agains any backgorund color


itamargiv (approved):
Thanks for the fix!

PR #258, opened 2020-11-05
Create animation tokens

SaiSan-WMDE:
This makes available global and alias tokens to be used as values for animations' duration and delay properties.

To be decided (please let me know your thoughts) whether we should split duration and delay tokens into separated pages.


itamargiv (approved):
LGTM, thank you!

PR #259, opened 2020-11-06
make block-level components full-width

SaiSan-WMDE:
This applies a default width of 100% to the existing block-level system components. All width props and tests applied to said components  have been removed.

⚠️ This is a breaking change! ⚠️

ADR: https://wmde.github.io/wikit/?path=/story/documentation-decisions-and-adrs-adrs-12-remove-predefined-widths--page


itamargiv (approved):
Thank you! Looking good!

PR #260, opened 2020-11-09
Expose Message component

Ladsgroup:
It's not really usable without it

### DIFF
--- a/vue-components/src/main.ts
+++ b/vue-components/src/main.ts
@@ -1,9 +1,11 @@
 import Button from './components/Button.vue';
 import TextInput from './components/TextInput.vue';
 import Lookup from './components/Lookup.vue';
+import Message from './components/Message.vue';
 
 export {
 	Button,
 	TextInput,
 	Lookup,
+	Message,
 };


micgro42:
Ugh, it is annoying that we forgot that before the release :/

Why is there a test error? o.O

Ladsgroup:
> Ugh, it is annoying that we forgot that before the release :/
> 
> Why is there a test error? o.O

It seems there is a test that builds the components and checks if they have "wikit" class but this component requires `property` property. I don't have a straightforward solution to it :sob:

micgro42:
I played around with it for quite a bit, but didn't find a way to suppress the error. Tried catching the error, which resulted in the component not being mounted and the wrapper being undefined and tried suppressing the error by [registering a custom error handler](https://vue-test-utils.vuejs.org/api/createLocalVue.html#createlocalvue), but that didn't seem to have any effect at all?

Only way forward that I see is to give the type `type` prop on the Message component a default value, probably `notice`. But this is not nice either, because it makes sense for that prop to be required.

But maybe @itamargiv has an idea?

itamargiv:
> Only way forward that I see is to give the type `type` prop on the Message component a default value, probably `notice`. But this is not nice either, because it makes sense for that prop to be required.

That was the first thought I had, but of course then it invalidates the need for a required prop...

Honestly I have some doubts about this test in general as it seems to be testing a component internal with the assumption that components are all loaded under the same circumstances... Therefore, I might need @tzhelyazkova to shed some light on why this test was added in the first place.

Otherwise, if we must keep this test, another alternative I can think of is a hash map from the component name to it's required prop values, which will be passed to the component on mount in the test.

wiese:
> shed some light on why this test was added in the first place

The test was added to ensure that all components exhibit the class "wikit" in their root node in order to handle them in #117.

Required properties were not a thing at the time, hence the - admitted - oversight. Will take a look.

wiese:
tl;dr The gist of the check should be kept but IMO moved into the static analysis world (eslint).

Due to the nature of how vue works, the real resulting mark-up - which is what we want to assert - does only come together after the component is completely processed. And to do that, vue needs its props. Chances are the props influence which version of the component (and the mark-up) we see, so this for the most part makes sense.
It's thinkable to try and play tricks by introspecting the components, e.g.
```
	it( 'have the class wikit in their root node', () => {
		Object.values( allComponents ).forEach( ( component ) => {
			const options = ( component as any ).options;
			const propsData = {} as any;
			Object.keys( options.props ).forEach( ( prop ) => {
				if ( options.props[ prop ].required ) {
					propsData[ prop ] = 'foo'; // impossible to match validators, though
				}
			} );
			const wrapper = mount( component, { propsData } );
			expect( wrapper.classes() ).toContain( 'wikit' );
		} );
	} );
```
but it's a bottomless barrel.

As a consequence I see two ways forward 
1. maintaining a list of props to feed to certain components so they can be built
2. solving the same problem through static analysis, e.g. inspired by [this](https://github.com/vuejs/eslint-plugin-vue/blob/80b8983/lib/rules/static-class-names-order.js). This will fail should someone try and attach the "wikit" class to the root node through a function instead of directly but I think we can ignore that for this use case (the "wikit" class has no business being dynamic).

N°1 could also act as a short term solution to unblock this, until we get to do N°2.
Would be interesting to hear the opinion of other DS hikers about this.

micgro42:
Moving this check to eslint seems like a good idea as we are likely to get more components with required properties

itamargiv (approved):
Fiddling with exposed wiring

PR #261, opened 2020-11-09
Remove usage of TS non-null-assertion by using

bereket-WMDE:
refs to identify the input compoentcomponent


bereket-WMDE:
no longer necessary

PR #264, opened 2020-11-11
vue-components: check components' wikit class on required props

wiese:
The test was added to ensure that all components exhibit the class
"wikit" in their root node in order to be able to have a selector for
styles (resets).

At the time we missed that components may have required props which will
have to be specified to render (mount) the component.

This allows for a quick way to specify those props per component.
However, as this an endless job, I think we may be better off swapping
this jest test for something based on static analysis (e.g. inspired by
[0]). Granted, this can not do the full complexity of a vue component
justice, but we probably don't want dynamic behavior in the wikit class
anyway.

[0]: https://github.com/vuejs/eslint-plugin-vue/blob/80b8983/lib/rules/static-class-names-order.js

### DIFF
--- a/vue-components/tests/unit/components.spec.ts
+++ b/vue-components/tests/unit/components.spec.ts
@@ -3,8 +3,18 @@ import * as allComponents from '@/main';
 
 describe( 'All components', () => {
 	it( 'have the class wikit in their root node', () => {
+		// props to feed to certain components so they can be built
+		// TODO in order not to maintain this forever, consider switching this to an eslint-based solution
+		// See https://phabricator.wikimedia.org/T267741
+		const dummyProps: Record<string, object> = {
+			Message: {
+				type: 'success',
+			},
+		};
 		Object.values( allComponents ).forEach( ( component ) => {
-			const wrapper = mount( component );
+			const wrapper = mount( component, {
+				propsData: dummyProps[ ( component as any ).options.name ] || {},
+			} );
 			expect( wrapper.classes() ).toContain( 'wikit' );
 		} );
 	} );


Ladsgroup:
That is slightly weird. The test fail due to visual changes but this doesn't touch anything. Here says the warning message got green(?) https://www.chromatic.com/snapshot?appId=5efdb3b5f65950002286285d&id=5fabb9e57bed9300214248da

Maybe needs a rebase?

vue-components/tests/unit/components.spec.ts
```diff
@@ -3,8 +3,17 @@ import * as allComponents from '@/main';
 
 describe( 'All components', () => {
 	it( 'have the class wikit in their root node', () => {
+		// props to feed to certain components so they can be built
+		// TODO in order not to maintain this forever, consider switching this to an eslint-based solution
```
itamargiv, line 7: I'm not a big fan of inline TODO as they tend to get forgotten and fall out of sync with their surrounding code. Can you please open up a ticket in the DS Phab board to consider the eslint method, that way it will also be prioritized.

itamargiv (changes requested):
Looks great! thank you! one small note about the TODO comment. and we should definitely lookinto wht this change affects the visual tests.

vue-components/tests/unit/components.spec.ts
micgro42: Created the task: https://phabricator.wikimedia.org/T267741
And explicitly mentioned it in the comment, so readers of the comment can directly evaluate whether it is still relevant or can be removed.

micgro42:
The visual change is just some artifact. For some reason the baseline is still the old yellow warning message, whereas on master it is actually already black-ish.

itamargiv:
> The visual change is just some artifact

I think that in order to unblock this impasse and reduce the amount of changes in the base branch, we should rebase this PR on master, and merge it as soon as tests pass, and then rebase the PR from base branch on master (which needs to be done anyway due to the CI changes).

itamargiv (approved):

PR #265, opened 2020-11-11
Rename Button prop 'type' to 'flavor'

tzhelyazkova:
The `Button` component is simply a `<button>` HTML element, which means the native attribute `type` cannot be set on this
component due to the component already having a prop named 'type'.
Renaming the prop to 'flavor' allows the consuming application to set the `type` attribute of Button to `submit`, `reset` etc. while still getting the desired visual type of button e.g. 'primaryProgressive'.

### **This is a breaking change!** :bomb: :boom:


micgro42 (approved):

itamargiv:
I think this is an important change. I kinda wish we didn't start from version 1.0.0 though, since now we have to bump to 2.0.0 for something that looks relatively trivial. 

Also, I think maybe it will be easier for us to track the next version if we had a `CHANGELOG.md` in master (this could also help with creating the release notes, when they need to be created) What do you think, should we start one here?

tzhelyazkova:
> Also, I think maybe it will be easier for us to track the next version if we had a `CHANGELOG.md` in master (this could also help with creating the release notes, when they need to be created) What do you think, should we start one here?

I consider GitHub's Release Notes feature to be the same as a `CHANGELOG.md` doc in the repo. So IMHO it should be one or the other.
I really don't have a strong opinion how to document the changes between versions. If you think it'd be more useful to have a changelog because it's easier to discover, then ok let's do it :rocket:

itamargiv:
> If you think it'd be more useful to have a changelog because it's easier to discover, then ok let's do it :rocket: 

I just think it's useful in a sense that when we make a release (which includes many changes across components), it's easier to parse
A) Which is the next version
B) What to include in the release notes, as we don't need to compare diffs or fish them out of the history or phab board.

I don't see them as mutually exclusive, but as one being a draft of the other.

vue-components/src/components/Button.vue
```diff
@@ -23,9 +23,9 @@ export default Vue.extend( {
 	name: 'Button',
 	props: {
 		/**
-		 * The type of the button
+		 * The flavor of the button
```
SaiSan-WMDE, line 26: Maybe using the word "typology" or "style"  instead of flavor in the description could make it easier to understand the prop?

tzhelyazkova:
> I just think it's useful in a sense that when we make a release (which includes many changes across components), it's easier to parse
> A) Which is the next version
> B) What to include in the release notes, as we don't need to compare diffs or fish them out of the history or phab board.
Ok, makes sense.
My concern is that not everyone will update the doc, as we are many people contributing to one repo. 
OTOH, the person creating a new release has the responsibility to include everything that has been changed between tags.

vue-components/src/components/Button.vue
tzhelyazkova: `style` is also an attribute, so I wouldn't name it like that.
I don't know about `typology`. Is it more to the point than `flavor`? What do others think?

itamargiv:
Yeah, I have the same concern, but we can for sure make it part of the release process to also bring the changelog file to a good state (noting it down to update the docs)

micgro42:
Wait, wouldn't the idea of a Changelog be that it is updated with basically every pull request? I.e. make it part of the process. Maybe there could even be linting to ensure that this file was touched?

If the Changelog is to be created after the fact as well, then I'm not sure that I understand the benefit from it over just the GitHub release notes.

itamargiv:
> Wait, wouldn't the idea of a Changelog be that it is updated with basically every pull request?

Yes, that's the ideal scenario. but not every PR requires to add a change to a changelog (think of updating package dependencies for instance), and sometimes things fall between the cracks, my suggestion is just to add a step to the docs to make sure to proof the changelog upon release.

micgro42:
I like that Changelog file, thank you :pray:.
Let's see how it goes and whether we run into those frequent changelog merge conflicts that [other people seem to have](https://www.google.com/search?q=changelog+merge+conflicts)

vue-components/src/components/Button.vue
itamargiv: Personally I prefer `flavor`, as it's more snazzy than `typology`. Could live with `variant` as well. In [reactstrap](https://reactstrap.github.io/components/buttons/) they call it `color` but I don't like that...

SaiSan-WMDE: Hey! 👋  I wasn't questioning the use of `flavor`. I just wanted to make sure that the description of the prop is clear enough for those not familiar with the jargon. 
As a side note, the `type` prop is also used with the Message component. I believe we can keep `type` there, since `flavor` might not work well. Let me know if you agree.

tzhelyazkova: > Hey! wave I wasn't questioning the use of `flavor`. I just wanted to make sure that the description of the prop is clear enough for those not familiar with the jargon.

Ok cool, can you please make a suggestion from the UI here on github to change it to `typology` and I'll accept it.

> As a side note, the `type` prop is also used with the Message component. I believe we can keep `type` there, since `flavor` might not work well in that case. Let me know if you agree.

It's fine to leave the `type` prop in `Message` because that component's element is a `<div>`, which cannot have a native attribute `type`

SaiSan-WMDE, line 26:
```suggestion
		 * The typology of the button
```

jakobw:
This prop was called `flavor` originally and was changed to `type` in #128 ccae6ad1a7f5cd7d5db58d78bafad16a21b41823 after realizing ([relevant comment](https://github.com/wmde/wikit/pull/128#issuecomment-672067439)) that `type` is the more conventional term.

We could introduce a new prop `native-type`/`html-type` to set the button's native `type` when using the component. This may be slightly confusing to developer the using the component, but I think it's something that can be mitigated with documentation and keeps the problem contained in the dev realm. I think it's more important to align the language between dev and UX.

tzhelyazkova:
Off-github discussion happened, and we concluded that instead of renaming this prop from `type` to `flavor`, we should provide another prop e.g. `nativeType` to pass the value of the native attribute type.
Ticket tracking the proposed change: https://phabricator.wikimedia.org/T267938

PR #266, opened 2020-11-11
vue-components: check components' wikit class via eslint

wiese:
Use static analysis of the component code to make sure that all
components carry a "wikit" class on their root node so that our styles
and style resets can work properly (see #117).

The pre-existing test-driven approach had the disadvantage that one
needed to maintain the list of required props of all components for them
to be mountable.

Bug: https://phabricator.wikimedia.org/T267741


itamargiv (commented):
Nice! :heart_eyes: I really love it,  and it's already being applied on this branch with this cool new GH feature:

Screenshot from 2020-11-13 08-51-19

The only comment I would have, is that I think it would make more sense (and perhaps be more consistent) If we put this plugin at the project's root instead of nested within the components library.

wiese:
> I think it would make more sense (and perhaps be more consistent) If we put this plugin at the project's root instead of nested within the components library

Definitely. I think I'd go one step further - let's make the desired class configurable and move the plugin into a dedicated repo (given that I don't find it an exotic feature at all I actually expected to find it somewhere from someone else but could not). Just scribbled it here for ease of development.

itamargiv (approved):
So happy to see this error :heart: 

Screenshot from 2020-11-25 14-09-08

Fantastic work @wiese ! Thank you!

PR #268, opened 2020-11-12
firefox_doesnt_show_active_styles_while_LookupMenu-item_is_clicked

bereket-WMDE:
@itamargiv  thanks for the help yesterday. we did miss one thing, though.
the focus on the input is lost after selecting an item.
@micgro42 had another approach and it's working with FF and chrome (haven't tested with safari and IE)

### DIFF
--- a/vue-components/src/components/LookupMenu.vue
+++ b/vue-components/src/components/LookupMenu.vue
@@ -10,10 +10,12 @@
 			:key="index"
 			v-for="(menuItem, index) in menuItems"
 			:class="{
-				'wikit-LookupMenu__item--selected': index === selectedItemIndex
+				'wikit-LookupMenu__item--selected': index === selectedItemIndex,
+				'wikit-LookupMenu__item--active': index === activeItemIndex,
 			}"
 			@click="$emit( 'select', menuItem )"
-			@mousedown.prevent
+			@mousedown.prevent="activeItemIndex = index"
+			@mouseup="activeItemIndex = -1"
 			ref="menu-items"
 		>
 			<div class="wikit-LookupMenu__item__label">
@@ -41,6 +43,7 @@ export default Vue.extend( {
 	data() {
 		return {
 			maxHeight: null as number|null,
+			activeItemIndex: -1,
 		};
 	},
 	props: {
@@ -126,6 +129,8 @@ $base: '.wikit-LookupMenu';
 			cursor: $wikit-LookupMenu-item-hover-cursor;
 		}
 
+		&--active,
+		&--active:hover,
 		&:active {
 			background-color: $wikit-LookupMenu-item-active-background-color;
 


vue-components/src/components/LookupMenu.vue
```diff
@@ -9,10 +9,12 @@
...
 			v-for="(menuItem, index) in menuItems"
 			:class="{
-				'wikit-LookupMenu__item--selected': index === selectedItemIndex
+				'wikit-LookupMenu__item--selected': index === selectedItemIndex,
+				'wikit-LookupMenu__item--active': index === activeItemIndex,
 			}"
 			@click="$emit( 'select', menuItem )"
-			@mousedown.prevent
+			@mousedown.prevent="activeItemIndex = index"
+			@mouseup="activeItemIndex = -1"
```
itamargiv, line 17: It looks good. However, I think we should avoid adding more javascript to a problem that is caused by javascript. I think the addition of this `preventDafault` on mousedown might have been an oversight in the past,and if we need to prevent default browser behaviour, it's probably in that `onBlur` method we created yesterday to overcome the problem (this would also move this focus responsibility back to the parent, where it belongs)

itamargiv (changes requested):
Generally speaking this is a good approach, @bereket-WMDE and @micgro42 but I still think yesterdays approach was better, as it moved the responsibility of handling the blur event back to the parent, where it is more explicit. See comment below for more info.

vue-components/src/components/LookupMenu.vue
micgro42: I very much agree with the general sentiment of not doing with JavaScript what can be done with CSS.

However, I don't see how _that_ functionality belongs into the parent. Quite the opposite, it breaks the logic-encapsulation of the child, as every parent component that uses LookupMenu is now forced to add a handler for `onBlur` because otherwise the onSelect will never fire for mouse clicks as the LookupMenu will already have disappeared. That seems rather like a good example of leaking internal logic.

Otherwise, it would be the responsibility of that `onBlur`-handler to figure out whether the focus-loss has happened due to an mouse-down above an item in the LookupMenu or not in order determine whether the LookupMenu should still be shown. That seems a strange thing for a parent component to be worrying about.

micgro42: While still having different views on some other aspects, @itamargiv and I agreed that the LookupMenu will unavoidably require a number of extra logic in its parent component anyway, so adding to that list isn't too much of an antipattern anymore.
This extra logic that is required for the use of LookupMenu will need good documentation. It will probably be best to add that extra documentation in the course of extracting the Dropdown out of the Lookup.

itamargiv (approved):
After looking at this issue from several different angles it was decided that for now, it seems the best way to prevent focus issues. Thank @micgro42 and @bereket-WMDE for your patience.

micgro42:
I'll rebase and merge it quickly, so I can base my refactoring for extracting the Dropdown on top of it

itamargiv (approved):

PR #271, opened 2020-11-13
Set LookupMenu's maxHeight via Vue's virtual DOM

tzhelyazkova:
Any dynamic changes to the style of a component should happen through Vue's Virtual DOM rather than directly editing the browser's DOM. Doing the opposite will result in unexpected behavior.

Info about Vue's virtual DOM can be found in [this section of the Vue docs](https://vuejs.org/v2/guide/render-function.html).


itamargiv (approved):
Looks good, thank you!

PR #277, opened 2020-11-16
Refactor keyboard handling into LookupMenu

micgro42:
Todo:
- [x] Esc handling
- [x] typing for this.$refs.menu
- [x] marking the already selected item as selected on focus
- [x] better commit message
- [x] tests

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -10,11 +10,8 @@
 			:feedback-type="feedbackType"
 			:placeholder="placeholder"
 			:disabled="disabled"
-			@keydown.up.native.prevent="onArrowUp"
-			@keydown.down.native.prevent="onArrowDown"
-			@keyup.enter.native="onEnter"
-			@keydown.tab.native="onTab"
-			@keyup.esc.native="onEsc"
+			@keydown.native="onKeyDown"
+			@keyup.native="onKeyUp"
 			autocomplete="off"
 		/>
 		<LookupMenu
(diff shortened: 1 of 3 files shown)


vue-components/src/components/Lookup.vue
```diff
@@ -10,20 +14,16 @@
...
-			@keyup.enter.native="onEnter"
-			@keydown.tab.native="onTab"
-			@keyup.esc.native="onEsc"
 			autocomplete="off"
 		/>
 		<LookupMenu
 			class="wikit-Lookup__menu"
 			:menu-items="menuItems"
-			v-if="showMenu"
+			v-show="showMenu"
```
bereket-WMDE, line 22: nice catch :+1:

vue-components/src/components/Lookup.vue
```diff
@@ -10,20 +14,16 @@
...
 			@select="onSelect"
 			@scroll="onScroll"
-			:selected-item-index="selectedItemIndex"
```
bereket-WMDE, line 26: you removed it here but it's used on the LookupMenu.vue file on line 13.
does this still work ?

micgro42: I think so, the LookupMenu has it as a local state now: 
image

bereket-WMDE: ahh.. I missed that one.
thanks

vue-components/src/components/Lookup.vue
```diff
@@ -10,20 +14,16 @@
...
+			@esc="onEsc"
```
itamargiv, line 25: Great name :+1:

vue-components/src/components/Lookup.vue
```diff
@@ -52,15 +52,14 @@ import { MenuItem } from '@/components/MenuItem';
  *
  * Uses the following components internally: Input, ValidationMessage and LookupMenu
  */
-export default Vue.extend( {
+export default ( Vue as VueConstructor<Vue & { $refs: { menu: InstanceType<typeof LookupMenu> } }> ).extend( {
```
itamargiv, line 55: I think we might need to find a more readable solution to this in the near future.

itamargiv (approved):
Well Done! works fantastically!



Leaving open until the end of the day in case anyone else wants to have a look, I also left a non blocking comment.

vue-components/src/components/Lookup.vue
micgro42: I think there were ideas for that in the links I read about it:

* vuejs/vue-class-component#94
* bluecanvas/vue-typed-refs

They are also in the commit message, added here for future reference

micgro42:
> Leaving open until the end of the day in case anyone else wants to have a look

I'm sorry for merging this, I was so much on auto-pilot that I over-read this comment :facepalm:

PR #275, opened 2020-11-16
Use latest version of wikimedia-ui-base

SaiSan-WMDE:
This way, our global tokens are based on the variables contained by the latest version of wikimedia-ui-base whenever possible.

Issue: https://phabricator.wikimedia.org/T263722

PR #278, opened 2020-11-16
Fix package-lock.json

itamargiv:
When merging #275 we did not notice that the updated `package-lock.json` was not included in the PR. This causes the latest commit on master to fail during documentation deployment.


micgro42:
Why did this not fail in #275 CI?

micgro42 (approved):
This shifting around of dependencies in the vue-components/package-lock.json is weird, but probably harmless? I'm not yet sure what causes it

itamargiv:
> Why did this not fail in #275 CI?

Because the dependencies cache was not invalidated (it invalidates with changes to `package-lock.json`)

PR #279, opened 2020-11-17
Enable Controls in composed Storybooks

micgro42:
Apparently, any plugin that one wants to use in a composed storybook has to be present in both storybooks, not only the one where it makes sense.

This was missed in c8aea09 / #225


vue-components/stories/Message.stories.ts
```diff
@@ -7,22 +7,7 @@ export default {
...
-				<p><Message type="notice">A neutral message which inform about general higher priority notices to the user. Stay neutral!</Message></p>
-				<p><Message type="success">A success message which means that the user did everything right and should know about it. Yay!</Message></p>
-			</div>
-		`,
-	};
-}
-
-
-export function controllableMessage( args: { fullwidth: boolean, content: string, type: string } ): Component {
+export function message( args: { fullwidth: boolean; content: string; type: string } ): Component {
```
SaiSan-WMDE, line 10:
```suggestion
export function message( args: { content: string; type: string } ): Component {
```
We should remove the `fullwidth` argument. Wondering if this is the only place where it should be removed from?

docs/package.json
```diff
@@ -36,6 +36,7 @@
     "@babel/core": "^7.12.3",
     "@babel/preset-react": "^7.12.5",
     "@storybook/addon-a11y": "^6.0.28",
+    "@storybook/addon-controls": "^6.0.28",
     "@storybook/addon-docs": "^6.0.28",
     "@storybook/addon-knobs": "^6.0.28",
```
itamargiv, line 41: I think it's best to remove the knobs plugin as part of this change as theses two features are meant to replace each other. Also it seems like you took on [T261652](https://phabricator.wikimedia.org/T261652) (which is a good thing, less work for me), Can you please assign it to yourself and move it to doing in accordance with our DS workflows?

vue-components/stories/Message.stories.ts
micgro42: Done, I think

docs/package.json
micgro42: Can do, though this PR was merely intended as fixing something that I thought already worked :sweat_smile:.

I'll add a commit that drops Knobs and assign this PR.

vue-components/stories/Message.stories.ts
```diff
@@ -7,22 +7,7 @@ export default {
...
-				<p><Message type="notice">A neutral message which inform about general higher priority notices to the user. Stay neutral!</Message></p>
-				<p><Message type="success">A success message which means that the user did everything right and should know about it. Yay!</Message></p>
-			</div>
-		`,
-	};
-}
-
-
-export function controllableMessage( args: { fullwidth: boolean, content: string, type: string } ): Component {
+export function message( args: { content: string; type: string } ): Component {
```
itamargiv, line 10: I think we should follow the naming conventions @SaiSan-WMDE present to us in the [storybook structure document](https://docs.google.com/document/d/1loENZ13z5s_Nes4exm5Z9d5S3PnR7F_0brv5azQ5XB4/edit). In that case the story should be called `Basic`.

vue-components/stories/Message.stories.ts
```diff
@@ -53,6 +37,21 @@ controllableMessage.argTypes = {
 	content: {
 		control: {
 			type: 'text',
-		}
+		},
 	},
 };
+
+export function allMessageTypes(): Component {
```
itamargiv, line 44: Same comment as above, only here according to the structure guideline, the story should be called `All`.

itamargiv (changes requested):
Looking Great! :sunglasses: Thank you for taking this on! I have one small request for a change and we should be good to go.

vue-components/stories/Message.stories.ts
micgro42: Done.

vue-components/stories/Message.stories.ts
micgro42: I interpreted the outcome of this [discussion](https://mattermost.wikimedia.de/swe/pl/i5ik69ijzpdt9jf3cr6z98ocyh) as it being agreed that it makes sense for the story titles in the menu to be more specific if possible. But maybe I misread it?

itamargiv: Oh, I didn't realize it was concluded, If @SaiSan-WMDE agrees, then I approve as well :)

SaiSan-WMDE: After giving it some more thought: being specific (but yet consistent) would mean using either "All" (contains: flavors, disabled, validation states...in case of buttons, for example) or "All types" (component typologies) as titles for these sections, and I'm not sure how much helpful adding the word "type" in certain cases would really be? So I'd say that we should start using only the word "All" in every section. Do you agree?

micgro42: If the actual desire is to always have all permutations in a single page, then I will implement that. Will update the PR.

micgro42: Done

itamargiv (approved):
Looks Great, thank you again @micgro42 ! (Awaiting @SaiSan-WMDE's storybook approval)

Animation of a hand trying to choose a button to click

vue-components/stories/Message.stories.ts
SaiSan-WMDE: thanks, Michael! 🙏

SaiSan-WMDE (approved):
looking good! Danke schön, Michael 😄

PR #282, opened 2020-11-18
Make bold titles in LookupMenu optional

micgro42:
This is done in preparation for remaking the LookupMenu in to a more general OptionsMenu that is then to also be used for the Dropdown. And the Dropdown has options who's label is not bold.

Two options and controls are added to the story to showcase this.

Todo:
- [x] get clarity on that we actually want to switch the boldness of options via a prop for _all_ options in a menu instead of it being decided dynamically depending on whether we are description available
- [x] we probably want to adjust the tokens in order to define the boldness of the normal-weight font (done in [46533a3](https://github.com/wmde/wikit/pull/282/commits/46533a3aefd9f983ccddb0b9f356c62d844591da))

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -19,6 +19,7 @@
 		<LookupMenu
 			class="wikit-Lookup__menu"
 			:menu-items="menuItems"
+			:bold-labels="true"
 			v-show="showMenu"
 			@select="onSelect"
 			@scroll="onScroll"
--- a/vue-components/src/components/LookupMenu.vue
+++ b/vue-components/src/components/LookupMenu.vue
@@ -18,7 +18,12 @@
 			@mouseup="activeItemIndex = -1"
 			ref="menu-items"
 		>
-			<div class="wikit-LookupMenu__item__label">
+			<div
+				:class="{
+					'wikit-LookupMenu__item__label': true,
+					'wikit-LookupMenu__item__label--bold': boldLabels,
+				}"
+			>
 				{{ menuItem.label }}
 			</div>
 			<div class="wikit-LookupMenu__item__description">
@@ -51,6 +56,13 @@ export default Vue.extend( {
 		menuItems: {
 			type: Array,
 			default: (): [] => [],
+		},
+		/**
+		 * Whether the label of the options should be bold
+		 */
+		boldLabels: {
+			type: Boolean,
+			default: false,
 		},
 	},
 	methods: {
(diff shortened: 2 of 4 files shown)


micgro42:
@SaiSan-WMDE and @itamargiv Your input/thoughts on the Todos and the general approach would be welcome :)

vue-components/src/components/LookupMenu.vue
```diff
@@ -52,6 +57,10 @@ export default Vue.extend( {
 			type: Array,
 			default: (): [] => [],
 		},
+		boldTitles: { // FIME: are we sure we need a property here and its not just a function of a description present?
```
itamargiv, line 60: I agree and I think this should be implicit in this case, so that the menu appears according to what kind of Menu Items it receives.

SaiSan-WMDE: Making the style of the item labels (they're technically not titles) dynamic, depending on a function of whether a description is present, could then generate menus where some labels are bolded and some aren't, which could communicate a false hierarchy and be confusing for users (and even impact findability?). 

This is I would rather go for what this PR does: giving implementers the freedom to decide whether they want to make the labels more distinguishable in case they know a description will be present.

micgro42: Thank you for clarifying it :)

micgro42:
@SaiSan-WMDE Thank you for the adjusted tokens! :blush:

micgro42:
From my side this should now be ready to be officially reviewed and merged.

SaiSan-WMDE (dismissed):

vue-components/src/components/Lookup.vue
```diff
@@ -19,6 +19,7 @@
 		<LookupMenu
 			class="wikit-Lookup__menu"
 			:menu-items="menuItems"
+			:bold-titles="true"
```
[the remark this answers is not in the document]
micgro42: I'll rename it, no problem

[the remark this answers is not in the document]
micgro42: Fixed, I think

micgro42:
As far as I can tell, the new visual changes not caused by this change. Maybe sauce labs updated their Chrome-driver or something?

itamargiv:
> Maybe sauce labs updated their Chrome-driver or something?

Good instinct, though the visual diffs are computed by chromatic. I think the visual changed (changes in height if you look at the diff) are caused by a combination of changing the tokens and what seems like changes in uuid? (wonder why the storybook is not stopping us for each build then)

itamargiv (approved):
LGTM, thank you! 

Baristas saying "she's bold" to each other

PR #283, opened 2020-11-19
Update package-lock.json

tzhelyazkova:
A fresh `npm install` inside the `vue-components` folder on `master` generates changes in `package-lock.json`.

### DIFF
--- a/vue-components/package-lock.json
+++ b/vue-components/package-lock.json
@@ -8242,6 +8242,11 @@
 			"integrity": "sha512-x7H7dAf/x3Q+OzckzrQqQs6TrSx03v07qkvB/EDzmvOnreYy6q/fJKHQoFei7kXzQmKuwzuuD8OdKXbfLs8sew==",
 			"dev": true
 		},
+		"@wmde/wikit-tokens": {
+			"version": "1.2.0-alpha.0",
+			"resolved": "https://registry.npmjs.org/@wmde/wikit-tokens/-/wikit-tokens-1.2.0-alpha.0.tgz",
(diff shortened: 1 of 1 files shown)


micgro42 (approved):
Out of curiosity: Did you use the Docker container for this or you local node/npm? If the local one, then what is your node version?

We occasionally have some issues with some changes to package-lock.json that looks like some dependencies of dependencies being moved around but not actually changed.

tzhelyazkova:
> Out of curiosity: Did you use the Docker container for this or you local node/npm? If the local one, then what is your node version?

I ran it locally. 
I did try it with docker as well, but since that has to be executed from the root folder it only installs packages based on the root package.json, right? It audits 758 packages.
If I do the same in `/vue-components` it audits 3130.

> We occasionally have some issues with some changes to package-lock.json that looks like some dependencies of dependencies being moved around but not actually changed.

Hmm, interesting. I looked at the package-lock before opening this and seems like it's legit changes, like packages being removed. :woman_shrugging:

tzhelyazkova:
> Out of curiosity: Did you use the Docker container for this or you local node/npm? If the local one, then what is your node version?

I ran it locally. 
I did try it with docker as well, but since that has to be executed from the root folder it only installs packages based on the root package.json, right? It audits 758 packages.
If I do the same in `/vue-components` it audits 3130.

> We occasionally have some issues with some changes to package-lock.json that looks like some dependencies of dependencies being moved around but not actually changed.

Hmm, interesting. I looked at the package-lock before opening this and seems like it's legit changes, like packages being removed. :woman_shrugging:

itamargiv:
What strikes me as particularly odd is that the tokens dependency is excluded.. I would think that running `lerna bootstrap` from root would ensure these are installed properly...

PR #284, opened 2020-11-19
Rename LookupMenu to OptionsMenu

micgro42:
Builds upon: #282 
Is followed by: #288 

That component is now generic enough to be used both by the Lookup and the Dropdown and so it should have a name that reflects this.


guergana (approved):
Looks good. It seems to be stuck on Chromatic but once it gets approved it's good to go.

guergana:
Looks like it needs a rebase.

guergana (approved):
LGTM!

itamargiv (approved):

PR #285, opened 2020-11-19
vue-components: Update npm dependencies

tzhelyazkova:
`added 564 packages from 447 contributors, removed 462 packages, updated 255 packages, moved 54 packages and audited 3238 packages in 86.312s`

Note: dependabot probably doesn't look for nested `package.json` files.


micgro42:
> TEST FAILURE: 28 errors during execution; 24 assertions failed, 21 passed (6m 42s)

That's a lot of Browser test errors :thinking: 

I'll start the tests again, but if that is flakyness, then it reached a new quality.

tzhelyazkova:
> > TEST FAILURE: 28 errors during execution; 24 assertions failed, 21 passed (6m 42s)
> 
> That's a lot of Browser test errors thinking

Yes, something's off. Will have a look tomorrow.

itamargiv:
> Yes, something's off. Will have a look tomorrow.

There is a ticket for it, I was going to look into it on Monday... If you pick it up, please assign it to yourself and move it to the doing column https://phabricator.wikimedia.org/T268107

itamargiv:
It looks like the test failures here are different, saucelabs can't find `__Lookup-menu` on any of the browsers.

micgro42:
Could you update the storybook dependencies to 6.1.6? Otherwise it will keep bugging us^^

tzhelyazkova:
> Could you update the storybook dependencies to 6.1.6? Otherwise it will keep bugging us^^

Unfortunately the new version of Storybook does not seem to work on IE11.
I'm seeing a Syntax error in `vendors~main.0b8843e11b5751b64f8c.bundle.js` which is in `/storybook-static`.
The error is related to `/node_modules/prettier/parser-html.js`

micgro42:
Thank you for checking in on the IE errors :pray:

There seems to a couple of problems with IE11 and the new storybook, e.g. storybookjs/storybook#12399. Apparently, they were fine releasing the new storybook version without supporting it :shrug: 

How do we move forward from here? I'm not sure we have already officially abandoned IE11, so we could move forward with this PR without the storybook updates?

The alternative would be to drop the IE11 tests entirely from our CI. I wouldn't be terribly sad with this solution either.

itamargiv:
> How do we move forward from here?

I think that for now we should refrain from updating storybook, as we made a conscious decision to support IE11, and I don't think that the fact that storybook doesn't support it should cause us to follow suite.

tzhelyazkova:
> There seems to a couple of problems with IE11 and the new storybook, e.g. [storybookjs/storybook#12399](https://github.com/storybookjs/storybook/issues/12399). Apparently, they were fine releasing the new storybook version without supporting it shrug

Yes, but they claimed the bugs are fixed and it works on IE11 now, so I'm inclined to think it's our setup that's at fault :thinking: 

> How do we move forward from here? I'm not sure we have already officially abandoned IE11, so we could move forward with this PR without the storybook updates?

I downgraded Storybook back to 6.0.28 which is the last version that worked in wikit, so we can move forward with the original idea behind this patch - update dependencies. I'll open a ticket for the storybook update because it seems to be an adventure on its own.

> The alternative would be to drop the IE11 tests entirely from our CI. I wouldn't be terribly sad with this solution either.

Supporting IE11 was a decision taken by the whole team back at the start of the DS, so I think it's more of a product question than a dev one.
The whole "using nightwatch and saucelabs" was because of IE11. So I'd be suprised if a few month later we are ready to drop support.

itamargiv (approved):
\o/ Thank you!

Person sparkling glitter

PR #289, opened 2020-11-20
Ensure tokens are present before building storybook

itamargiv:
This PR fixes a problem in #282 Where the tokens are not up to date with the current state of the code. This is done by ensuring that each time the storybook is built, the tokens are built as well, as a prerequisite.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "postinstall": "lerna bootstrap",
     "test": "lerna run test",
     "test:e2e": "lerna run e2e:saucelabs --scope @wmde/wikit-vue-components",
-    "build:storybook": "npm-run-all build:storybook:*",
+    "build:storybook": "lerna run build --scope @wmde/wikit-tokens && npm-run-all build:storybook:*",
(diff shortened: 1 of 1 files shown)


micgro42 (approved):
Shouldn't that happen anyway in CI?

I'm not exactly sure I get all the implications of this, but I'm pretty confident that it wont break anything, maybe only take a bit longer.

itamargiv:
> Shouldn't that happen anyway in CI?

No, the CI is split into 4 separate jobs, each gets it's own runner. The token are actually built in job 1: "Setup" via `lerna bootstrap` and was used in job 3: "Storybook build and deployment", via the cached build artifacts. (since I didn't want to have to bootstrap the whole thing in each job).

The problem in our case was that the cached tokens were out of sync to what the storybooks were expecting (since I broke the caching in one of my former PRs)

Either Or, this change is necessary, since the tokens are a dependency of the storybook even outside of CI, and therefore, for the scripts to function properly and "fool proof"ly in personal dev environments, it needs to ensure that the tokens are available to it prior to building the docs / vue components storybooks.

PR #290, opened 2020-11-20
Replace artifact caching with storybook caching

itamargiv:
This PR implements caching for storybooks instead of artifact caching, as artifact caching was only needed for building the story book, which was resolved in #289. This also has the advantage of shaving off 2.5 minutes off of storybook deployment.

The changes in this PR were validated and tested by pushing noop changes to files in each of the following folders to ensure cach busting and retrieval:
`vue-components/stories`
`vue-components/src`
`tokens/properties`

Issue: [T267828](https://phabricator.wikimedia.org/T267828)


.github/workflows/test-and-demo.yaml
```diff
@@ -94,13 +87,16 @@ jobs:
...
+    - name: Cache storybook
       uses: actions/cache@v2
+      id: storybook-cache
       with:
         path: |
-          */dist
-        key: ${{ runner.os }}-${{ hashFiles('*/dist') }}
+          docs/dist
+          vue-components/storybook-static
+        key: ${{ runner.os }}-${{ hashFiles('tokens/properties/**/*', 'vue-components/stories/**/*', 'vue-components/src/**/*') }}
```
micgro42, line 97: could we also add a hash for the package-lock files here? So that the storybooks are also rebuild if some dependencies change. E.g. we are updating storybook 6.1.1 (which we should do)

Also, shouldn't this include the stories in `docs`?

And maybe the `.storybook` directories

After all these are included, I'm not that sure that there will be a lot PRs that don't touch the cache :thinking:

itamargiv: Sure, will do on Monday

itamargiv:
@micgro42 I have update the PR in accordance to your review. Thank you! I did not include the docs storybook to the cache key, as it is not actually deployed for testing (it is only deployed once a merge to master happens, and this happens in a separate workflow, that does not include caching).

micgro42 (approved):
> I did not include the docs storybook to the cache key, as it is not actually deployed for testing (it is only deployed once a merge to master happens, and this happens in a separate workflow, that does not include caching).

This makes sense to me, thank you!

addshore:
What's the status of this one?
Still needed? Shall we rebase it and try and get it in? or is it a bit out of date now? :)

itamargiv:
yes, it's quite out of date at the moment, especially since we now started deploying to netlify as well, I think I'll close this and move the task back to todo.

PR #293, opened 2020-11-23
[PoC] Remove init from e2e test cases

guergana:
+ testing if we really need to init the tests before each test
removing them can potentially solve the problem of making too many
calls to chromatic


guergana:
closing because this didn't solve the concurrent access problem in Chromatic in the right way.

PR #294, opened 2020-11-23
Add e2e delay

itamargiv:
After discovering that the tests runner was causing chromatic to respond with 503 due to heavy traffic, we decided to try and add a pause between each test, to mitigate the amount of requests.


micgro42 (approved):
This could be moved to a `before` method, but that can also happen in a follow-up

guergana (approved):
It looks good. But also surprised that half a second is enough.

PR #296, opened 2020-11-23
Correct value of progressive-active font color

SaiSan-WMDE:
The value of font-color-progressive-active should be Accent 30 instead of Accent 50.

This is not a dejà vu. I failed to actually replace the color value in the original PR (#272). I only had one job :facepalm:


itamargiv:
LOL, I can't believe I didn't catch it... :facepalm:

micgro42 (approved):

PR #297, opened 2020-11-23
Convert to bdd tests

itamargiv:
Follow up PR to #294 in preparation for PR #293


vue-components/tests/e2e/specs/Lookup.test.js
```diff
@@ -1,50 +1,44 @@
-module.exports = {
-	'Lookup Menu shows on user input': ( client ) => {
+describe( 'Lookup Menu', function () {
```
micgro42, line 1: Why are we not using arrow functions, i.e., `() => { ... }`, here?

itamargiv: Since the docs specify the ability to change general testing settings with `this` I decided to just follow their example and use functions with a `this`context. See https://nightwatchjs.org/guide/using-nightwatch/using-bdd-describe.html

micgro42 (approved):

micgro42 (dismissed):

micgro42 (approved):

PR #298, opened 2020-11-24
Add Dropdown component

micgro42:
The `<label>` does not seem to natively focus non-form elements even if they have a `tabindex >= 0` (at least on Firefox), so this is done explicitly via a click event.

The SVG of the small arrow is added inline for now. This allows us to style it with CSS, which wouldn't be possible if we were to use it via background-image. It can be extract if and when we need to use it somewhere else as well. It was not added to the Icon component as that component explicitly sets a color instead of allowing the usage of the current text color.

The scroll events are ignored as a Dropdown always has all its possible options available.

The validation message will be added in another commit.

Bug: [T266545](https://phabricator.wikimedia.org/T266545)

### DIFF
--- a/vue-components/src/components/Dropdown.vue
+++ b/vue-components/src/components/Dropdown.vue
@@ -1,0 +1,228 @@
+<template>
+	<div
+		:class="[ 'wikit', 'wikit-Dropdown' ]"
+		@keydown="triggerKeyDown"
+	>
+		<label
+			class="wikit-Dropdown__label"
+			@click="$refs.select.focus()"
+		>{{ label }}</label>
+		<div
+			class="wikit-Dropdown__select"
+			:tabindex="!disabled && '0'"
+			:disabled="disabled"
+			@click="onClick"
+			@blur="showMenu = false"
+			ref="select"
+		>
+			<span
+				class="wikit-Dropdown__selectedOption"
+				v-if="value"
+			>{{ value.label }}</span>
+			<span
+				class="wikit-Dropdown__placeholder"
+				v-else
+			>{{ placeholder }}</span>
+			<svg
+				class="wikit-Dropdown__arrow"
+				fill="none"
+				viewBox="0 0 12 12"
+				xmlns="http://www.w3.org/2000/svg"
+			><path fill="currentColor" d="M6 10L12 3.94231L11 3L6 7.98077L1 3L5.2958e-07 3.94231L6 10Z" /></svg>
+		</div>
+		<OptionsMenu
+			class="wikit-Dropdown__menu"
+			:menu-items="menuItems"
+			:selected-item-index="selectedItemIndex"
+			v-show="showMenu"
+			@select="onSelect"
+			@esc="onEsc"
+			ref="menu"
+		/>
+	</div>
+</template>
+
+<script lang="ts">
+import Vue, { PropType, VueConstructor } from 'vue';
+import { MenuItem } from '@/components/MenuItem';
+import OptionsMenu from '@/components/OptionsMenu.vue';
+import isEqual from 'lodash.isequal';
+
+export default ( Vue as VueConstructor<Vue & { $refs: {
+	menu: InstanceType<typeof OptionsMenu>;
+	select: HTMLElement;
+}; }> ).extend( {
+	name: 'Dropdown',
+	data() {
+		return {
+			showMenu: false,
+		};
+	},
+	props: {
+		menuItems: {
+			type: Array as PropType<MenuItem[]>,
+			default: (): [] => [],
+		},
+		disabled: {
+			type: Boolean,
+			default: false,
+		},
+		label: {
+			type: String,
+			default: '',
+		},
+		/**
+		 * The selected menu item, can be of type `MenuItem` or `null`.
+		 *
+		 * The data usually comes from the consumer's `v-model` annotation on the Lookup component.
+		 */
+		value: {
+			type: Object,
+			default: null,
+		},
+		placeholder: {
+			type: String,
+			default: '',
+		},
+	},
+	computed: {
+		selectedItemIndex(): number {
+			if ( this.value === null || this.menuItems.length === 0 ) {
+				return -1;
+			}
+
+			return this.menuItems.findIndex(
+				( menuItem ) => { return isEqual( menuItem, this.value ); },
+				this,
+			);
+		},
+	},
+	methods: {
+		triggerKeyDown( event: KeyboardEvent ): void {
+			switch ( event.key ) {
+				case 'Enter':
+					if ( !this.showMenu ) {
+						this.startShowingTheMenu();
+						return;
+					}
+					break;
+				case 'ArrowDown':
+					this.startShowingTheMenu();
+			}
+			this.$refs.menu.onKeyDown( event );
+		},
+		async startShowingTheMenu(): Promise<void> {
+			this.showMenu = true;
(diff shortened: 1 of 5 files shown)


SaiSan-WMDE:
Excellent job, Michael! Should we wrap the other component stories with that 512px div too? What do you think, @itamargiv? It makes stories look nicer and closer to what an instance would look like in production, but I wonder if it could be somehow misleading (?)

I believe we should implement scroll events. Like in the case of the lookup, I documented that dropdown menus should display a maximum number of options ([specifically 6](https://docs.google.com/presentation/d/1GB9WcQDQDYaf_kr0RvuaAuadKLPRytdcZ1vL78N2vx0/edit#slide=id.g9f5594cff3_0_122)) rather than making available all their content right away. A persistent scroll bar should indicate whether more options are available.

micgro42:
> I believe we should implement scroll events. Like in the case of the lookup, I documented that dropdown menus should display a maximum number of options ([specifically 6](https://docs.google.com/presentation/d/1GB9WcQDQDYaf_kr0RvuaAuadKLPRytdcZ1vL78N2vx0/edit#slide=id.g9f5594cff3_0_122)) rather than making available all their content right away. A persistent scroll bar should indicate the presence of more options.

I think that are multiple different aspects mixed together here:
1. "at most 6 events should be shown, then a scroll bar should be visible." -- I think that is already the case, I can improve the story to showcase that.
1. "A persistent scroll bar should indicate the presence of more options." -- but that scrollbar should only be there iff there are more than 6 options, right? Or do you also want to see a scrollbar for 3 Options? (That can easily be done with CSS, if that is actually wanted.)
1. "Events should be emitted to let the consuming component know that the user scrolled." -- My understanding was that the Dropdown component is static in so far that the consuming component does not change the Dropdown's menu items based on the user's scroll position, but rather provides a fixed number of options from which the user can choose. Did I misunderstand something?

SaiSan-WMDE:
> The scroll events are ignored as a Dropdown always has all its possible options available.

@micgro42 I think I completely misunderstood this bit 👆  and that everything is actually in Ordnung. What I understood is that all available dropdown options were going to be displayed by default, ignoring the 6 option limitation. I completely overlooked the actual meaning of the word "events".

Answering to your points:

1. It would be great if we could showcase that! To prevent silly misinterpretations like mine

2. The persistent scroll bar should only be displayed if there are more than 6 options, that's right. (Persistent only means that it should show up right away when the menu is opened, instead of being triggered when the menu is scrolled)

3. I'm not sure where this sentence is coming from, but that only refers to the Lookup for sure, not the Dropdown (which only provides static options, yes). Let me know where you found it, so I can correct it 🙏 

Sorry about the misunderstanding!

vue-components/src/components/Dropdown.vue
```diff
@@ -0,0 +1,227 @@
+<template>
+	<div
+		:class="[ 'wikit', 'wikit-Dropdown' ]"
+		@keydown="triggerKeyDown"
+		@keyup="triggerKeyUp"
+	>
+		<label
+			class="wikit-Dropdown__label"
+			@click="$refs.select.focus()"
```
itamargiv, line 9: Nicely done! Good workaround!

vue-components/src/components/Dropdown.vue
```diff
@@ -0,0 +1,227 @@
...
+	},
+	methods: {
+		triggerKeyDown( event: KeyboardEvent ): void {
+			switch ( event.key ) {
+				case 'Enter':
+					if ( !this.showMenu ) {
+						this.showMenu = true;
+						return;
+					}
+					break;
```
itamargiv, line 109: I think this causes an issue where the menu does open on `keydown` but immediately closes (since the `OptionsMenu` registers a `keyup` event an emits a `select`...). See screen capture below:

https://drive.google.com/file/d/1oszJ2HPv4sbDdk32_nrSwNAgmtd25SsT/view?usp=sharing

itamargiv (changes requested):
Looking great, we can clean up the history a bit by dropping all the commits from up in the chain that got merged. Also I found a bug on enter, I think we should probably add a browser test to cover that scenario at least.

Besides that, good job! Thank you!

itamargiv:
Oh, one more thing, don't forget to add a line to `CHANGELOG.md` to celebrate your success :champagne: :smile_cat:

vue-components/src/components/Dropdown.vue
micgro42: Ah, got it. This is because we open the menu on key-down, but select something and close the menu on key-up. We probably should both on the same event :thinking: 

Do you or @SaiSan-WMDE have a preference for which event to use for the Enter-key behavior, Key-Down or Key-Up? I think I can also make it work for different events, but it will be more complex.

micgro42:
> Looking great, we can clean up the history a bit by dropping all the commits from up in the chain that got merged.

Ah, I expected GitHub to work its magic with that. Will do it manually, it's not much work.

I also found that the resizing of the menu options needs some more work, will look into that before I push an updated commit.

vue-components/src/components/Dropdown.vue
itamargiv: > We probably should both on the same event thinking

No, I think the problem here is the scope of the event, I think maybe the `keydown` enter should only operate on the child div itself rather than the whole parent (not entirely positive, didn't try it out)

SaiSan-WMDE: I only see the issue happening whenever `enter` is pressed to open the menu after a selection has been made. Keyboard navigation seems to work fine until then?  The desired behavior in that case would be: the menu expands and the focus is placed on the currently selected option in the list.

While checking this I realized that another keyboard navigation behavior is missing in the menu:

- If down arrow is pressed when the last option in the menu is focused, the focus moves to the first menu item.
- If up arrow is pressed when the first option in the menu is focused, the focus moves to the last item of the menu. 

I wasn't sure of this behavior (some W3C components – like [this combobox](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.1pattern/listbox-combo.html) – display it, some –like [this dropdown](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html) –  don't) and ended up documenting it only because this is how OOUI menus work (accessibility was an important focus in that project).

micgro42: I think the fix for this issue is now in #302 as the root cause for this problem already exists in master, even though it is not apparent for the Lookup.

itamargiv (approved):
\o/ Another component to the list! Thank you @micgro42 

Lisa Simpson reading "One of three components of the psyche" out of an open book

SaiSan-WMDE:
Just to be sure:

1. Triggering the menu scroll via keyboard will be incorporated when T267829 is solved, right? (I should probably update the title of the story)
2. @micgro42 you mention in the description that validation messages will be added in another commit. I guess you meant in another PR?

(The PR was merged while I was writing this)

micgro42:
> Just to be sure:
> 
>     1. Triggering the menu scroll via keyboard will be possible when [T267829](https://phabricator.wikimedia.org/T267829) is solved, right? (I should probably update the title of that ticket)

Exactly :)

>     2. @micgro42 you mention in the description that validation messages will be added in another commit. I guess you meant in another PR?

Yes, that was the intention. There is [T266546](https://phabricator.wikimedia.org/T266546) for it.

PR #299, opened 2020-11-24
Create quiet icon button tokens

SaiSan-WMDE:
This PR makes available the component-level tokens that will be used to define the visual properties of the quiet icon button.

Story: [T268658](https://phabricator.wikimedia.org/T268658)


SaiSan-WMDE:
The icon-only button tokens will be added to the single button tokens file.

SaiSan-WMDE:
The icon-only button tokens will be added to the single button tokens file.

PR #301, opened 2020-11-25
vue-components: Change the way e2e tests access a component

tzhelyazkova:
Instead of going through the DOM tree of Storybook to find the component being tested, open the story of the component in a new tab. This eliminates iframes on the page as well which makes locating elements easier.

This is a solution that also fixes local browser tests failing when Storybook has been updated to a new version.
In dev builds Storybook hijacks the URL and opens a release notes tab that does not have any components in its DOM tree.


itamargiv (approved):
:star_struck: "Test and Deploy / browser-tests (push) Successful in 2m" :confetti_ball: :champagne:

PR #302, opened 2020-11-25
Fix hovering the correct item after reopening the menu

micgro42:
By waiting till after the next tick, we can ensure that also the selectedItemIndex prop was updated, if applicable. If no item is selected, then that index should be -1 just as expected.

This PR also switches to only listening to the keydown events instead of a mixture of keydown and keyup  events. This makes the behavior now more consistent and also brings it in line with the official recommendations: https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html


itamargiv (approved):
Looking Good, thank you!

PR #304, opened 2020-11-26
Make OptionsMenu items controllable in Storybook

micgro42:


guergana:
Looks good. Why is it a draft?

micgro42:
> Looks good. Why is it a draft?

When it was created, it was unclear whether people would want it. Setting it ready for review now.

vue-components/stories/OptionsMenu.stories.ts
```diff
@@ -79,12 +74,11 @@ export function withItems( args: { boldLabels: boolean, selectedItemIndex: numbe
 
 withItems.args = {
 	boldLabels: true,
+	menuItems: vegetableItems,
```
itamargiv, line 77: This control feels a bit cluncky, which makes me question how usable it actually is, maybe we should create a couple of controls for only one item from the list? (i.e. label, description, tag)

itamargiv (changes requested):
Juhu, looks fantastic, I do have one qualm about the `menuItems` control. Also, I would ask @SaiSan-WMDE where the Options menu should go in terms of the storybook structure she envisioned.

SaiSan-WMDE:
> Also, I would ask @SaiSan-WMDE where the Options menu should go in terms of the storybook structure she envisioned.

Well, first let me say that I'm actually against displaying internal subcomponents (such as the Input and the OptionsMenu) in our Storybook 😑  In my opinion, simply referring/pointing at them from the "finished" component's documentation should be enough. 

But I understand the engineers' position, so I would say that displaying the OptionsMenu in a separate section (as suggested by this PR) might make more sense than other options: displaying the component several times in its multiple parents' sections; or creating a separate subsection to collect internal components.

micgro42:
Making this a draft again as it needs more conceptual work

PR #305, opened 2020-11-27
Extract validation message logic with composition

micgro42:
With these being the first meaningful case of code-reuse, it is a good moment to introduce the vue3 composition api via a plugin to this vue2 library.

To get the type inference for the setup option, we have to use the defineComponent function for creating our components. This has the side-effect that we have to rethink how to type the menu ref. The solution was to move this into the setup function as well because there we can use the vue3 way of accessing and _typing_ refs.

The actual validation message will be added to the Dropdown component in a follow-up pull request.

Todo:
- [x] Figure out typing of $refs
- [x] move and rename extracted code to something like `compositions/validateable.ts`
- [x] add tests for extracted methods!

Bug: [T266546](https://phabricator.wikimedia.org/T266546)

### DIFF
--- a/vue-components/tests/unit/compositions/validateable.ts
+++ b/vue-components/tests/unit/compositions/validateable.ts
@@ -1,0 +1,29 @@
+import { getFeedbackTypeFromProps, errorProp } from '@/compositions/validatable';
+
+describe( 'validatable composition', () => {
+	it( 'gets the feedback type from an error prop', () => {
+		const feedbackType = 'warning' as const;
+		const testProps = {
+			error: {
+				type: feedbackType,
+				messsage: 'something',
+			},
+		};
+
+		expect( getFeedbackTypeFromProps( testProps ) ).toBe( feedbackType );
+	} );
+
+	describe( 'error prop validation', () => {
+		it( 'accepts "warning" error type', () => {
+			expect( errorProp.validator( { type: 'warning' } ) ).toBe( true );
+		} );
+
+		it( 'accepts "error" error type', () => {
+			expect( errorProp.validator( { type: 'error' } ) ).toBe( true );
+		} );
+
+		it( 'disallows other error types', () => {
+			expect( errorProp.validator( { type: 'notice' } ) ).toBe( false );
+		} );
+	} );
+} );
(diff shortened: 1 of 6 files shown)


itamargiv (commented):
This is looking good, I have one comments though:

I think perhaps the filename is not very succinct as it doesn't actually describe what the composition is meant for (it's not the message that is composed, it's the pieces of logic surrounding the validation), also I wouldn't necessary consider it as util as it's a bit more than purely agnostic helper functions. How about putting it in an addition directory `compositions/validateable.ts`.

Also, we might need a test for `getFeedbackTypeFromProps`

micgro42:
Thank you for your feedback! I will implement it tomorrow (have to leave now)

itamargiv (changes requested):
Looking great, I think I might have made a typo when I suggested `validateable` (It should be `validatable`, sorry :/), also, there's one more thing to change

vue-components/tests/unit/compositions/validateable.ts
```diff
@@ -0,0 +1,29 @@
+import { getFeedbackTypeFromProps, errorProp } from '../../../dist/components/util/ValidationMessageComposition';
```
itamargiv, line 1: Forgot to add to to the review above the path here should be to `@/compositions/validateable`

micgro42: Good catch! I will fix it right away

micgro42: Wait, `components/util/ValidationMessageComposition` doesn't exist anymore in this commit. How did this even work and pass CI? :upside_down_face:

itamargiv: Oy vay, I think it's because dist is still cached...

micgro42:
> Looking great, I think I might have made a typo when I suggested `validateable` (It should be `validatable`, sorry :/), also, there's one more thing to change

I think and hope that both are fixed now

vue-components/tests/unit/compositions/validateable.ts
micgro42: Mh, we may want to look into that again. But this code should be fixed now

itamargiv (approved):
Thank you, great job!

PR #306, opened 2020-11-27
Add Error and Warning state to Dropdown

micgro42:
The `menuItems` `PropType` is removed from the prop definition as it seems to be completely ignored. However, the menuItem prop must be defined in the signature of the setup function even if it is not used inside the setup function to receive typing in the computed methods added via the usual options API.

Bug: T266546

### DIFF
--- a/vue-components/src/components/Dropdown.vue
+++ b/vue-components/src/components/Dropdown.vue
@@ -8,7 +8,7 @@
 			@click="$refs.select.focus()"
 		>{{ label }}</label>
 		<div
-			class="wikit-Dropdown__select"
+			:class="classesForSelect"
 			:tabindex="!disabled && '0'"
 			:disabled="disabled"
 			@click="onClick"
@@ -39,28 +39,43 @@
 			@esc="onEsc"
 			ref="menu"
 		/>
+		<ValidationMessage
+			v-if="error"
+			:type="error.type"
+			:message="error.message"
+		/>
 	</div>
 </template>
 
 <script lang="ts">
-import Vue, { PropType, VueConstructor } from 'vue';
+import Vue from 'vue';
+import VueCompositionAPI, { defineComponent, computed, ref } from '@vue/composition-api';
 import { MenuItem } from '@/components/MenuItem';
 import OptionsMenu from '@/components/OptionsMenu.vue';
+import { getFeedbackTypeFromProps, errorProp, ErrorProp } from '@/compositions/validatable';
 import isEqual from 'lodash.isequal';
+import ValidationMessage from '@/components/ValidationMessage.vue';
 
-export default ( Vue as VueConstructor<Vue & { $refs: {
-	menu: InstanceType<typeof OptionsMenu>;
-	select: HTMLElement;
-}; }> ).extend( {
+Vue.use( VueCompositionAPI );
+
+export default defineComponent( {
 	name: 'Dropdown',
-	data() {
+	setup( props: { error: ErrorProp; menuItems: MenuItem[]; value: MenuItem|null } ) {
+		const menu = ref<InstanceType<typeof OptionsMenu>|null>( null );
+		const showMenu = ref( false );
+
+		const feedbackType = computed( getFeedbackTypeFromProps( props ) );
+
 		return {
-			showMenu: false,
+			showMenu,
+			feedbackType,
+			menu,
 		};
 	},
 	props: {
+		error: errorProp,
 		menuItems: {
-			type: Array as PropType<MenuItem[]>,
+			type: Array,
 			default: (): [] => [],
 		},
 		disabled: {
@@ -96,6 +111,13 @@ export default ( Vue as VueConstructor<Vue & { $refs: {
 				this,
 			);
 		},
+		classesForSelect(): string[] {
+			const classes = [ 'wikit-Dropdown__select' ];
+			if ( this.feedbackType ) {
+				classes.push( `wikit-Dropdown__select--${this.feedbackType}` );
+			}
+			return classes;
+		},
 	},
 	methods: {
 		triggerKeyDown( event: KeyboardEvent ): void {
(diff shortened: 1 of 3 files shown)


micgro42:
This should now be fine to be reviewed, some maybe flaky browser tests notwithstanding.

vue-components/src/components/Dropdown.vue
```diff
@@ -39,28 +39,64 @@
...
+					if ( !showMenu.value ) {
+						startShowingTheMenu();
+						return;
+					}
+					break;
+				case 'ArrowDown':
+					startShowingTheMenu();
+			}
+			menu.value?.onKeyDown( event );
+		}
```
itamargiv, line 85: Why was all this code moved into the composition setup?

micgro42: We needed to define the refs in the composition setup to get the TypeScript type working. And I _think_ the problem may afterwards have been how to access them, but I'll check that again.

micgro42: Yes, moving those things back is working now! The diff should now be much smaller :)

micgro42: Or so I thought until I saw the CI complaining. Will have another look this afternoon.

micgro42: Ok, so no. This has to happen inside the `setup`, because only there the ref `menu` has the correct type. It doesn't infer its type correctly when used outside of the setup.

I added a line to the commit message abut that

micgro42: If you want, you can play around with the problem in https://github.com/wmde/wikit/tree/badlyTypedRef

itamargiv: I'm not sure if I like this pattern, I will play around with it some more tomorrow to see what can be done to mitigate this... I wonder if Vue 3 can handle this better without relying on pulling everything into setup just because of a ref..

micgro42: Yeah, I suspect this might be a limitation of the plugin. Happy to find a nicer alternative.

OTOH, I could also see us (in the future) transition to the setup-only way of writing single file components:
```
<script setup>
  import { ref } from 'vue'

  export const count = ref(0)
  export const inc = () => count.value++
</script>
```
https://github.com/vuejs/rfcs/blob/sfc-improvements/active-rfcs/0000-sfc-script-setup.md

itamargiv: That actually looks beautiful! I'd love it if this rfc is accepted

itamargiv: I guess this would be alright for now, but I would open a ticket to investigate this topic further.

micgro42:
GitHub thought that there were conflicts, but I could rebase it on top of master without problems. It should be ready to review again

jakobw (changes requested):
-1 to get you to look at the attempt to move the menu ref back into the methods, but otherwise looks solid!! :100: 

I thought I saw two commits here earlier today. Did that get squashed into one or did one disappear? (or did I make it up?)

micgro42:
> I thought I saw two commits here earlier today. Did that get squashed into one or did one disappear? (or did I make it up?)

This PR was previously based on top of #305. When that merged, GitHub changed the target automatically to master. Though that doesn't always work well and so the commit from #305 may have still been visible for this PR.
I merely did a local rebase of this branch, not a squash.

vue-components/src/components/Dropdown.vue
```diff
@@ -90,32 +126,20 @@ export default ( Vue as VueConstructor<Vue & { $refs: {
...
-			switch ( event.key ) {
-				case 'Enter':
-					if ( !this.showMenu ) {
-						this.startShowingTheMenu();
-						return;
-					}
-					break;
-				case 'ArrowDown':
-					this.startShowingTheMenu();
+		selectClass(): string[] {
```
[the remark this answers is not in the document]
micgro42: Could you link me to the conventions that you are referring to?

Note that `select` in that function name is not a verb but refers to that specific div which is the equivalent to the `<select>` element for native HTML in order to differentiate it from the classes for the label or the menu.

But I agree that it is very ambiguous and at the very least should be `selectClasses`. If you have recommendation for how to make less ambiguous, I'm happy to hear them :)
Maybe `classesForSelect()`? :thinking:

vue-components/src/components/Dropdown.vue
micgro42: I can't believe I looked twice into this and didn't figure it out! :facepalm: 

Thanks! Will update the PR

vue-components/src/components/Dropdown.vue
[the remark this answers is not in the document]
jakobw: Ah sorry, I totally read "select" as a verb. `selectClasses` or `classesForSelect` are fine!

> Could you link me to the conventions that you are referring to?

I meant the general Vue convention (i.e. same as in the [docs](https://vuejs.org/v2/guide/computed.html#Basic-Example)) to name computed properties like properties, not like methods or functions, but that was already the case here :)

jakobw (commented):
> This PR was previously based on top of #305. When that merged, GitHub changed the target automatically to master. Though that doesn't always work well and so the commit from #305 may have still been visible for this PR.
I merely did a local rebase of this branch, not a squash.

Got it. I think I stared at the already merged commit for a few minutes in the morning and then got confused when the code looked different in the evening. The way github deals with PR chains is silly sometimes :man_facepalming:

micgro42:
@jakobw I updated the PR with your suggested solution, please have another look when you have the time :)

jakobw (approved):
Wooo looks good!! :rocket:

SaiSan-WMDE:
Looks 💯 

Just a heads-up: the disabled dropdown should be included in an "All" page. This page should also display the basic and validation states (as in TextInput - All). Check updated [SB structure](https://docs.google.com/document/d/1loENZ13z5s_Nes4exm5Z9d5S3PnR7F_0brv5azQ5XB4/edit?usp=sharing).

PR #307, opened 2020-11-30
Fix eslint comment

lucaswerkmeister:
There’s already an eslint-disable earlier in the file, so I assume this one is meant to reenable the rule, now that the inline SVGs are done.


micgro42 (approved):

itamargiv:
Thank you!

PR #310, opened 2020-11-30
Add .nvmrc with Node version

lucaswerkmeister:
This makes it a bit easier (run `nvm use`) to use the correct Node version without Docker. (The version number matches the Dockerfile.)


itamargiv (approved):
:pray: Thank you!

lucaswerkmeister:
Hehe, I thought there might be more pushback against accommodating non-Docker users :P

manicki:
> Hehe, I thought there might be more pushback against accommodating non-Docker users :P

oh there is! Go Docker or Go Home fraction is still here

PR #311, opened 2020-11-30
Add package.json to cache key

itamargiv:
While working on #290 I noticed that I forgot to add `package.json` to the cache, this ensures the cache invalidates if anything changes in `package.json` as well


tzhelyazkova (approved):
makes sense :+1:

PR #315, opened 2020-12-07
fix dropdown spacing

SaiSan-WMDE:
This adds a 12px margin to the left of the expand icon so it's properly separated from the dropdown label.

TBD whether it would be necessary to apply `white-space: nowrap` to the dropdown text in order to prevent unwanted wrapping after this change.


itamargiv (approved):
Fantstic! thanks!

PR #322, opened 2020-12-14
[WiP] Added slot and prop for icons in Button component

guergana:
- Add named slot to the Button component
- both slots should be wrapped in <span> with classes
- add prop "iconOnly" to Button, defaulting to false
- add a button with an icon and a label to storybook
- add a button with "iconOnly=true" to storybook


guergana:
there were big changes [in the task](https://phabricator.wikimedia.org/T269141). will start another one.

PR #323, opened 2020-12-14
Remove need to have custom built Docker image

tarrow:
This is a draft PR to attempt to demonstrate the kind of direction we could move in to have everyone use the same docker image rather than all building them independently

Bug: https://phabricator.wikimedia.org/T269423


tzhelyazkova (commented):
Hey @tarrow :wave: I get an error locally when I try `docker-compose up` on this branch.
Is there something else that needs to be changes or passed to the command?

tzhelyazkova:
> Hey @tarrow wave I get an error locally when I try `docker-compose up` on this branch.
> Is there something else that needs to be changes or passed to the command?

```
Recreating 89cfa92dbfe6_wikit_node_1 ... 
Recreating d2156f85d1ef_wikit_storybook-vue_1 ... 
Recreating 1be5a07d4b99_wikit_storybook-docs_1 ... 

ERROR: for 89cfa92dbfe6_wikit_node_1  a bytes-like object is required, not 'str'

ERROR: for 1be5a07d4b99_wikit_storybook-docs_1  a bytes-like object is required, not 'str'

ERROR: for d2156f85d1ef_wikit_storybook-vue_1  a bytes-like object is required, not 'str'
```
and some more when I scroll
```
docker.errors.APIError: 500 Server Error: Internal Server Error ("b'unable to find user 1000 # Users need to set this to `id -u`: no matching entries in passwd file'")
```

addshore:
> > Hey @tarrow wave I get an error locally when I try `docker-compose up` on this branch.
> > Is there something else that needs to be changes or passed to the command?
> 
> ```
> Recreating 89cfa92dbfe6_wikit_node_1 ... 
> Recreating d2156f85d1ef_wikit_storybook-vue_1 ... 
> Recreating 1be5a07d4b99_wikit_storybook-docs_1 ... 
> 
> ERROR: for 89cfa92dbfe6_wikit_node_1  a bytes-like object is required, not 'str'
> 
> ERROR: for 1be5a07d4b99_wikit_storybook-docs_1  a bytes-like object is required, not 'str'
> 
> ERROR: for d2156f85d1ef_wikit_storybook-vue_1  a bytes-like object is required, not 'str'
> ```
> 
> and some more when I scroll
> 
> ```
> docker.errors.APIError: 500 Server Error: Internal Server Error ("b'unable to find user 1000 # Users need to set this to `id -u`: no matching entries in passwd file'")
> ```

Did you add the correct value of your uid and gid to the `.env` file?

I'm going to work on a PR that proposes moving a bunch of this stuff around to get away from having this docker image at all as the current state of things here, users, docker image and dependencies is a little confusing to me

addshore:
I think we should aim to get rid of this Dockerfile entirely

addshore:
From the commit msg:

Tarrow started down the route of removing the need to
build a custom docker image in this repo, and I have taken
over the branch with this change fully removing the Dockerfile.

Change summary:
 - Remove Dockerfile, just using the base image now
 - Make all lerna commands use lerna from node_modules rather
   than any globally installed version.
 - Rework readme to just specify npm commands (not via docker-compose)
 - Add lerna command to package.json
 - Stop nightwatch from starting chromedriver when it is not needed

README moving away from docker-compose:
In other WMDE repos we moved away from including READMEs stating how
to run tooling via docker. Instead just using the much shorter, easier
to read base commands.
It would then be up to an individual to run npm, or other tooling in
a docker container if they want to for their development setup.

lerna usage:
The Dockerfile loaded lerna globally, and this was then inconsistently
used throughout various npm commands and README suggestions.
When reviewing other projects that use lerna they all seem to specify
the lerna version in package.json and just use the version installed
there, so I went for that approach.

nightwatch chromedriver:
Before this change nightwatch would start a chromedriver instance in
all situations, even when not used. It is not use for most situations
such as docker or saucelabs.
For now I have only set this to not start for the docker command.
This was needed as having chromedriver start means extra deps are needed
in the docker container.

addshore:
@tarrow seemingly I can't add you as a reviewer as this PR was yours, so here is your personal ping.

itamargiv (commented):
Personally, you know what I think about docker commands :stuck_out_tongue_closed_eyes:, I would defer to the original hikers on this one though, since they made a conscious decision to have docker as the "official" way of working, what do you think the a-team (@tzhelyazkova m @jakobw)? I really just need your eyeballs on this. 

If no response or strong feelings, I'm up for merging this.

tarrow:
In general I have no strong objections to the following parts:
* Remove Dockerfile, just using the base image now
* Make all lerna commands use lerna from node_modules rather than any globally installed version.
* Add lerna command to package.json
* Stop nightwatch from starting chromedriver when it is not needed

These are basically going in the direction I was heading for: a consistent development environment.


---

"Rework readme to just specify npm commands (not via docker-compose)" is not a direction I'm personally keen to move in though (happy to be convinced).

I think we should aspire to having a situation with a recommended development environment. Using whatever random version of `npm` you have in your path at the moment of running is heading for a more diverse env.

I'm interested that we are doing this: "In other WMDE repos we moved away from including READMEs stating how to run tooling via docker. Instead just using the much shorter, easier to read base commands." I'd be interested in seeing where we made this decision (maybe I just missed it)

itamargiv:
> I think we should aspire to having a situation with a recommended development environment. Using whatever random version of `npm` you have in your path at the moment of running is heading for a more diverse env.

Actually not really, `nvm` has been around for quite a long time, and it does a great job in ensuring the correct node version is used regardless of the user's preferences. We even made a point of including an `.nvmrc` file in our repo for that purpose in particular. Of course, people who prefer to use docker still may use it. But it's really not the only way of ensuring all environments use the same version of node.

In another point, at least when it comes to my experience with front end development and libraries, most packages there tend to avoid including docker commands in their Dev README. I'm not sure what individual motive is behind each library's description, but I can tell you this: As someone who writes occasionally, and as someone who has taught beginners, there is more chance that you would be more inviting for the general frontend community, if you work with tools they know already.

The barrier of entry to nvm and npm for someone who constantly works with JavaScript is very low, as opposed to requiring them to install docker, or at the very least obscuring the fact that they can perform everything they know without having to investigate each part of a lengthy and obtuse docker-compose command.

TL;DR IMHO npm commands are simpler to understand and are more inviting for beginners and people in the JavaScript development realm, which can result in more volunteers.

tarrow:
> Actually not really, `nvm` has been around for quite a long time, and it does a great job in ensuring the correct node version is used regardless of the user's preferences. We even made a point of including an `.nvmrc` file in our repo for that purpose in particular. Of course, people who prefer to use docker still may use it. But it's really not the only way of ensuring all environments use the same version of node.
> 

I was trying to include using `nvm` (and trendy alternatives like `volta`) in "Using whatever random version of npm you have in your path" but I wasn't very clear.

I personally think we should have a recommended way to develop our code in a reproducible environment. I could imagine us moving to saying that `nvm` is the way but we should probably offer either `nvm` or `docker` but not both.

As far as I know, most people at WMDE aren't using `nvm`. I think this is because running a sandboxed version of it isn't trivial and we generally discourage running `node` on metal. This is because of the neverending discussion about packages on npm not being signed and we don't want them possibly stealing our ssh keys etc..

Also I'm unsure if `nvm` allows us to fix the chromium and chromedriver version in the same way that we are able to by using the docker image.

> In another point, at least when it comes to my experience with front end development and libraries, most packages there tend to avoid including docker commands in their Dev README. I'm not sure what individual motive is behind each library's description, but I can tell you this: As someone who writes occasionally, and as someone who has taught beginners, there is more chance that you would be more inviting for the general frontend community, if you work with tools they know already.

This seems to reflect my experience as well of the general front-end community. I think this package would be most approachable to a new volunteer dev if it just used `npm`

> The barrier of entry to nvm and npm for someone who constantly works with JavaScript is very low, as opposed to requiring them to install docker, or at the very least obscuring the fact that they can perform everything they know without having to investigate each part of a lengthy and obtuse docker-compose command.
> 
> TL;DR IMHO npm commands are simpler to understand and are more inviting for beginners and people in the JavaScript development realm, which can result in more volunteers and easier onboarding.

I think that given this is *mostly* used by WMDE devs who already have a docker based development environment then the software barrier to entry isn't so big.

I could see an argument for providing a vanity script called `./node` or `./npm` that hides the `docker-compose` command if the main objections to the current state is readability of the very ugly long `docker-compose` lines

tzhelyazkova (commented):
I think we all agree with the initial change in this PR - remove need to have custom built Docker image.
The discussion if the README should be changed is larger than the context of this PR.
How about we keep this PR with the old README and move the discussion to a separate PR or maybe ADR proposal if we want docker to be the standard for all FE projects. @addshore @tarrow

manicki:
@tzhelyazkova has a good point, the broader topic conversation should move out of the pull request review. I will moderate some discussion on the to docker or not to docker topic

README.md
```diff
@@ -54,9 +41,12 @@ To run browser tests locally make sure `storybook-vue` is up and running.
...
 
+Note: If you were not already running the main docker-compose services your tests may fail their first run as things are still being built.
+
 ##### On the host machine
 
-Navigate to `vue-components` subdirectory and run it with your storybook URL specified, e.g.
+Navigate to `vue-components` subdirectory and run it with your storybook URL specified:
+
 ```sh
 STORYBOOK_URL=localhost:6005 npm run e2e
```
addshore, line 51: This command did not used docker-compose before this change so I am leaving it as it is.

addshore:
Picking out a few points from the conversation above around the docker or not docker in readme topic.

> This seems to reflect my experience as well of the general front-end community. I think this package would be most approachable to a new volunteer dev if it just used npm

+1

> As far as I know, most people at WMDE aren't using nvm. I think this is because running a sandboxed version of it isn't trivial and we generally discourage running node on metal. This is because of the neverending discussion about packages on npm not being signed and we don't want them possibly stealing our ssh keys etc..

Indeed, and IMO for the people that it concerns they should deal with this on their system, rather than us bending documentation that is meant for everyone to cover this case.
If there are other valid reasons to use docker in the README then thats a seperate point, but I don't see many cases where it makes a whole lot of sense, with the exception of services like selenium hub for example.

> @tzhelyazkova has a good point, the broader topic conversation should move out of the pull request review. I will moderate some discussion on the to docker or not to docker topic

Sounds good!

addshore:
Pinging @tarrow for a re review, as I can't add him as a reviewer, as this PR was his to start with.

itamargiv (approved):
:+1: LGTM, leaving open to give the others a chance to have a look as well.

tzhelyazkova (approved):

tzhelyazkova:
I will merge this EOD if noone objects till then.
Anything after that can be addressed in a follow up PR.

manicki:
Please do. I somehow thought it got merged weeks ago.

PR #325, opened 2020-12-16
Use netlify for branch hosting the docs storybook

jakobw:
This PR finally brings back branch hosting for the composed (tokens+docs+vue-components) storybook :tada: 

I haven't found a good way/place to show the netlify link. I added a "Print deploy URL" step in the `deploy` job, which may be good enough for now? `netlify-cli` advertises an `--alias` flag for predictable deployment URLs which looks like a good option to create deployments on a per-branch instead of a per-commit basis so that it can be permanently attached to a PR, but I didn't want to recreate [past problems](https://phabricator.wikimedia.org/T262270) by building the branch name into the URL.

Bug: [T261650](https://phabricator.wikimedia.org/T261650)

### DIFF
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -74,7 +74,7 @@ jobs:
     - name: Test
       run: npm run test
 
-  deploy:
+  ui-review:
     runs-on: ubuntu-latest
     needs: setup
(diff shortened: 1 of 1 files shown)


tzhelyazkova (approved):
:rocket:

PR #326, opened 2020-12-16
Run browser tests against netlify

jakobw:
Our CI browser tests often fail or time out due to problems with Chromatic and running tests against the storybook preview was arguably abusing the service to begin with. Pointing them at netlify will hopefully result in more stable browser test results.


guergana (dismissed):
Looks great!! Good job!

guergana (approved):
Approving again!

PR #329, opened 2020-12-17
Restructure Button props

micgro42:
We want to support buttons that are of normal variant and progressive type.

Added a method to the updated hook to throw an error if an unsupported combination of variant and type is used by the application. For typing reasons this had to happen inside the Vue3 setup method.

### DIFF
--- a/vue-components/src/components/Button.vue
+++ b/vue-components/src/components/Button.vue
@@ -1,11 +1,14 @@
 <template>
-	<button :class="[ 'wikit', 'wikit-Button', `wikit-Button--${ type }` ]">
+	<button :class="[ 'wikit', 'wikit-Button', `wikit-Button--${ type }`, `wikit-Button--${ variant }` ]">
 		<slot />
 	</button>
 </template>
 
 <script lang="ts">
 import Vue from 'vue';
+import VueCompositionAPI, { defineComponent, onMounted } from '@vue/composition-api';
+
+Vue.use( VueCompositionAPI );
 
 /* eslint-disable no-trailing-spaces */
 /**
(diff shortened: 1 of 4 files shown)


vue-components/src/components/Button.vue
```diff
@@ -19,20 +26,50 @@ import Vue from 'vue';
...
 			},
 			default: 'neutral',
 		},
+		/**
+		 * @values normal, primary
+		 */
+		variant: {
+			type: String,
+			validator( value: string ): boolean {
+				return [ 'normal', 'primary' ].includes( value );
```
Ladsgroup, line 50: Should it include quiet too?

micgro42: not yet, I think that will be added in your PR

micgro42: but eventually yes :)

Ladsgroup: ah I see.

Ladsgroup (approved):

vue-components/src/components/Button.vue
```diff
@@ -19,19 +22,49 @@ import Vue from 'vue';
...
+				normal: [ 'neutral' ],
+				primary: [ 'progressive', 'destructive' ],
+			};
+			if ( !supportedCombinations[ props.variant ].includes( props.type ) ) {
+				throw new Error(
+					`The combination of variant "${props.variant}" and type "${props.type}" is not yet supported!`,
+				);
+			}
+		} );
 	},
```
jakobw, line 68: I was hoping this could move into the prop validation but apparently it's not possible to access a prop in another prop's validator. Is there a benefit to doing this in `setup`? Why not just put it in `mounted`?

```suggestion
	mounted() {
		const supportedCombinations = {
			normal: [ 'neutral' ],
			primary: [ 'progressive', 'destructive' ],
		};
		if ( !supportedCombinations[ this.variant ].includes( this.type ) ) {
			throw new Error(
				`The combination of variant "${this.variant}" and type "${this.type}" is not yet supported!`,
			);
		}
	},
```

jakobw (commented):
Left a question, but otherwise this is looking very good \o/

vue-components/src/components/Button.vue
micgro42: That was my first attempt as well^^

I tried to explain it in the commit message. For me TypeScript is complaining that the string in `this.variant` does not index the object `supportedCombinations` and therefore `supportedCombinations[ this.variant ]` has type any and that is disallowed.

And since one of the explicitly stated benefits/purposes of the setup function is better type support I decided to use that instead of trying to weirdly typecast things somehow or use `// ts-disable`.

But maybe I did something slightly wrong and your solution works with TypeScript? Sometimes it is strange what causes type errors.

jakobw: I thought it could be managed with a `PropType` on `variant` and `type` and a corresponding index type on `supportedCombinations`, but I also struggled with getting TS to properly recognize the `PropType` and gave up. Strange, I thought I've seen a similar case somewhere.

jakobw (approved):
:rocket:

PR #330, opened 2020-12-17
Add a color to have the icon reuse the parent text color

micgro42:
This reuses the color in the css `color` attribute of the parent.

Bug: [T270330](https://phabricator.wikimedia.org/T270330)

### DIFF
--- a/vue-components/src/components/iconProps.ts
+++ b/vue-components/src/components/iconProps.ts
@@ -9,6 +9,6 @@ export enum IconTypes {
 	INFOOUTLINED = 'info-outlined',
 }
 
-export const iconColors = [ 'base', 'warning', 'error', 'notice', 'success' ];
+export const iconColors = [ 'base', 'warning', 'error', 'notice', 'success', 'inherit' ];
 
 export const iconSizes = [ 'medium', 'large' ];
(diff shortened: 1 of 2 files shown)


vue-components/src/components/iconProps.ts
```diff
@@ -9,6 +9,6 @@ export enum IconTypes {
 	INFOOUTLINED = 'info-outlined',
 }
 
-export const iconColors = [ 'base', 'warning', 'error', 'notice', 'success' ];
+export const iconColors = [ 'base', 'warning', 'error', 'notice', 'success', 'current' ];
```
Ladsgroup, line 12: Maybe using "inherit"? is there any place that we can look for inspiration?

micgro42: Yeah, I am split between `inherit` and `current`. `inherit` implies inheriting any style in CSS, but `currentColor` is the magic color in SVG that tells it to use the text color

micgro42: After thinking about it a bit more, I like `inherit` as well. I changed it to that :)

Ladsgroup (approved):

vue-components/src/components/iconProps.ts
jakobw: What about calling it `chameleon` because it makes the icon change color based on its surroundings?
(kidding, `inherit` sounds good)

PR #331, opened 2020-12-18
Use <br> instead of trailing double-whitespace

jakobw:
HTML tags in doc comments aren't pretty, but many IDEs remove trailing
whitespaces automatically.


micgro42 (approved):
loving it!

PR #333, opened 2020-12-18
Comment deploy URL to PR

jakobw:
... so that we don't have to dig it out of the job output.


jakobw:
:facepalm: close enough. That is certainly a URL, but not the right one.

PR #334, opened 2020-12-18
Comment deploy URL to PR

jakobw:
**Please review #335 first**

This will comment the netlify deploy URL to the PR as soon as it's opened, which makes it more visible than it was before.

The way the URL is being generated is a little adventurous, but it should work as long as our netlify site isn't renamed.

... and ignore the two failed comments by the bot. For future PRs there will only be one comment with a URL that will remain the same (because of #335) even when the PR is updated.

### DIFF
--- a/.github/workflows/comment-deploy-url-to-pr.yaml
+++ b/.github/workflows/comment-deploy-url-to-pr.yaml
@@ -1,0 +1,19 @@
+name: Comment deploy URL to PR
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  comment-on-pr:
+    runs-on: ubuntu-latest
+    name: Comment deploy URL to PR
+    steps:
+      - name: Get branch hash
+        id: get_hash
+        run: echo "::set-output name=hash::$(echo 'refs/heads/${{ github.head_ref }}' | shasum | awk '{print $1}')"
+      - name: Comment on PR
+        uses: thollander/actions-comment-pull-request@1.0.1
+        with:
+          message: "<https://${{ steps.get_hash.outputs.hash }}--wmde-wikit.netlify.app>"
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


micgro42:
What use do you have in mind for that?

jakobw:
> What use do you have in mind for that?

I'm not sure I understand the question. Are you asking what use we have for branch hosting in general or what use this comment is?

micgro42:
> > What use do you have in mind for that?
> 
> I'm not sure I understand the question. Are you asking what use we have for branch hosting in general or what use this comment is?

My understanding is that we already always have direct link to a storybook in every PR where it is relevant. I can see that we want to host it on netlify for browser-tests, but I'm not sure what we as humans gain from that second storybook?

image

jakobw:
> > > What use do you have in mind for that?
> > 
> > 
> > I'm not sure I understand the question. Are you asking what use we have for branch hosting in general or what use this comment is?
> 
> My understanding is that we already always have a direct link to a storybook in every PR where it is relevant. I can see that we want to host it on netlify for browser-tests, but I'm not sure what we as humans gain from that second storybook?

We didn't move to netlify for the browser tests. [The ticket](https://phabricator.wikimedia.org/T261650) was actually created long before there were any problems with the browser tests. The goal was to make a branch previews available for the composed (docs + tokens + vue-components) storybook, whereas Chromatic only shows the components. As far as I know we want the netlify link, and keep chromatic primarily for visual regression tests.

SaiSan-WMDE:
>  I'm not sure what we as humans gain from that second storybook?

Hosting the Docs Storybook in Netlify allows previewing changes to the token documentation online. Chromatic only hosts the components Storybook, so relevant token changes (e.g. removing unnecessary cursor tokens) are not reflected.

micgro42:
Ah, so there is a ticket for this. Yes that motivation makes sense, thanks!

tzhelyazkova (dismissed):

tzhelyazkova (approved):

PR #338, opened 2021-01-04
Export Icon component

micgro42:
To be able to to create the icon-only buttons as they are showcased in Storybook, the consuming application needs access to the Wikit Icon component.
Therefore, it can no longer be an internal component.


SaiSan-WMDE (approved):

micgro42:
Meh, I forgot to add Information about this to the Changelog :/

PR #339, opened 2021-01-04
Add information about using aria-label to icon only buttons

micgro42:
Also, we are being a good role-model by using it ourselves.

Bug: [T269140](https://phabricator.wikimedia.org/T269140)

### DIFF
--- a/vue-components/stories/Button.stories.ts
+++ b/vue-components/stories/Button.stories.ts
@@ -41,17 +41,22 @@ export function iconOnly(): Component {
 		components: { Button, Icon },
 		template: `
 			<div>
+				<p>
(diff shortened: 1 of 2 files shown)


guergana (approved):
Looks good.

tzhelyazkova (approved):

SaiSan-WMDE:
I didn't find a quick way to correct the font style of the paragraph added to the Docs page. While doing that, I realized that the button stories need to be reorganized (quiet is not a story of button), and the Button Doc page corrected: so I'd say we can move ahead (this PPR gets the main job of adding labels done) and I'll make changes in a different PR.

PR #340, opened 2021-01-04
vue-components: Add "display: block" to Input component

Ladsgroup:
Without it, the input gets extremely long when there something after it
to support the "width: 100%" css rule. With this, it gets just as long
as needed.


guergana (commented):
This fixes the problem partly, the field still gets bigger up the width of the text in the ValidationMessage. Is this what we want?

Ladsgroup:
Yeah, let's do it in https://github.com/wmde/query-builder/pull/125

PR #341, opened 2021-01-05
vue-components: Add native type prop to Button

tzhelyazkova:
Since the name 'type' is already taken the native html
`type` attribute will have its value passed as `nativeType`.

Bug: [T267938](https://phabricator.wikimedia.org/T267938)


micgro42:
I think this deserves an entry in the Changelog as well, probably in the "new features" section

tzhelyazkova:
> I think this deserves an entry in the Changelog as well, probably in the "new features" section

I totally forgot about it, thanks!

micgro42 (approved):

PR #342, opened 2021-01-05
Reorganize button stories

SaiSan-WMDE:
This collects the button types and variants under the main component, and makes sure that they are correctly reflected as button stories in a single Button Doc page. I'm not 100% convinced about that crowded Icon Only page. This could be considered a first step. The alternative would be to create Icon Only subpages for all button types as originally suggested.

Early feedback on the grouping would be very much appreciated.

TBD: Find a way to add descriptions to individual stories in the Docs page, to avoid displaying text in the Canvas page – see current Icon only page. (I couldn't make any of the solutions described in [this thread](https://github.com/storybookjs/storybook/issues/8527) work.)


itamargiv (approved):
Looking good to me, if tests pass. @SaiSan-WMDE I also added a commit to exemplify that the variant and type property are **not** always needed and that the default Button is Normal-Neutral.

PR #345, opened 2021-01-05
eslint: Disable duplicate eslint rule for indentation

tzhelyazkova:
The eslint rules `@typescript-eslint/indent` and `indent` are the same.
This commit turns off  `@typescript-eslint/indent`.


vue-components/.eslintrc.js
```diff
@@ -45,6 +45,9 @@ module.exports = {
 		'@typescript-eslint/no-require-imports': 'off',
 		'@typescript-eslint/no-var-requires': 'off',
 
+		// duplicate with eslint's indent
+		'@typescript-eslint/indent': 'off',
```
micgro42, line 49: Wait, but shouldn't we turn off `indent` instead? At least that is how I understand [typescript-eslint/indent.md](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/indent.md). Or are we affected by the issue mentioned there?

tzhelyazkova: I overlooked this, you're right.

micgro42 (approved):
The visual change is probably just a missing rebase or something like that.

PR #347, opened 2021-01-05
vue-components: Fix running browser tests locally

tzhelyazkova:
The global afterEach function runs after each test *suite* which
is not ideal when we want to send info for each test that was executed.

The logic is moved to a more appropriate place - a custom command.
This means every test suite needs to define an afterEach hook where the
custom command is called before ending the session.

This improvement fixes the problems with running browser tests locally (docker or not).

Bug: T268975, T266343


guergana:
This is awesome!

guergana (approved):
works beautifully.

PR #349, opened 2021-01-05
Update CHANGELOG.md

SaiSan-WMDE:
Trying to document all relevant changes retrospectively in preparation for the future release of v 2.0.0. Additions and corrections are totally welcomed 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,17 +21,17 @@ development)_
 - Additional icons to the Icon component
 - New tokens based on wikimedia-ui-base v 0.18.0
 - New Dropdown Component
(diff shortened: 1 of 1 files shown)


Ladsgroup (approved):

tzhelyazkova:
The commit was rebased to resolve conflicts with master.
I removed the Netlify bit intentionally, because it's related to internal DS infrastructure, rather than changes that are important to a consumer of the DS.
I added "breaking change" in front of the Input.json line because technically if someone is using only the tokens package, this change will break their app.

tzhelyazkova (approved):

PR #350, opened 2021-01-08
Create toggle button component

SaiSan-WMDE:
This makes available the tokens that will be used to style the toggle button.

Issue: [T271788](https://phabricator.wikimedia.org/T271788)

The **Toggle Button** component was build on top of this PR.

Issue: [T271788](https://phabricator.wikimedia.org/T271788)


tokens/properties/components/ToggleButton.json
```diff
@@ -0,0 +1,132 @@
...
+            },
+            "border-color": {
+                "value": "{border.color.progressive.active.value}"
+            },
+            "background-color": {
+                "value": "{background.color.progressive.active.value}"
+            },
+            "focus": {
+                "box-shadow": {
+                    "value": "{box-shadow.inset.focus.button-primary.value}"
```
SaiSan-WMDE, line 116:
```suggestion
                    "value": "{box-shadow.inset.inverted.value}"
```

tokens/properties/components/ToggleButton.json
```diff
@@ -0,0 +1,132 @@
...
+            },
+            "border-color": {
+                "value": "{border.color.progressive.active.value}"
+            },
+            "background-color": {
+                "value": "{background.color.progressive.active.value}"
+            },
+            "focus": {
+                "box-shadow": {
+                    "value": "{box-shadow.inset.inverted.value}"
```
SaiSan-WMDE, line 116:
```suggestion
                    "value": "{box-shadow.inset.focus.inverted.value}"
```

vue-components/src/components/ToggleButtonGroup.vue
```diff
@@ -0,0 +1,84 @@
...
+
+	[dir='ltr'] & :not(:last-child) {
+		border-top-right-radius: 0;
+		border-bottom-right-radius: 0;
+	}
+
+	[dir='rtl'] & :not(:last-child) {
+		border-top-left-radius: 0;
+		border-bottom-left-radius: 0;
+	}
```
itamargiv, line 81: It turns out these are not actually needed anymore, the problem was Storybook itself, as @Ladsgroup  mentioned in #358, which means it is not a problem with logical-css after all, and is unrelated to https://github.com/csstools/postcss-logical/issues/27

vue-components/src/components/ToggleButtonGroup.vue
```diff
@@ -0,0 +1,84 @@
...
+	border-block: none;
+}
+
+.wikit-ToggleButtonGroup :focus {
+	outline: none;
+}
+
+.wikit-ToggleButtonGroup :not(:first-child):disabled {
+	border-inline-start-color: $color-base-100;
+}
```
itamargiv, line 60: Please follow the pattern established in other components and either use a variable for the base class name or nest all these rules under the base class.

itamargiv (changes requested):
Fantastic Job @SaiSan-WMDE, @Ladsgroup and @micgro42. Great work! I made a few small comments about things to change, and then I believe we are good to go.

@SaiSan-WMDE About the outline issues noticed in #358, This is due to the face that the global token's offest value is `inset 0 0 0 1px` when in fact what we want is to create the shdow box outline 1px inside the borders. See this example [on codepen](https://codepen.io/xumium/pen/ZEpgpNo). But we can tackle this later as you suggested.

@Ladsgroup I tried the Change in my machine and removed the extra rules we added as a chrome fallback, and it still works on chrome, turns out it was the `dir="ltr"` in storybook all along, I'll open a task to document this for future reference.

vue-components/src/components/ToggleButtonGroup.vue
Ladsgroup: Oh thanks. Removed now.

vue-components/src/components/ToggleButtonGroup.vue
Ladsgroup: ahaaaa I did `& :not(:last-child)` and got stylelint error `35:3  ✖  Unnecessary nesting       scss/selector-no-redundant-nesting-selector` now I realize what was wrong. Thanks.

itamargiv (approved):
Thank you! Great Job, and the answer is **Artichoke**:

PR #351, opened 2021-01-11
Create Checkbox component

SaiSan-WMDE:
Both the component tokens and code are available in this PR.

Story: [T270166](https://phabricator.wikimedia.org/T270166)


vue-components/src/components/Checkbox.vue
```diff
@@ -0,0 +1,151 @@
...
+
+		&:active {
+			&::before {
+				border-color: $wikit-Checkbox-input-active-border-color;
+				background-color: $wikit-Checkbox-input-active-background-color;
+			}
+		}
+	}
+
+	&__box {/* see: https://www.w3schools.com/howto/howto_css_custom_checkbox.asp */
```
SaiSan-WMDE, line 99: Would it make sense to call this "input" instead of "box", this way it would match the tokens and be more inline with the usual element markup? Same naming would apply to radio buttons. Wdyt?

micgro42: Sounds like a good suggestion to me :+1:

vue-components/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+	// the `/` prefix in the title is needed for "Input" to appear as a folded navigation item, and not a headline
+	title: 'Checkbox',
+};
+
+export function all(): Component {
+	return {
+		components: { Checkbox },
+		template: `
+			<div>
+				<div><Checkbox label="Checked" :value="true" /></div>
```
SaiSan-WMDE, line 15:
```suggestion
				<div><Checkbox label="Checked" aria-checked="true" :value="true" /></div>
```

SaiSan-WMDE: What about keeping up with the display of good practices? Would the aria-checked attribute be added here, in this order?

vue-components/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+				<br>
+				<div><Checkbox label="Unchecked" :value="false" /></div>
```
SaiSan-WMDE, line 17:
```suggestion
				<div><Checkbox label="Unchecked" aria-checked="true" :value="false" /></div>
```

vue-components/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+				<br>
+				<div><Checkbox label="Checked disabled" :value="true" disabled="disabled"/></div>
```
SaiSan-WMDE, line 19:
```suggestion
				<div><Checkbox label="Checked disabled" aria-checked="true" :value="true" disabled="disabled"/></div>
```

vue-components/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,25 @@
...
+				<br>
+				<div><Checkbox label="Unchecked disabled" :value="false" disabled="disabled"/></div>
```
SaiSan-WMDE, line 21:
```suggestion
				<div><Checkbox label="Unchecked disabled" aria-checked="false" :value="false" disabled="disabled"/></div>
```

vue-components/stories/Checkbox.stories.ts
micgro42: No, since we are using a native checkbox in our template, which should still be accessible to a screen-reader.

micgro42: Similarly, we are not using `aria-disabled`. Most of these rules are intended to be used on completely custom build elements (i.e. most of the time `<span>` and `<div>`).

SaiSan-WMDE: thanks for the clarification!

tokens/properties/components/Checkbox.json
```diff
@@ -0,0 +1,114 @@
...
+            "disabled":{
+                "color": {
+                    "value":"{font.color.disabled.value}"
+                }
+            }
+        },
+        "input": {
+            "size": {
+                "value": "{dimension.static.size.300.value}",
+                "comment": "Set as a reminder to rethink dimension alias tokens"
```
SaiSan-WMDE, line 13:
```suggestion
```

tokens/properties/components/Checkbox.json
```diff
@@ -0,0 +1,113 @@
...
+        "label": {
+            "disabled":{
+                "color": {
+                    "value":"{font.color.disabled.value}"
+                }
+            }
+        },
+        "input": {
+            "size": {
+                "value": "{dimension.static.size.300.value}",
```
SaiSan-WMDE, line 12:
```suggestion
                "value": "{dimension.min-height.small.value}",
                "comment": "Not the ideal value, but all these min-height tokens already in use could be renamed as "size" in a coordinated way",
```

SaiSan-WMDE:
```suggestion
                "value": "{dimension.static.size.300.value}"
```

vue-components/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,70 @@
+import Checkbox from '@/components/Checkbox';
+import { Component } from 'vue';
+
+export default {
+	component: Checkbox,
+	// the `/` prefix in the title is needed for "Input" to appear as a folded navigation item, and not a headline
```
micgro42, line 6: We can drop this comment

vue-components/src/components/Checkbox.vue
```diff
@@ -0,0 +1,153 @@
...
+		},
+		disabled: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	data() {
+		return {
+			// https://github.com/vuejs/vue/issues/5886
+			id: generateUid( 'wikit-Checkbox' ),
```
itamargiv, line 46: :+1:

vue-components/src/components/Checkbox.vue
```diff
@@ -0,0 +1,153 @@
...
+
+<style lang="scss">
+$label: '.wikit-checkbox__label';
+
+.wikit-checkbox {
+	&__label {
+		display: flex;
+		align-items: center;
+
+		// TODO: replace with Label mixin
```
itamargiv, line 60: I assume this is what we were discussing in PR #352 ?

itamargiv (approved):
:100: Great work @SaiSan-WMDE and @micgro42 ! Looks fabulous!

itamargiv:
As far as I'm concerned this is ready to merge, apart from the small change I think Michael requested and the open #351 (which could also just be merged as a followup). However, I will wait until Michael's comment is resolved.

vue-components/src/components/Checkbox.vue
SaiSan-WMDE: yes!

vue-components/stories/Checkbox.stories.ts
SaiSan-WMDE, line 6:
```suggestion
```

vue-components/src/components/Checkbox.vue
```diff
@@ -0,0 +1,153 @@
+<template>
+	<span class="wikit wikit-checkbox">
+		<input
+			type="checkbox"
+			:id="id"
+			class="wikit-checkbox__input"
+			@change="$emit( 'input', $event.target.checked )"
+			:value="nativeValue"
+			:checked="value"
```
tzhelyazkova, line 9: I’m very much against switching the native attribute names again in this component (as it happened in Button).
`checked` can just be called `checked` or `isChecked` if you wish and `value` can be `value` instead of `nativeValue`.
Also, I can't quite explain it, but having a boolean variable named `value` seems strange to me.

micgro42: The motivation for connecting the `value` prop with the `checked` attribute was to enable the application to use this component directly with `v-model`.

Though, if we were to drop that concern, then I agree that we should choose the props as you suggest.

itamargiv: @micgro can you exemplify this concern? How is changing the names back going to affect it?

micgro42: With how it is now:

```vue
<template>
	<Checkbox v-model="isChecked" label="loremIpsum">
</template>

<script>
export default Vue.extend( {
	name: 'QueryBuilder',
	data() {
		return {
			isChecked: false,
		};
	},
} );
</script>
```

With the props changed like suggested by Tonina:
```vue
<template>
	<Checkbox :isChecked="isChecked" @input="isChecked = event" label="loremIpsum">
</template>

<script>
export default Vue.extend( {
	name: 'QueryBuilder',
	data() {
		return {
			isChecked: false,
		};
	},
} );
</script>
```

I'm ok with the second, I just want us to make that trade-off consciously.

Also, this might no longer be a problem in vue3? I think v-model changed there to be more flexible, but I'm not sure.

itamargiv: Honestly I am also okay with both, I slightly like the first one better due to brevity, but since @tozh raised this issue I'll wait to see what she thinks.

tzhelyazkova: I think this won't be a problem in vue3, because `v-model` [can accept arguments](https://v3.vuejs.org/guide/migration/v-model.html#v-model-arguments).
On the other hand, I have no idea if there are plans  to migrate the vue components to vue3 anytime soon, so I don't know if that should be taken into consideration yet. 
IMHO the name of the props take precedent over the shorthand syntax when using the Checkbox component.
If others think differently then please go ahead with the current version, I don't want to block your work over something like this. :v:

micgro42: I have to touch this PR again anyway to add the new Label mixin, so I can update the props here as well. I will update the emitted event to be `updated:checked`, so that the final usage will be:

vue2:
```vue
<template>
	<Checkbox :checked.sync="isChecked" label="loremIpsum">
</template>

<script>
export default Vue.extend( {
	name: 'QueryBuilder',
	data() {
		return {
			isChecked: false,
		};
	},
} );
</script>
```

vue3:
```vue
<template>
	<Checkbox v-model:checked="isChecked" label="loremIpsum">
</template>

<script>
export default Vue.extend( {
	name: 'QueryBuilder',
	data() {
		return {
			isChecked: false,
		};
	},
} );
</script>
```

I think that will give us the best of both worlds.

vue-components/stories/Checkbox.stories.ts
```diff
@@ -0,0 +1,75 @@
...
+		components: { Checkbox },
+		props: Object.keys( args ),
+		template: `
+			<div>
+				<div>
+					<Checkbox
+						:label="label"
+						:checked.sync="checked"
+						:disabled="disabled"
+					/>
```
itamargiv, line 23: Looks Great :sunglasses: Thank you!

itamargiv (approved):
Checkbox:

Check

PR #352, opened 2021-01-11
Introduce BlockLabel mixin

micgro42:
The current label mixin is always used in a block context and only in such a context is the included padding sensible. By extracting both the display and padding into a wrapping mixin, we can now
1: only use a single mixin for block labels without further declarations
2: use the Label mixin also for inline labels (which will be useful for the checkbox component)


SaiSan-WMDE:
In my opinion (but of course, take this with a grain of salt), it'd be ideal if, instead of creating a new label mixin, we could  add some arguments to the existing one that allow implementers to change its display and the end padding.  

What do you think, @itamargiv? Deciding this will [unblock the Checkbox component](https://github.com/wmde/wikit/pull/351/files#diff-a0e218e0417524c9eb4d37f833c6ed1b8cd2530bb36ec0ba17c79b46f9726d4aR60).

itamargiv:
I tend to agree with @SaiSan-WMDE. I think we can simply wrap the additional CSS rules with an `@if` operator and pass a boolean to determine if it's block or not, or even pass the value we would like for the display property and accordingly use a conditional to decide on adding `padding-block-end` or not.

micgro42:
I chose to make a second Mixin, because in most programming languages that I'm familiar with, it is considered a code smell to add a flag parameter to a public method and thereby increasing its complexity. Rather the preferred way is usually to have simple methods with descriptive names instead of more general methods with flags and conditionals where possible.

But, since I'm not familiar with the established patterns in Sass, I'll happily adjust the PR as you suggest. Probably tomorrow.

itamargiv:
> it is considered a code smell to add a flag parameter to a public method and thereby increasing its complexity.

Yes, in any other programming language I would agree with you, and also that is why I would prefer that you implement my second suggestion, of having callers of the mixin pass the desired value for the display property over a simple boolean.

Either or, a function with different outcomes depending on passed argument is not completely unheard of and is a pattern I've seen before in sass for example: https://github.com/twbs/bootstrap/blob/main/scss/mixins/_caret.scss#L28-L62

micgro42:
I'm sorry, I still fail to see the benefits that justify the trade-offs and I don't think the example compares to this situation. The example mixin is about adding the caret and letting the caller control the direction via a parameter. This mixin is about changing either just the font or doing that and also   do a second unrelated thing and change the display type and padding.

Nonetheless, I have adjust the mixin as you suggested and I hope the PR can now be merged so that we can use it in the Checkbox PR.

vue-components/src/styles/mixins/Label.scss
```diff
@@ -1,8 +1,12 @@
...
     font-size: $wikit-Label-font-size;
     font-weight: $wikit-Label-font-weight;
     line-height: $wikit-Label-line-height;
-  }
\ No newline at end of file
+
+    @if $displayType != 'inline' {
+        display: $displayType;
+        padding-block-end: $wikit-Label-padding-block-end;
+    }
```
itamargiv, line 11: I think in this case rather than performing a negative check, I would prefer the display property to be fully controlled by the call and only the relevant styles applied on a positive basis rather than on negative (i.e. only on blaock rather than, "anything that is not inline"):

```suggestion
     display: $displayType;
     
    @if $displayType == block {
        padding-block-end: $wikit-Label-padding-block-end;
    }
```

itamargiv (changes requested):
Really one minor change. I get where you are coming from, but I think it is more cohesive to rely on one label concept instead of starting to create multiple ones from each type of display.

vue-components/src/styles/mixins/Label.scss
```diff
@@ -1,8 +1,12 @@
...
-    padding-block-end: $wikit-Label-padding-block-end;
     font-family: $wikit-Label-font-family;
     font-size: $wikit-Label-font-size;
     font-weight: $wikit-Label-font-weight;
     line-height: $wikit-Label-line-height;
-  }
\ No newline at end of file
+     display: $displayType;
+     
+    @if $displayType == block {
```
itamargiv, line 9: Argh, sorry I fudged up the indentation here....

```suggestion
    display: $displayType;
     
    @if $displayType == block {
```

itamargiv (approved):
Looks good, included a minor correction to my suggestion. Sorry about that, otherwise it's good to go.

vue-components/src/styles/mixins/Label.scss
micgro42: It is strange that our style linting is not working here. Also, aren't we supposed to use tabs?^^

SaiSan-WMDE (approved):
Look good! Thanks, Michael 💯

PR #353, opened 2021-01-13
Toggle Proof of Concept

micgro42:
Since those are radio-buttons now, the behavior is a bit different only insofar as that they are now navigated and selected at the same time with the arrow keys.

I'm not yet sure about how I structured the props. They work well if there are multiple inputs, but they are cumbersome if there is only a single one.

Needs more thought.

### DIFF
--- a/vue-components/src/styles/mixins/ButtonGroup.scss
+++ b/vue-components/src/styles/mixins/ButtonGroup.scss
@@ -1,0 +1,17 @@
+@mixin ButtonGroup( $class: '' ) {
+  & > :first-child #{$class} {
+    border-start-end-radius: 0;
+    border-end-end-radius: 0;
+    border-inline-end: none;
+  }
+
+  & > :last-child #{$class} {
+    border-start-start-radius: 0;
+    border-end-start-radius: 0;
+  }
+
+  & > :not(:first-child, :last-child) #{$class} {
+    border-radius: 0;
+    border-inline-end: none;
+  }
+}
--- a/vue-components/stories/TooglePoC.stories.ts
+++ b/vue-components/stories/TooglePoC.stories.ts
@@ -1,6 +1,5 @@
 import TogglePoC from '@/components/TogglePoC';
 import { Component } from 'vue';
-import { MenuItem } from '../src/components/MenuItem';
 
 export default {
 	component: TogglePoC,
@@ -29,7 +28,7 @@ export function ProofOfConcept(): Component {
 		components: { TogglePoC },
 		data(): unknown {
 			return {
-				selectedOption: null,
+				selectedOption: '',
 			};
 		},
 		computed: {
@@ -49,8 +48,8 @@ export function ProofOfConcept(): Component {
 				</div>
 				<br>
 				<TogglePoC
-				  :options="[{label: 'single option', value: 'unclear if needed here'}]"
-				  :value="false"
+				  :options="[{label: 'single option', value: 'this is a bit cumbersome'}]"
+				  value="this is a bit cumbersome"
 				/>
 			</div>
 		`,
--- a/vue-components/stories/TooglePoC.stories.ts
+++ b/vue-components/stories/TooglePoC.stories.ts
@@ -51,6 +51,11 @@ export function ProofOfConcept(): Component {
 					Best topping on Pizza:
 					<span v-if="selectedOption" class="selected-item-label">{{ selectedOption }}</span>
 				</div>
+				<br>
+				<TogglePoC
+				  :options="[{label: 'single option', value: 'unclear if needed here'}]"
+				  :value="false"
+				/>
 			</div>
 		`,
 	};
--- a/vue-components/stories/TooglePoC.stories.ts
+++ b/vue-components/stories/TooglePoC.stories.ts
@@ -10,22 +10,18 @@ const options = [
 	{
 		label: 'Pineapple',
 		value: 'pineapple',
-		checked: true,
 	},
 	{
 		label: 'Mushrooms',
 		value: 'mushrooms',
-		checked: false,
 	},
 	{
 		label: 'Spinach',
 		value: 'spinach',
-		checked: false,
 	},
 	{
 		label: 'Artichoke',
 		value: 'artichoke',
-		checked: false,
 	},
 ];
 export function ProofOfConcept(): Component {
--- a/vue-components/stories/TooglePoC.stories.ts
+++ b/vue-components/stories/TooglePoC.stories.ts
@@ -1,0 +1,57 @@
+import TogglePoC from '@/components/TogglePoC';
+import { Component } from 'vue';
+import { MenuItem } from '../src/components/MenuItem';
+
+export default {
+	component: TogglePoC,
+	title: 'TogglePoC',
+};
+const options = [
+	{
+		label: 'Pineapple',
+		value: 'pineapple',
+		checked: true,
+	},
+	{
+		label: 'Mushrooms',
+		value: 'mushrooms',
+		checked: false,
+	},
+	{
+		label: 'Spinach',
+		value: 'spinach',
+		checked: false,
+	},
+	{
+		label: 'Artichoke',
+		value: 'artichoke',
+		checked: false,
+	},
+];
+export function ProofOfConcept(): Component {
+	return {
+		components: { TogglePoC },
+		data(): unknown {
+			return {
+				selectedOption: null,
+			};
+		},
+		computed: {
+			options() {
+				return options;
+			},
+		},
+		template: `
+			<div>
+				<TogglePoC
+					:options="options"
+					v-model="selectedOption"
+				/>
+				<div style="margin-top: 16px; font-family: sans-serif; color: #202122">
+					Best topping on Pizza:
+					<span v-if="selectedOption" class="selected-item-label">{{ selectedOption }}</span>
+				</div>
+			</div>
+		`,
+	};
+}
--- a/vue-components/src/components/TogglePoC.vue
+++ b/vue-components/src/components/TogglePoC.vue
@@ -1,6 +1,12 @@
 <template>
 	<span class="wikit wikit-TogglePoC">
-		<span class="singleOption" v-if="options.length===1"></span>
+		<span class="singleOption" v-if="options.length===1">
+			<input
+				class="option-input"
+				type="checkbox"
+				:id="id"
+			><label class="option-label" :for="id">{{ options[0].label }}</label>
+		</span>
 		<span class="multipleButtons" v-else>
 			<span class="option" v-for="option in options" :key="option.value">
 				<input
@@ -80,6 +86,7 @@ export default Vue.extend( {
 	}
 
 }
+
 .option-input {
 	width: 0;
 	height: 0;
--- a/vue-components/src/components/TogglePoC.vue
+++ b/vue-components/src/components/TogglePoC.vue
@@ -1,0 +1,111 @@
+<template>
+	<span class="wikit wikit-TogglePoC">
+		<span class="singleOption" v-if="options.length===1"></span>
+		<span class="multipleButtons" v-else>
+			<span class="option" v-for="option in options" :key="option.value">
+				<input
+					class="option-input"
+					@change="change"
+					type="radio"
+					:id="id + option.value"
+					:checked="option.value === value"
+					:name="id"
+					:value="option.value"
+				><label class="option-label" :for="id + option.value">{{ option.label }}</label>
+			</span>
+		</span>
+	</span>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import generateUid from '@/components/util/generateUid';
+
+export default Vue.extend( {
+	name: 'TogglePoC', // Name not yet decided!
+	props: {
+		options: { // option.value needs to be unique!
+			required: true,
+			type: Array,
+		},
+		value: {
+			required: true,
+			type: String,
+		},
+	},
+	data() {
+		return {
+			// https://github.com/vuejs/vue/issues/5886
+			id: generateUid( 'wikit-TextInput' ),
+		};
+	},
+	methods: {
+		change( event: InputEvent ): void {
+			this.$emit( 'input', event.target.value );
+		},
+	},
+} );
+</script>
+
+<style lang="scss">
+
+.option-label {
+	color: $wikit-ToggleButton-color;
+	font-family: $wikit-ToggleButton-font-family;
+	font-weight: $wikit-ToggleButton-font-weight;
+	font-size: $wikit-ToggleButton-font-size;
+	line-height: $wikit-ToggleButton-line-height;
+	border-width: $wikit-ToggleButton-border-width;
+	border-style: $wikit-ToggleButton-border-style;
+	border-radius: $wikit-ToggleButton-border-radius;
+	border-color: $wikit-ToggleButton-border-color;
+	background-color: $wikit-ToggleButton-background-color;
+	padding-inline: $wikit-ToggleButton-medium-padding-horizontal;
+	padding-block: $wikit-ToggleButton-medium-padding-vertical;
+	transition-property: $wikit-ToggleButton-transition-property;
+	transition-duration: $wikit-ToggleButton-transition-duration;
+	transition-timing-function: $wikit-ToggleButton-transition-timing-function;
+
+	//FIXME: add large padding in media query
+	&:hover {
+		color: $wikit-ToggleButton-hover-color;
+		border-color: $wikit-ToggleButton-hover-border-color;
+		background-color: $wikit-ToggleButton-hover-background-color;
+	}
+
+	&:active {
+		color: $wikit-ToggleButton-active-color;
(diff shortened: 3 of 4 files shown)


SaiSan-WMDE:
This is a excellent PoC. Some little things I'm noticing as a heads-up:

- For some reason I couldn't figure out yet (all the dimensions seem correct), the height of the toggle buttons is 30px instead of 32px
- It looks like the `ButtonGroup` mixin is not altering the border radius of the buttons? (as always, checking on Chrome/Mac)
- Regarding the naming: I think  we should stick to ToggleButton (following OOUI & the WMF styleguide). Both terms are used, but I believe Toggle button is a bit more common. In terms of classification, this will group it closer to the future Toggle Swith component in our alphabetical structure.

vue-components/src/styles/mixins/ButtonGroup.scss
```diff
@@ -0,0 +1,17 @@
+@mixin ButtonGroup( $class: '' ) {
+  & > :first-child #{$class} {
```
itamargiv, line 2: I wonder, why are we targeting a class descendant here instead of the first element itself? Same for the other two.

vue-components/stories/TooglePoC.stories.ts
```diff
@@ -0,0 +1,57 @@
...
+	},
+	{
+		label: 'Spinach',
+		value: 'spinach',
+	},
+	{
+		label: 'Artichoke',
+		value: 'artichoke',
+	},
+];
```
itamargiv, line 25: I see where you're coming from regarding the options and not being convinced by the single value case, although I think it's a fair compromise to make for the sake of cohesion and simplicity. 

We can add a ticket to the DS board, if we want to, to enable support for a single object or even a single string in the future.

itamargiv: In any case, we can already say that by default the value could be equal to the label if not explicitly set. At least that's the way I would expect it to work.

itamargiv (commented):
Looking good, added a few more comments.

vue-components/src/styles/mixins/ButtonGroup.scss
micgro42: because the styles are on the `<label>` and not on the `<span>` that wraps the `<label>` and the `<input>`. And the styles must be on the `<label>` so they can depend on the state of the input.

micgro42:
About the naming, I think my favorite suggestion is "_ButtonToggle_", because I think it is more correct and less confusing at the same time.

More correct:
_ToggleButton_ would mean that it is a button to be used for toggling things. But that is false, it is not a button, conceptually it doesn't trigger an action every time one clicks it and it cannot be used in place of other buttons.
_ButtonToggle_ is more correct as it _is_ a Toggle that happens to look like a button. It is used to toggle things and can be used in place of other toggle elements.

Less confusing:
It can visually consist of multiple elements, so calling it _ToggleButton_ and then visually seeing multiple "buttons" coming out of that single element might be strange?
With _ButtonToggle_, I would expect to be able to both toggle a single thing on or off and toggle between multiple things.

That being said, the decision is obviously with @SaiSan-WMDE as the DS Designer and I'll happily implement whatever is decided.

vue-components/stories/TooglePoC.stories.ts
micgro42: Yep, that makes sense

vue-components/src/styles/mixins/ButtonGroup.scss
itamargiv: I think this might create an issue if later, as @SaiSan-WMDE hinted before, we might want to also combine these toggles with other non toggle buttons. In that case we will not be able to control the border radii of the toggle group. Although now that I think of it, combining regular buttons with the toggle buttons is going to be mighty confusing... @SaiSan-WMDE  did I misunderstand you?

SaiSan-WMDE: I don't actually think we discussed that! But yes, it is unlikely that toggle and regular buttons need to be grouped together. And in case a mixed group is needed in the future, we'd come up with a solution then.

SaiSan-WMDE:
> _ToggleButton_ would mean that it is a button to be used for toggling things. But that is false, it is not a button, conceptually it doesn't trigger an action every time one clicks it and it cannot be used in place of other buttons.
  
(*)That is exactly what this component does: it toggles things and triggers actions when clicked. It's just that the actions that it triggers are more "permanent", a change of state that is represented by the activated style of the button (e.g. aligning a text to the left, muting yourself, liking or favoring something).

> It can visually consist of multiple elements, so calling it _ToggleButton_ and then visually seeing multiple "buttons" coming out of that single element might be strange? With _ButtonToggle_, I would expect to be able to both toggle a single thing on or off and toggle between multiple things.

This is why I insisted on making a difference between ToggleButton and ToggleButton group, but I agree on how ButtonToggle can encompass multiple selectable options.

That said: I really think both names are fine, and I wish we didn't need to discuss this, given the recent feedback :-/ To me the order does not impact the concept so much: it's a button that works as a toggle(*); it's a toggle that looks like a button. I'd say that ButtonToggle matches the way it's implemented, so if we are all fine with the approach of the PoC, **let's go for ButtonToggle.**

SaiSan-WMDE:
(*)That is exactly what this component does: it toggles things and triggers actions when clicked. It's just that the actions that it triggers are more "permanent", a change of state that is represented by the activated style of the button (e.g. aligning a text to the left, muting yourself, liking or favoring something).

> It can visually consist of multiple elements, so calling it _ToggleButton_ and then visually seeing multiple "buttons" coming out of that single element might be strange? With _ButtonToggle_, I would expect to be able to both toggle a single thing on or off and toggle between multiple things.

This is why I insisted on making a difference between ToggleButton and ToggleButton group, but I agree on how ButtonToggle can encompass multiple selectable options.

That said: I really think both names are fine, and I feel bad for discussing this, given the recent feedback 😓 
To me, the order does not alter the concept so much: it's a button that works as a toggle(*); it's a toggle that looks like a button. I'd say that ButtonToggle matches the way it's implemented, so if we are all fine with the approach of the PoC, **let's go for ButtonToggle.**

SaiSan-WMDE:
> _ToggleButton_ would mean that it is a button to be used for toggling things. But that is false, it is not a button, conceptually it doesn't trigger an action every time one clicks it and it cannot be used in place of other buttons.

(*)That is exactly what this component does: it toggles things and triggers actions when clicked. It's just that the actions that it triggers are more "permanent", a change of state that is represented by the activated style of the button (e.g. aligning a text to the left, muting yourself, liking or favoring something).

> It can visually consist of multiple elements, so calling it _ToggleButton_ and then visually seeing multiple "buttons" coming out of that single element might be strange? With _ButtonToggle_, I would expect to be able to both toggle a single thing on or off and toggle between multiple things.

This is why I insisted on making a difference between ToggleButton and ToggleButton group, but I agree on how ButtonToggle can encompass multiple selectable options.

That said: I really think both names are fine, and I feel bad for discussing this, given the recent feedback 😓 
To me, the order does not alter the concept so much: it's a button that works as a toggle(*); it's a toggle that looks like a button. I'd say that ButtonToggle matches the way it's implemented, so if we are all fine with the approach of the PoC, **let's go for ButtonToggle.**

SaiSan-WMDE:
> _ToggleButton_ would mean that it is a button to be used for toggling things. But that is false, it is not a button, conceptually it doesn't trigger an action every time one clicks it and it cannot be used in place of other buttons.

(*)That is exactly what this component does (or should do): it's a button that toggles things and triggers actions when clicked. It's just that the actions that it triggers are more "permanent", a change of state that is represented by the activated style of the button (e.g. aligning a text to the left, muting yourself, liking or favoriting something).

> It can visually consist of multiple elements, so calling it _ToggleButton_ and then visually seeing multiple "buttons" coming out of that single element might be strange? With _ButtonToggle_, I would expect to be able to both toggle a single thing on or off and toggle between multiple things.

This is why I insisted a bit on making a difference between ToggleButton and ToggleButton group, but I agree on how ButtonToggle can encompass multiple selectable options.

That being said: I really think both names are fine, and I feel bad for discussing this, given the recent feedback 😓 
To me, the order does not alter the concept so much: it's a button that works as a toggle(*); it's a toggle that looks like a button. Although I agree that Button toggle somehow matches the current implementation suggested by this PoC better, I think we should go for **Toggle button**, which is the generally adopted term for this component or family, with the function preceding the form (e.g. Toggle switch).

SaiSan-WMDE:
> _ToggleButton_ would mean that it is a button to be used for toggling things. But that is false, it is not a button, conceptually it doesn't trigger an action every time one clicks it and it cannot be used in place of other buttons.

(*)But: that is exactly what this component can (or should) do:  although in the QB the result of clicking the toggle button is not immediately apparent (which makes it easy to interpret it as an input), in general this component can indeed trigger immediate actions when clicked. It's just that the actions that it triggers are more "permanent”, a change of state that is represented by the concurrent activated style of the button (e.g. making text bold, muting yourself, liking or favoriting something).

> It can visually consist of multiple elements, so calling it _ToggleButton_ and then visually seeing multiple "buttons" coming out of that single element might be strange? With _ButtonToggle_, I would expect to be able to both toggle a single thing on or off and toggle between multiple things.

This is why I insisted a bit on differentiating between the individual Toggle Button and the combined Toggle Button group. I think that, rather than finding a term that covers both presentations (I agree that Button Toggle does that a bit better, and is more aligned with how the component is implemented in the PoC), we should probably differentiate between them.   

My suggestion: Toggle Button and Composed Toggle Button. (Again, I consider them different stories) 

Why: a group of radio buttons can be exclusive on selection (acting like radio-buttons) or it can allow multiple selection (acting like checkboxes: with no interaction between the individual options; this is, a simple button group made of toggle buttons). The hike needs the former,  and I think that this PoC does a good job creating that one, which we could differentiate from the simple form calling it “Composed toggle button” (to avoid long names such as: Toggle button group with single select).

That being said: I really think both names are fine, and I feel bad for discussing this, given the recent feedback 😓 
To me, the order does not alter the concept so much: it's a button that works as a toggle(*); it's a toggle that looks like a button. But, as mentioned, I think we should go for the more standardised naming: **Toggle button**, with the form semantically preceding its function (e.g. Toggle switch).

micgro42:
> That being said: I really think both names are fine, and I feel bad for discussing this, given the recent feedback :sweat:

I don't, the recent feedback was about the discussions while a component is implemented by the hike.
This here is still the exploratory phase even before hike story-time. So we're preparing the story better so that there are less discussions needed when the actual implementation starts :slightly_smiling_face:  


> I think we should go for the more standardised naming: **Toggle button**, with the form semantically preceding its function (e.g. Toggle switch).

But ...  that is not form preceding function, but function preceding form? In my understanding Toggle is the function and Button is the visual form?



>> ToggleButton would mean that it is a button to be used for toggling things. But that is false, it is not a button, conceptually it doesn't trigger an action every time one clicks it and it cannot be used in place of other buttons.
> 
> (*)But that is exactly what this component can (or should) do: although in the QB the result of clicking the toggle button is not immediately apparent (which makes it easily interpreted as an input), in general this component can indeed trigger immediate actions when clicked. It's just that the actions that it triggers are more "permanent”, a change of state that is represented by the concurrent activated style of the button (e.g. making text bold, muting yourself, liking or favoriting something).

By action I mean that a (conceptual) button does an action repeatedly every time it is clicked. E.g. increasing the zoom-level of a page by one more step. Or it does an action that can only be done once, like submitting a form, which makes the button disappear (e.g. progressing to the next step).
But a (conceptual) toggle does not (primarily) _do_ an action, but it changes state and _represents_ that state and its alternatives.


I think my main point is that a "ToggleButton" would seem to be a sub-type of Button (like QuietButton) that could be swapped in place of other buttons and things would mostly still work. But that is not the case.
However, a ButtonToggle (in its one-out-of-multiple mode) _can_ take the place of a RadioToggle and things will still work.

----

I think we _can_ make a ToggleButton. Which will be a Button (without the `variant` prop, because it basically is `variant="toggle"`) that also has an "isActive" prop. That button would then depend on the application to manage its state. Maybe we could also have a ToggleButtonGroup component that has a slot where the application can add multiple ToggleButton components and it would apply the ButtonGroup mixin and maybe abstract some of the state-management. But that is not this PoC and I think @itamargiv wasn't very fond of that approach.

I think ToggleButton and ButtonToggle are two different concepts and components.
At the end of the day it is your call which one you want :)

vue-components/src/styles/mixins/ButtonGroup.scss
micgro42: Also, note that the $class parameter is optional. If it is not provided, we would be just targeting the first element itself like in the example.

SaiSan-WMDE:
> I think my main point is that a "ToggleButton" would seem to be a sub-type of Button (like QuietButton) that could be swapped in place of other buttons and things would mostly still work. But that is not the case.
However, a ButtonToggle (in its one-out-of-multiple mode) can take the place of a RadioToggle and things will still work.

I see the point, but I really don't think the Toggle button name strictly implies that this has to be a sub-type of button that can replace any other of its similarly-looking peers. The "toggle" part still makes clear the functional distinction, in my opinion. And even the implication is not wrong: again, [WMF classifies it as such](https://design.wikimedia.org/style-guide/components/buttons.html), also (just to name a generally valuable resource) [Material deign](https://material-ui.com/components/toggle-button/)  –  as long as it works as it should and is used accordingly (alone and combined), we shouldn't be blocked by the name, but for sure decide on the implementation (the underlying disagreement here).

> I think we _can_ make a ToggleButton. Which will be a Button (without the `variant` prop, because it basically is `variant="toggle"`) that also has an "isActive" prop. That button would then depend on the application to manage its state. Maybe we could also have a ToggleButtonGroup component that has a slot where the application can add multiple ToggleButton components and it would apply the ButtonGroup mixin and maybe abstract some of the state-management. But that is not this PoC and I think @itamargiv wasn't very fond of that approach.

I'm intrigued to know what Itamar thinks, but to me this would also be a conceptually valid approach. I know styling is not a defining factor, but toggle buttons could also be visually quiet and icon-only (or vice-versa), so we should make sure whether a variant is the way to go (I guess combinations are possible?), or if the particular way in which this button works "elevates" it to the status of component.

itamargiv:
Not to add any oil to the fire, but I fully agree with Michael's point of view, and, naming aside, ToggleButton and ButtonToggle are different concepts. You both know how I feel about splitting this into separate components, and I prefer the ButtonToggle approach. That being said, @SaiSan-WMDE if you think that this is actually worth considering as two separate things (I.e. ToggleButton and ToggleButtonGroup) then I suggest we follow Michael's approach as he outlined in his comment.

itamargiv:
Ha, it seems like we were typing in the same time :D To re-iterate with Michael's words:

> I think my main point is that a "ToggleButton" would seem to be a sub-type of Button (like QuietButton) that could be swapped in place of other buttons and things would mostly still work. But that is not the case.
> However, a ButtonToggle (in its one-out-of-multiple mode) can take the place of a RadioToggle and things will still work.

The fact that material design calls it Toggle Button doesn't mean that they can actually replace regular buttons, there is a reason why even in the material UI standard they are considered separate from the [Button component](https://material-ui.com/components/buttons/) or Even [Button Group](https://material-ui.com/components/button-group/). The only thing we are suggesting to do differently than the Material design standard is to name this concept in a slightly different manner. I also think that the fact that the WMF grouped these two separate concepts together is probably something we need to bring up with them, as I'm not sure if this is the correct classification. (perhaps visually it is, but not behaviorally)

SaiSan-WMDE:
> The fact that material design calls it Toggle Button doesn't mean that they can actually replace regular buttons.

Exactly! That's the whole point: you can call something Toggle button and still consider it separate from the regular button due to their functional differences, and without implying that these are interchangeable. But yet, if you both agree with ButtonToggle being a more accurate way to refer to this component and the way it's implemented (which I can see!) let's go with that.

In any case. What we need the ButtonToggle to do is:

- Works alone as a checkbox, communicating a binary state change  (< base element)
- Can be grouped to provide a set of mutually exclusive options (< SQB need)

Long term (not needed at the moment):

- Can present a quiet and icon-only styles
- Can be grouped in a regular button group (allowing multiple select)

If the ButtonToggle fulfill those needs, fine with me 👍

micgro42:
> In any case. What we need the ButtonToggle to do is:
> 
>     * Works alone as a checkbox, communicating a binary state change  (< base element)
:heavy_check_mark: 


>     * Can be grouped to provide a set of mutually exclusive options (< SQB need)
:heavy_check_mark: 

> Long term (not needed at the moment):
> 
>     * Can present a quiet and icon-only styles
:thinking: I think we can achieve that via additional fields in the `options`-prop objects


>     * Can be grouped in a regular button group (allowing multiple select)
:heavy_check_mark:

itamargiv:
> thinking I think we can achieve that via additional fields in the `options`-prop objects

This is actually an interesting edge case, since currently the relationship between buttons and Icons is that we pass the icon into the button. I wonder how feasible it will be to pass an icon through the options :/

micgro42:
We decided against this approach. Current PoC is in #358

PR #354, opened 2021-01-18
Enforce Icon alignment responsibilities

micgro42:
Components should only handle their inner positioning and alignment, not how they are positioned inside their parent. Consequently, vertical-align: middle is removed from the Icon component.

As a consequence of that the inner alignment of the component that uses the Icon, i.e. ValdationMessage, is adjusted.
Also adjusted is the Button story, as it now closer resembles what needs to be done in the application: positioning the Icon inside the Button as the Button cannot know what it will get inside its slot.

The visual changes of the icon story should be fine, as the icon does now no longer decide how it is positioned in the surrounding element. If we want to position the icon differently in that story, then that, of course, can be done.

This came to light for a Query-Builder story, but the icon-only Buttons on Wikit-Storybook were wrong as well. They have now a height of `32.8px` as they are supposed to have.

Bug: [T268982](https://phabricator.wikimedia.org/T268982)


vue-components/src/components/ValidationMessage.vue
```diff
@@ -50,7 +50,7 @@ export default Vue.extend( {
 	line-height: $wikit-ValidationMessage-line-height;
 	padding-block-start: $wikit-ValidationMessage-padding-block-start;
 	display: inline-flex;
-	align-items: flex-start;
```
SaiSan-WMDE, line 53: I believe this shouldn't be changed. It makes the icon to center-align with the validation message when the latter wraps, while it should preserve its alignment at the top.

Screenshot 2021-01-19 at 11 42 25

SaiSan-WMDE:
I agree with the parent being the one deciding the alignment of the element it contains, it makes total sense 💯  Thanks, Michael!

I'm just leaving some inline comments, and will commit some changes to improve the appearence of the icon stories, yes 👉 👈

vue-components/src/components/ValidationMessage.vue
micgro42: Do you want me to look into this? Also, I think it would make sense to have one example with such a long error message in a story as these are exactly the things that the visual regressions tests are supposed to catch.

vue-components/stories/Button.stories.ts
```diff
@@ -60,30 +60,30 @@ export function iconOnly(): Component {
 				<div>
 					<div style="margin-bottom:16px;">
 						<Button variant="normal" type="neutral" iconOnly aria-label="More Information">
-							<Icon type="info" size="large" color="inherit" />
+							<Icon type="info" size="large" color="inherit" style="vertical-align: top;" />
```
SaiSan-WMDE, line 63: So, in a way, we're removing the alignment properties from the icon component, but then we're applying them to it again inline, in the context of the parent. The responsibility is not fully removed. I wonder, could the alignment be defined within the button component code? Meaning, making sure that whatever is inserted in the button slot(s) has a default center alignment (there are no exceptions to this).

vue-components/src/components/ValidationMessage.vue
SaiSan-WMDE: I think it's a matter of undoing this change, since only the value is changing, right? I'm not sure of the conceptual implications, but it sounds better to me that the alignment of the children is defined in the parent's code (rather than inline, at the story level).

About the copy, you're absolutely right, and I'll update it.

vue-components/stories/Button.stories.ts
SaiSan-WMDE: In case this comment is blocking (I'm aware there's a QB PR depending on this one) we could take this as an early step and explore how to ensure proper alignment of icons within components in a separate task.

micgro42: > Meaning, the button is wired in a way that makes whatever is inserted in its slot(s) has a default center alignment

But, that is not the case. Text is correctly and by default aligned at baseline, as it should. But images need to be aligned top (or bottom, that barely makes a difference).

SaiSan-WMDE: yep, I know it's not the case. I was trying to ask if that alignment should be defined within the component's code (like in the case of the validation message?) instead of in the story.

vue-components/src/components/ValidationMessage.vue
micgro42: Yes, this change and also re-add the padding below.

micgro42: Yes, this change and also re-add the padding below.

vue-components/stories/Button.stories.ts
micgro42: Yes, it might make sense to discuss that in an extra task. It would be speculative CSS in the button, probably trying to match `.wikit-Icon`

SaiSan-WMDE: @itamargiv just a quick check: do you agree to merging this PR as is and creating an extra task? 

Wondering if this could be related to/covered by [T270327](https://phabricator.wikimedia.org/T270327)

SaiSan-WMDE (approved):

PR #358, opened 2021-01-19
ToggleButton and ToggleButtonGroup

micgro42:
Works, but not sure if those props make any sense at all

### DIFF
--- a/vue-components/src/components/ToggleButton.vue
+++ b/vue-components/src/components/ToggleButton.vue
@@ -1,0 +1,80 @@
+<template>
+	<button
+		:class="['wikit', 'wikit-ToggleButton', isActive ? 'wikit-ToggleButton--isActive' : null]"
+		@click="$emit('click', value)"
+	>
+		{{ label }}
+	</button>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend( {
+	name: 'ToggleButton',
+	props: {
+		label: {
+			required: true,
+			type: String,
+		},
+		value: {
+			required: true,
+			type: String,
+		},
+		isActive: {
+			required: true,
+			type: Boolean,
+		},
+	},
+} );
+</script>
+
+<style lang="scss">
+.wikit-ToggleButton {
+	color: $wikit-ToggleButton-color;
+	font-family: $wikit-ToggleButton-font-family;
+	font-weight: $wikit-ToggleButton-font-weight;
+	font-size: $wikit-ToggleButton-font-size;
+	line-height: $wikit-ToggleButton-line-height;
+	border-width: $wikit-ToggleButton-border-width;
+	border-style: $wikit-ToggleButton-border-style;
+	border-radius: $wikit-ToggleButton-border-radius;
+	border-color: $wikit-ToggleButton-border-color;
+	background-color: $wikit-ToggleButton-background-color;
+	padding-inline: $wikit-ToggleButton-medium-padding-horizontal;
+	padding-block: $wikit-ToggleButton-medium-padding-vertical;
+	transition-property: $wikit-ToggleButton-transition-property;
+	transition-duration: $wikit-ToggleButton-transition-duration;
+	transition-timing-function: $wikit-ToggleButton-transition-timing-function;
+
+	&:hover {
+		color: $wikit-ToggleButton-hover-color;
+		border-color: $wikit-ToggleButton-hover-border-color;
+		background-color: $wikit-ToggleButton-hover-background-color;
+	}
+
+	&:active {
+		color: $wikit-ToggleButton-active-color;
+		border-color: $wikit-ToggleButton-active-border-color;
+		background-color: $wikit-ToggleButton-active-background-color;
+	}
+
+	&:focus {
+		color: $wikit-ToggleButton-focus-color;
+		border-color: $wikit-ToggleButton-focus-border-color;
+		background-color: $wikit-ToggleButton-focus-background-color;
+		box-shadow: $wikit-ToggleButton-focus-box-shadow;
+	}
+}
+
+.wikit-ToggleButton.wikit-ToggleButton--isActive {
+	color: $wikit-ToggleButton-selected-color;
+	border-color: $wikit-ToggleButton-selected-border-color;
+	background-color: $wikit-ToggleButton-selected-background-color;
+
+	&:focus {
+		box-shadow: $wikit-ToggleButton-selected-focus-box-shadow;
+	}
+}
+
+</style>
--- a/vue-components/src/components/ToggleButton.vue
+++ b/vue-components/src/components/ToggleButton.vue
@@ -1,7 +1,7 @@
 <template>
 	<button
 		:class="['wikit', 'wikit-ToggleButton', isActive ? 'wikit-ToggleButton--isActive' : null]"
-		@click="$emit('click', value)"
+		@click="onClick"
 	>
 		{{ label }}
 	</button>
@@ -12,7 +12,19 @@ import Vue from 'vue';
 
 export default Vue.extend( {
 	name: 'ToggleButton',
+	methods: {
+		onClick(): void {
+			if ( this.listener ) {
+				this.listener( this.value );
+			}
+			this.$emit( 'click', this.value );
+		},
+	},
 	props: {
+		listener: {
+			type: Function,
+			default: null,
+		},
 		label: {
 			required: true,
 			type: String,
@@ -22,7 +34,7 @@ export default Vue.extend( {
 			type: String,
 		},
 		isActive: {
-			required: true,
+			default: false,
 			type: Boolean,
 		},
 	},
--- a/vue-components/src/components/ToggleButton.vue
+++ b/vue-components/src/components/ToggleButton.vue
@@ -3,6 +3,7 @@
 		:class="['wikit', 'wikit-ToggleButton', buttonIsActive ? 'wikit-ToggleButton--isActive' : null]"
 		@click="onClick"
 	>
+		<!-- @slot required. add the content of the button here (label, icon, etc.) -->
 		<slot />
 	</button>
 </template>
@@ -17,8 +18,12 @@ export default ( Vue as VueConstructor<Vue & ToggleButtonGroupInjection> ).exten
 		onClick(): void {
 			if ( this.toggleListener !== null ) {
 				this.toggleListener( this.value );
+				return;
 			}
-			this.$emit( 'click', this.value );
+			/**
+			 * only emitted when not use as part of a ToggleButtonGroup
+			 */
+			this.$emit( 'click' );
 		},
 	},
 	inject: {
@@ -34,10 +39,17 @@ export default ( Vue as VueConstructor<Vue & ToggleButtonGroupInjection> ).exten
 		},
 	},
 	props: {
+		/**
+		 * required when the ToggleButton is used as part of a ToggleButtonGroup.
+		 */
 		value: {
 			default: null,
 			type: String,
 		},
+		/**
+		 * required when the ToggleButton is used as a standalone component.
+		 * Ignored when used as part of a ToggleButtonGroup
+		 */
 		isActive: {
 			default: false,
 			type: Boolean,
--- a/vue-components/src/components/ToggleButton.vue
+++ b/vue-components/src/components/ToggleButton.vue
@@ -1,9 +1,9 @@
 <template>
 	<button
-		:class="['wikit', 'wikit-ToggleButton', isActive ? 'wikit-ToggleButton--isActive' : null]"
+		:class="['wikit', 'wikit-ToggleButton', buttonIsActive ? 'wikit-ToggleButton--isActive' : null]"
 		@click="onClick"
 	>
-		{{ label }}
+		<slot />
 	</button>
 </template>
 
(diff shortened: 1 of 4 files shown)


Ladsgroup:
So far looks actually pretty okay and we can get it in IMHO, I don't see the point on calling this POC but the only thing is the you can't unselect a button (returning a group to its default state depending default state is selected a button or nothing is selected)

The story book also needs work (disabled state, etc.)

Ladsgroup:
Everything I did is in this commit: https://github.com/wmde/wikit/pull/358/commits/744d6c13629e72ba9ece11e81fdd0ca092cafe20

vue-components/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,122 @@
...
+		box-shadow: $wikit-ToggleButton-focus-box-shadow;
+	}
+
+	&:disabled {
+		color: $wikit-ToggleButton-disabled-color;
+		border-color: $wikit-ToggleButton-disabled-border-color;
+		background-color: $wikit-ToggleButton-disabled-background-color;
+	}
+
+	@media (max-width: $width-breakpoint-mobile) {
```
micgro42, line 105: Looking at storybook, I have the strong suspicion that this media-query is not working

vue-components/src/components/ToggleButtonGroup.vue
```diff
@@ -0,0 +1,41 @@
...
+	props: {
+		value: {
+			required: true,
+			type: String,
+		},
+	},
+} );
+</script>
+<style lang="scss">
+.wikit-ToggleButtonGroup :not(:last-child) {
```
micgro42, line 34: The inner corners are not supposed to have any border radius

vue-components/src/components/ToggleButton.vue
Ladsgroup: I double checked it before making the patch and it worked fine. Maybe you're checking against tablet breakpoint? The mobile breakpoint is pretty small.

vue-components/src/components/ToggleButtonGroup.vue
Ladsgroup: Oh thanks. Done now.

vue-components/src/components/ToggleButton.vue
micgro42: My bad, the media-query is indeed working fine!
It is that `box-sizing: border-box;` is missing. That is why the buttons are so strangely large ^^

The m-sized buttons are supposed to have a total height of 32px, not 44px :)

Also, @SaiSan-WMDE , I think we might be missing some component tokens for the min-height.

Ladsgroup: Added `box-sizing: border-box;`

SaiSan-WMDE: Hi! Thanks for working on this, and for adding `box-sizing: border-box;`. The size of buttons looks good now. We do have min-height tokens, but we haven't been using them with the button component because they proved to be unnecessary. Should that change?

Ladsgroup: Quick question: In `they proved to be unnecessary`, by they you mean the min-height css style or the min-height tokens?
The min-height css styles are in the figma specs and I set them as it's there. If we need to not use them, that's fine.

If that's about the tokens themselves, I'm fine with moving like this.

SaiSan-WMDE: Using the CSS style, and therefore the tokens to define it, seemed to not have a real impact on the component, as far as I remember. I'd say we can move on like this.

I should be rephrasing many of the Figma specs, to avoid them from being interpreted as implementation decisions. Now that UX creates component tokens, these should be considered the design source of truth by developers.

(Leaving more feedback in a separate comment)

Ladsgroup: Done. Removed the min-height css.

SaiSan-WMDE:
This is looking fantastic. The main thing to solve from a UI point of view is still the border radius of the buttons that occupy a central position: for some reason is still not 0px.

Ladsgroup:
> This is looking fantastic. The main thing to solve from a UI point of view is still the border radius of the buttons that occupy a central position: for some reason is still not 0px.

I couldn't find any issues with on main toggle group but I saw the problem in the disabled one, It should be fixed now.

SaiSan-WMDE:
> I couldn't find any issues with on main toggle group but I saw the problem in the disabled one, It should be fixed now.

It appears that the buttons are not displaying the right border radius in none of the states (checking on Chrome/MacOS):

Screenshot 2021-01-25 at 12 27 32

Buttons should look more integrated in the group. Adjacent borders should not display any roundness:

Screenshot 2021-01-25 at 12 34 11

Ladsgroup:
Mine looks different
image

Maybe it's an outdated version?

Ladsgroup:
aha, it looks fine in firefox but not in chrome :(

Ladsgroup:
> aha, it looks fine in firefox but not in chrome :(

In Chrome it says "unknown property type" for  border-start-start-radius: 0; and the rest.

Ladsgroup:
It's not working on Chrome below 89 https://caniuse.com/?search=border-start-start-radius, we can have it now and slowly wait until it rolls out?

SaiSan-WMDE:
> aha, it looks fine in firefox but not in chrome :(

oh nooo. That's usually the other way around.

> It's not working on Chrome below 89 https://caniuse.com/?search=border-start-start-radius, we can have it now and slowly wait until it rolls out?

I actually opened the Chromatic link on Chrome 89 using BrowserStack and the borders still look broken there :-/

SaiSan-WMDE:
Should we use `border-bottom-right-radius` etc?

Ladsgroup:
That would break in RTL languages :( We can go with it for now?

itamargiv:
> It's not working on Chrome below 89 https://caniuse.com/?search=border-start-start-radius, we can have it now and slowly wait until it rolls out?

The installed postcss plugin `postcss-logical` is supposed to take care of transpilation of logical properties into physical ones, and it is supposed to support `border-start-end-radius` and `border-start-start-radius` etc. We should check first if the problem is in our side or the implementation, and then check if there's a bug with the plugin.

Ladsgroup:
> postcss-logical

Does that work in dev mode or it's in the build step?

> postcss-logical

This is very likely this: https://github.com/csstools/postcss-logical/issues/27

Ladsgroup:
> This is very likely this: [csstools/postcss-logical#27](https://github.com/csstools/postcss-logical/issues/27)

Until that issue is resolved. I made a change that basically is the post-build of postcss and it should work just fine now.

vue-components/src/components/ToggleButtonGroup.vue
```diff
@@ -0,0 +1,78 @@
...
+
+.wikit-ToggleButtonGroup :not(:last-child):dir(ltr) {
+	border-top-right-radius: 0;
+	border-bottom-right-radius: 0;
+}
+
+.wikit-ToggleButtonGroup :not(:last-child):dir(rtl) {
+	border-top-left-radius: 0;
+	border-bottom-left-radius: 0;
+}
```
itamargiv, line 69: @ladsgroup There is already code here to try and mitigate the issue you found (not sure if it was added by you if it was already here before because of the force pushing).

I think we should focus on ensuring the approach from line 50 onwards is working rather than repeat those rules as seen in lines 39 - 42. This is because the approach from line 50 onwards is closer to how the resulting code would look like after transpilation.

Also, instead of adding an inline to-do, it is preferable to add a task in the incoming column of the design system board, so that the fix will not be lost or fall out of sync with the code.

itamargiv (changes requested):
Looking great! Thanks for picking this up I've added a request for change as a result of a small discussion above, I will review again after the design is approved.

itamargiv:
> > This is very likely this: [csstools/postcss-logical#27](https://github.com/csstools/postcss-logical/issues/27)
> 
> Until that issue is resolved. I made a change that basically is the post-build of postcss and it should work just fine now.

LOL, didn't see this comment, as I was on the review screen, sorry for repeating what you said.

vue-components/src/components/ToggleButtonGroup.vue
Ladsgroup: I should remove the 39-42. I will make a ticket for it as well.

Ladsgroup: Ticket is made: https://phabricator.wikimedia.org/T272866

SaiSan-WMDE:
Agh, not sure what changed, but the borders were fixed before the last commit, and now they look off again 😮 

Last thing from a design pov. The focus state needs fixing:

- Would it be possible to apply the same logic and remove the border radius of the converging borders on focus? Ideally, they should look/behave like this:

Screenshot 2021-01-25 at 17 52 49

- It also looks like different focus styles are being applied to the buttons on the group? And that the tokens are being overridden somehow?
Screenshot 2021-01-25 at 17 49 44

Ladsgroup:
It's broken again because chrome doesn't :dir() either: https://caniuse.com/?search=dir()
Unbelievable

Ladsgroup:
Also looking at output of postcss-logical, it outputs something like `[dir="ltr"] .theselector` which doesn't work if you don't set dir attribute or set it to "auto" (to get it from the language) which is the case for the storybook. I have a feeling that the storybook will look hideous in Persian.

itamargiv:
> Also looking at output of postcss-logical, it outputs something like `[dir="ltr"] .theselector` which doesn't work if you don't set dir attribute or set it to "auto" (to get it from the language) which is the case for the storybook. I have a feeling that the storybook will look hideous in Persian.

I think a quick way around it is to wrap the story with a div that sets the `dir` attribute.

itamargiv:
>     * It also looks like different focus styles are being applied to the buttons on the group? And that the tokens are being overridden somehow?

BTW, This can be fixed by setting `outline: none` on the `:focus` state of the child of the group and rpeplacingit with an `inset` `box-shadow`.

SaiSan-WMDE:
Ok, current status.

Firefox:

Screenshot 2021-01-26 at 15 42 33

- Border radius are fine 💯 
- Focus state is consistent, but there're a couple of issues: 
1. One of the borders is always thinner than the other, but I understand this might not be easy to solve and therefore would make a compromise there.
2. There's an unwanted dotted border surrounding the buttons on focus

Chrome:

Screenshot 2021-01-26 at 15 43 43

Same issues persist: 

1. Inconsistent focus states (and – same as Firefox – uneven widths)
2. Broken border radius

Ladsgroup:
Are you sure you have the current version? This is the chrome for me now
image

Ladsgroup:
Can't reproduce the dotted border in firefox either :(

vue-components/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,121 @@
...
+	}
+}
+
+.wikit-ToggleButton.wikit-ToggleButton--isActive {
+	color: $wikit-ToggleButton-selected-color;
+	border-color: $wikit-ToggleButton-selected-border-color;
+	background-color: $wikit-ToggleButton-selected-background-color;
+
+	&:focus {
+		box-shadow: $wikit-ToggleButton-selected-focus-box-shadow;
```
SaiSan-WMDE, line 117:
```suggestion
		box-shadow: $wikit-ToggleButton-selected-focus-box-shadow;
		outline: none;
```

SaiSan-WMDE: Checking the current Chromatic link, I can still see the unwanted dotted line in Firefox/Mac around the single toggle buttons. Mind if we apply `outline:none` here?

vue-components/src/components/ToggleButton.vue
```diff
@@ -0,0 +1,121 @@
...
+		color: $wikit-ToggleButton-active-color;
+		border-color: $wikit-ToggleButton-active-border-color;
+		background-color: $wikit-ToggleButton-active-background-color;
+	}
+
+	&:focus {
+		color: $wikit-ToggleButton-focus-color;
+		border-color: $wikit-ToggleButton-focus-border-color;
+		background-color: $wikit-ToggleButton-focus-background-color;
+		box-shadow: $wikit-ToggleButton-focus-box-shadow;
```
SaiSan-WMDE, line 96:
```suggestion
		box-shadow: $wikit-ToggleButton-focus-box-shadow;
		outline: none;
```

vue-components/src/components/ToggleButton.vue
SaiSan-WMDE: and here too?

vue-components/src/components/ToggleButton.vue
SaiSan-WMDE: Checking the current Chromatic link, I can still see the unwanted dotted line in Firefox/Mac around the single toggle buttons. Mind if we apply outline:none here?

Ladsgroup:
Sure. Applied.

SaiSan-WMDE:
I was looking at the seemly fresh Chromatic link. The one that technically was generated after adding `outline:none` and the `dif`property to the div (I checked the Files changed) 🤔 The dotted line is finally gone after adding those extra `outline:none`'s. 

Full disclosure, the focus styles will have to be updated in a separate task that impacts tokens at a higher level ([T272626](https://phabricator.wikimedia.org/T272626)), but that's another story to be tackled by the constellation. 

Here everything looks good! 💯 Thanks sooo much, Amir 🙏

PR #359, opened 2021-01-21
Fix checkbox's width

SaiSan-WMDE:
Applying min-width (and min-height, just in case) to prevent the checkbox input from shrinking.

Issue: [T272609](https://phabricator.wikimedia.org/T272609) (1)

### DIFF
--- a/vue-components/src/components/Checkbox.vue
+++ b/vue-components/src/components/Checkbox.vue
@@ -60,8 +60,8 @@ $label: '.wikit-checkbox__label';
 		align-items: center;
 
 		&::before {
-			width: $wikit-Checkbox-input-size;
-			height: $wikit-Checkbox-input-size;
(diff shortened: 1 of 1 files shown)


vue-components/src/components/Checkbox.vue
```diff
@@ -60,8 +60,8 @@ $label: '.wikit-checkbox__label';
 		align-items: center;
 
 		&::before {
-			width: $wikit-Checkbox-input-size;
```
SaiSan-WMDE, line 63: I removed the `width` and `height` properties because I understand they're being overridden by the min properties, but please correct me if I'm wrong.

Ladsgroup: Tested. It works as expected.

Ladsgroup (approved):

PR #360, opened 2021-01-22
Added scroll functionality when naviagting lookup menu using keyboard

bereket-WMDE:
naviagting lookup menu using keyboard

Bug: T267829

### DIFF
--- a/vue-components/src/components/OptionsMenu.vue
+++ b/vue-components/src/components/OptionsMenu.vue
@@ -77,6 +77,13 @@ export default Vue.extend( {
 			type: Number,
 			default: -1,
 		},
+		/* If set to true, it wiil allow the user to loop through the menu using keyboard.
+		 * i.e, when the user presses the arrow down key on the last menu item, it jumps back to top and vise versa
+		 */
+		allowLooping: {
+			type: Boolean,
+			default: false,
+		},
 	},
 	methods: {
 		onKeyDown( event: KeyboardEvent ): void {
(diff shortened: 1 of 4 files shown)


vue-components/src/components/OptionsMenu.vue
```diff
@@ -77,6 +77,13 @@ export default Vue.extend( {
 			type: Number,
 			default: -1,
 		},
+		/**
+		 * scroll lookup menu items into view while naviagitng using keyboard
+		 */
+		scrollWithKeyboard: {
```
micgro42, line 83: I'm not sure there will ever be a situation where we would not want to scroll with the keyboard. Which circumstances do you have in mind for this?

vue-components/src/components/OptionsMenu.vue
```diff
@@ -106,6 +117,17 @@ export default Vue.extend( {
 					break;
 			}
 		},
+		keyboardScroll(): void {
+			const element = this.$refs[ 'menu-items' ] as HTMLElement[];
+
+			if ( this.scrollWithKeyboard && this.keyboardHoveredItemIndex !== -1 ) {
+				element[ this.keyboardHoveredItemIndex ].scrollIntoView( {
```
micgro42, line 124: And once you have removed the prop, I think you can fix the tests with this approach: https://stackoverflow.com/a/52446309/3293343

vue-components/src/components/OptionsMenu.vue
bereket-WMDE: ah that was something i forgot to remove. i was thinking about the dropdown menu. but we do need it in that case as well

vue-components/src/components/OptionsMenu.vue
bereket-WMDE: wow! thanks a lot for the tip

itamargiv (approved):
Great work, thank you!

PR #362, opened 2021-01-25
Apply minimum calculation for the max width

itamargiv:
This bug fix will cause the OptionsMenu to fall back to a maximum width of 95vw when the viewport width is smaller than 200% of the OptionsMenu parent (arbitrary 5vw reduction chosen to compensate for implementation padding + margin, in consultation with @SaiSan-WMDE).


tzhelyazkova (approved):
damn that css `min()` concept is confusing, had to read it a few times on the MDN docs, but I think I get it now.
+1 for the fix

PR #371, opened 2021-02-03
Add new icon type for search

Silvan-WMDE:
Bug: [T273650](https://phabricator.wikimedia.org/T273650)


itamargiv:
As luck would have it, you were the second to be merged, and naturally, there are some conflicts, mind solving these @Silvan-WMDE ?

PR #373, opened 2021-02-11
feat: add Popover component

micgro42:
This will be implemented across multiple subtasks of T273041.

Bug: [T273041](https://phabricator.wikimedia.org/T273041)


Ladsgroup:
It is small enough to be easily reviewed and have my blessings. If UX looks good, this can go in? (I know this is not mergable atm)

bereket-WMDE (commented):
its a good baseline to continue the rest of the subtasks.
Its looks good. maybe a very very nitpicky suggestion: should we rename `this.isShownData` to `this.isDataShown` ? unless there's a reason I'm missing.

ofc, the failing tests need to be fixed. then can be approved and merged.

micgro42:
> Its looks good. maybe a very very nitpicky suggestion: should we rename `this.isShownData` to `this.isDataShown` ? unless there's a reason I'm missing.

Yeah, or `.isContentShown` would be an even better name

bereket-WMDE:
> > Its looks good. maybe a very very nitpicky suggestion: should we rename `this.isShownData` to `this.isDataShown` ? unless there's a reason I'm missing.
> 
> Yeah, or `.isContentShown` would be an even better name

I like that one

bereket-WMDE (approved):

SaiSan-WMDE:
Lots of progress here! 💯 

Let me suggest the following Storybook changes hoping you can tell me if you agree this would be an improvement or not.  Maybe, instead of the iconOnly buttons, we could use normal buttons with text that provides some extra indications, such as:

- In the _Basic_ page, the button label could say "Popover trigger"
- In the _All_ page, the button labels could indicate the position of the popover ([like in this example](https://bootstrap-vue.org/docs/components/popover)).
It would also be awesome if, in that same page, the popovers were placed in an orderly fashion beyond the fold, so not so much scrolling is necessary.

I could try to do this myself in a different PR, since it doesn't impact the component. What do you think?

SaiSan-WMDE:
Unfortunately, many of the popovers look broken in Safari:

Screenshot 2021-02-26 at 18 40 51

micgro42:
> * In the _Basic_ page, the button label could say "Popover trigger"

Easily done. Will do it tomorrow.

> * In the _All_ page, the button labels could indicate the position of the popover ([like in this example](https://bootstrap-vue.org/docs/components/popover)).

Can do, but note that we would keep showing all the Popovers by default, to allow for visual regression tests to verify their position and appearance.

> It would also be awesome if, in that same page, the popovers were placed in an orderly fashion beyond the fold, so not so much scrolling is necessary.

I agree, but that becomes tricky together with the above requirement of showing all the Popovers right from the start. Though we could probably reduce the distance between two components a bit, that should already help. And maybe change the order of the Popover-positions a bit, that should also help.

> Unfortunately, many of the popovers look broken in Safari:

Yeah, we will have to look into fixing that. My suspicion is that our postcss is not translating things like `inset-block-end: ...` correctly into `bottom: ...`

micgro42:
In the mid-term, we may want to look into separating the storybook for visual regression tests from the storybook for public documentation.

SaiSan-WMDE:
> Can do, but note that we would keep showing all the Popovers by default, to allow for visual regression tests to verify their position and appearance.

Yes, showing the popovers by default is perfect, thanks for covering for that! At the moment, all the popovers disappear when any of the buttons in the _All_ page is clicked, so I thought that using the button labels to indicate direction could be nice for orientation.

> I agree, but that becomes tricky together with the above requirement of showing all the Popovers right from the start. Though we could probably reduce the distance between two components a bit, that should already help. And maybe change the order of the Popover-positions a bit, that should also help.

It looks so much better now! Thanks 🙏

micgro42:
@SaiSan-WMDE I adjusted storybook a bit. What do you think?

Safari is broken because of a problem with our PostCSS setup. I'll made a dedicated ticket for that: [T276028](https://phabricator.wikimedia.org/T276028).
It probably should go into its own pull request as it isn't really related to this PR, it is just the most obvious here. 
But I think the checkbox might be broken in Safari as well? (not sure)

micgro42:
> At the moment, all the popovers disappear when any of the buttons in the All page is clicked, so I thought that using the button labels to indicate direction could be nice for orientation.

Gotcha, will fix!

SaiSan-WMDE:
> It probably should go into its own pull request as it isn't really related to this PR, it is just the most obvious here.
> But I think the checkbox might be broken in Safari as well? (not sure)

You're probably right 😕  I'll do an overall review in Safari to collect all the issues. I could document them in a Phab ticket in case that's useful.

micgro42:
> > It probably should go into its own pull request as it isn't really related to this PR, it is just the most obvious here.
> > But I think the checkbox might be broken in Safari as well? (not sure)
> 
> You're probably right confused I'll do an overall review in Safari to collect all the issues. I could document them in a Phab ticket in case that's useful.

Yes, if you find issues other than this and in the Checkbox, then another Phabricator ticket would be a good idea

micgro42:
Mh, changing those button labels revealed some more positioning problems. Will look into them today.

micgro42:
This should now fix most of the rtl positioning issue, except one, the horizontally centered position. That one needs more thinking.

micgro42:
@SaiSan-WMDE This should now be finally ready to properly review :)

SaiSan-WMDE:
The _All_ page looks quite confusing now: 

- Why is the style of the buttons not matching normal neutral?
- Probably those h2 should have a 100% width, so the buttons go back to be aligned in columns of 4.
- Seeing RTL makes me think that probably we should rename all "right"& "left" props to "end" and "start" respectively. Otherwise, it's really confusing. What do you think?

micgro42:
> The _All_ page looks quite confusing now:
> 
>     * Why is the style of the buttons not matching normal neutral?

Yikes, the color was only for testing purposes to see where the middle of the button is exactly! Will fix that!
The higher height was necessary, to ensure that the buttons are actually aligned to the middle, because previously they were sometimes aligned to the top or the button and it worked just because the small height of the button matched with the small height of the Popover.

>     * Probably those h2 should have a 100% width, so the buttons go back to be aligned in columns of 4.

Not sure what you mean, but can do.

>     * Seeing RTL makes me think that probably we should rename all "right" and "left" props to "end" and "start" respectively. Otherwise, it's really confusing. What do you think?

so top-start, top, top-end, end-top, end-center, end-bottom, bottom-end, bottom-center, bottom-start, start-bottom, start, start-top? :thinking:

micgro42:
So, now it should be better. Let me know what you think, @SaiSan-WMDE :blush:

SaiSan-WMDE:
Thanks for all the changes! 🙏 

> so top-start, top, top-end, end-top, end-center, end-bottom, bottom-end, bottom-center, bottom-start, start-bottom, start, start-top? 🤔

Position will never stop being somewhat confusing, but maybe now the indicator is more aligned with the implementation, and it communicates the actual RTL position properly, unlike before: 

Screenshot 2021-03-02 at 11 59 26

SaiSan-WMDE:
Safari issues will still need to be tackled (in a separate PR, as discussed), and the responsive behavior of popovers should be discussed, but I'd say at this point this is ready to be merged?

micgro42:
> Safari issues will still need to be tackled (in a separate PR, as discussed), and the responsive behavior of popovers should be discussed, but I'd say at this point this is ready to be merged?

Wait, the Safari-Issues should be fixed. Do they still exist?

SaiSan-WMDE:
> Wait, the Safari-Issues should be fixed. Do they still exist?

I'm afraid they do, but only in the _Basic_ page:

Screenshot 2021-03-02 at 14 37 13

micgro42:
> > Wait, the Safari-Issues should be fixed. Do they still exist?
> 
> I'm afraid they do, but only in the _Basic_ page:
> 
> Screenshot 2021-03-02 at 14 37 13

Oh right. I have that fixed in a minute.

micgro42:
> > > Wait, the Safari-Issues should be fixed. Do they still exist?
> > 
> > 
> > I'm afraid they do, but only in the _Basic_ page:
> > Screenshot 2021-03-02 at 14 37 13
> 
> Oh right. I have that fixed in a minute.

Should be fixed now :)

PR #375, opened 2021-02-15
fix: keyboard scroll of OptionsMenu not working on Chrome

micgro42:
Apparently, Chrome is not supporting the options for scrollIntoView.
Therefore we fall back to the basic boolean parameter. This results in a
somewhat more abrupt scroll. That issue can maybe be revisited in future
versions, possibly after support for IE11 has been officially dropped.

Bug: [T272999](https://phabricator.wikimedia.org/T272999)

### DIFF
--- a/vue-components/src/components/OptionsMenu.vue
+++ b/vue-components/src/components/OptionsMenu.vue
@@ -130,10 +130,14 @@ export default Vue.extend( {
 			const element = this.$refs[ 'menu-items' ] as HTMLElement[];
 
 			if ( this.keyboardHoveredItemIndex !== -1 ) {
-				element[ this.keyboardHoveredItemIndex ].scrollIntoView( {
-					behavior: 'smooth',
-					block: 'end',
-					inline: 'nearest',
+				/**
+				 * This setTimeout shouldn't be needed, but it is a workaround to make scrollIntoView with options
+				 * work in Chrome 88, but that problem likely existed on older Chrome versions as well.
+				 *
+				 * TODO: With newer versions of Chrome, try removing the setTimeout and see if it works.
+				 */
+				setTimeout( () => {
+					element[ this.keyboardHoveredItemIndex ].scrollIntoView( { behavior: 'smooth', block: 'nearest' } );
 				} );
 			}
 		},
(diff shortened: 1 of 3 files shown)


SaiSan-WMDE:
Yeah, that's definitely abrupt 😞 Would there be a way (simple) to alleviate the "flickering" that occurs when moving the focus to and from the hidden option? In any case, it's good to see that the options can be reached on keyboard scroll. Thanks, Michael!

micgro42:
There might be a workaround with `setTimeout()` that makes it work even though it shouldn't. Let me have another look tomorrow.

Ladsgroup (approved):

micgro42:
@SaiSan-WMDE and @Ladsgroup: Can you have a look again? It should now work nicely in Chrome :)

vue-components/src/components/OptionsMenu.vue
```diff
@@ -130,10 +130,14 @@ export default Vue.extend( {
...
-					block: 'end',
-					inline: 'nearest',
+				/**
+				 * This setTimeout shouldn't be needed, but it is a workaround to make scrollIntoView with options
+				 * work in Chrome 88, but that problem likely existed on older Chrome versions as well.
+				 *
+				 * TODO: With newer versions of Chrome, try removing the setTimeout and see if it works.
+				 */
+				setTimeout( () => {
+					element[ this.keyboardHoveredItemIndex ].scrollIntoView( { behavior: 'smooth', block: 'nearest' } );
```
Ladsgroup, line 140: Shouldn't it have a timeout number? like 300ms?

micgro42: No, without a number (or with `0`) it is just scheduled run after the current thread has finished. We don't actually want a delay there, so this should be fine.

Ladsgroup: aha I remember now. it's mentioned in https://www.youtube.com/watch?v=8aGhZQkoFbQ

Ladsgroup (approved):

SaiSan-WMDE:
Hey! That behaviour really improved! Thank you 💯

PR #376, opened 2021-02-15
fix: only scroll menu if option is not yet visible

micgro42:
When navigating the OptionsMenu with the keyboard, then it should only
scroll to a new option, if that option is not yet fully visible.


micgro42:
implemented in a different way. This PR is longer needed.

PR #377, opened 2021-02-19
Introduce ExtendedNumberInput component

Ladsgroup:
Bug: T275114
Bug: T275113

PR #378, opened 2021-02-19
feat: add QuantityInput component

micgro42:
wip

#377 needs to be merged first because this PR here needs it validation functionality.


SaiSan-WMDE:
This looks great. I give it 1000 cabbage 🥬 

I made some small changes to the style of the entered value/quantity indicators (as always), and replaced the error copies just to see if the ones indicated in the designs are clear enough. While doing that, I noticed the nit-pickiest thing: that the error message of the second QuantityInput (when the unit field is being validated) is not properly aligned like the rest. I wasn't sure what's causing it or how to fix it:

Screenshot 2021-03-04 at 10 51 40

Regarding the transparent end border applied to the first field: this make the component look closer to the designs, but the solution is a bit hacky. In case there's not a better way to achieve creating that shared 1px border, it might be better to make the border visible instead.

vue-components/src/components/QuantityInput.vue
```diff
@@ -0,0 +1,221 @@
...
+
+		.wikit-Input {
+			border-start-start-radius: 0;
+			border-end-start-radius: 0;
+		}
+
+		.wikit-Lookup__label {
+			position: absolute;
+			width: 1px;
+			height: 1px;
```
bereket-WMDE, line 202: there are no tokens here ?

SaiSan-WMDE: There could be? But I don't get to understand what are they styling 🤔

vue-components/tests/unit/components/QuantityInput.spec.ts
```diff
@@ -0,0 +1,330 @@
...
+						unitLookupLabel: '',
+						unitLookupPlaceholder: '',
+						unitLookupMenuItems: [],
+						unitLookupSearchInput: '',
+					},
+				} );
+				const enteredNumberText = '123';
+
+				wrapper.findComponent( Input ).vm.$emit( 'input', enteredNumberText );
+				await wrapper.vm.$nextTick();
```
bereket-WMDE, line 323: maybe also here ?
see [comment here](https://github.com/wmde/wikit/pull/382/files#r588422585)

bereket-WMDE:
P.S. 
Its looks great, btw :)

vue-components/src/components/QuantityInput.vue
micgro42: Ah, this was hiding the Label from the Lookup component so that it is still readable to screen readers. But we're no longer using the Lookup, but the LookupInput, so this is now unused code and can be removed.

vue-components/tests/unit/components/QuantityInput.spec.ts
micgro42: Good catch, done

micgro42:
> While doing that, I noticed the nit-pickiest thing: that the error message of the second QuantityInput (when the unit field is being validated) is not properly aligned like the rest. I wasn't sure what's causing it or how to fix it:
> 
> Screenshot 2021-03-04 at 10 51 40


Mh, that looks perfectly fine to me: 
image

What browser are you using and does this also happen on other validation messages?

micgro42:
> Regarding the transparent end border applied to the first field: this make the component look closer to the designs, but the solution is a bit hacky. In case there's not a better way to achieve creating that shared 1px border, it might be better to make the border visible instead.

I have to confess, I'm only aware of different bad solutions.:
* The transparent border allows us to actually show 4 sides of red border in case of an error. But that means in the case of an error there is also a double border there.
* What we did for the ToggleButton group is to just drop the right border for all but the rightmost button. Luckily that is almost imperceptible unless one knows exactly where one has to look when.
* double border always is always consistent but also always a double border.

vue-components/src/components/QuantityInput.vue
bereket-WMDE: great

bereket-WMDE (dismissed):

SaiSan-WMDE:
> Mh, that looks perfectly fine to me:
> image
> 
> What browser are you using and does this also happen on other validation messages?

Chrome on MacOS. This issue is tinny, though. Finding whatever might be causing it is the interesting part, but it shouldn't be a blocker from a UI perspective.

Regarding the transparent border, I understand how this was the only approach possible to translate the design..but looking at the result, I'd say that we should use the double border for now (although it's not ideal) to avoid both the inconsistencies between states and the hacky-looking corners:

Screenshot 2021-03-04 at 10 35 44

PR #379, opened 2021-02-22
refactor: extract core of LookupComponent

micgro42:
We'll need that core in the QuantityInput component.

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -3,7 +3,7 @@
 		:class="[ 'wikit', 'wikit-Lookup' ]"
 	>
 		<label class="wikit-Lookup__label" :for="inputId">{{ label }}</label>
-		<LookupCore
+		<LookupInput
 			:id="inputId"
 			:feedback-type="feedbackType"
 			:menu-items="menuItems"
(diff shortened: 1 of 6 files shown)


Ladsgroup:
LookupCore sounds vague. Can it be LookupInput instead?

micgro42:
> LookupCore sounds vague. Can it be LookupInput instead?

The name is certainly open for improvement. But it is not only the Input, but also the OptionsMenu. My first working title was `InputAndMenu`, not great either.

More suggestions/ideas welcome.

micgro42:
@wmde/wmde-wikidata-ds This is now ready for review. Happy to see better ideas for the name

Ladsgroup (dismissed):

jakobw:
I agree with Amir that `LookupInput` gives a better idea of what it is, even if it is technically an input + a menu. Any concise name ends up being synonymous with Lookup in some way which would also be confusing.

micgro42:
I still find `LookupCore` more descriptive and accurate than `LookupInput`, but I'll yield to the larger number of voices saying otherwise and have renamed it accordingly.

bereket-WMDE (dismissed):
The PR looks good so far.
as far as the name goes, I think `Lookupcore` is the better choice. but its too late now, since its has been renamed.
I'm sure someone in the future, who's consuming Wikit, will raise this issue up :)

Ladsgroup (approved):

PR #380, opened 2021-02-22
position_the_popover_based_on_a_prop

bereket-WMDE:
Bug: T274259

TODO:

- [x] left-bottom, left-center, left-top classes
- [x] unit tests
- [x] a better way for class extensions

@micgro42 whenever you've time, can you look over this PR? I would appreciate some feedbacks on this.

### DIFF
--- a/vue-components/src/components/Popover.vue
+++ b/vue-components/src/components/Popover.vue
@@ -1,6 +1,7 @@
 <template>
 	<div
-		class="wikit wikit-Popover wikit-Popover--bottom"
+		class="wikit wikit-Popover"
+		:class="[ `wikit-Popover`, `wikit-Popover--${position}` ]"
 		@mouseenter="startHover"
 		@mouseleave="endHover"
 		v-detect-click-outside="clickOutsideHandler"
@@ -22,6 +23,7 @@
 <script lang="ts">
 import Vue from 'vue';
 import detectClickOutside from '@/directives/detectClickOutside';
+import { PopoverPositions } from './PopoverProps';
 
 const HOVER_SHOW_HIDE_DELAY_IN_MS = 100;
 
@@ -51,6 +53,16 @@ export default Vue.extend( {
 		reactToHover: {
 			type: Boolean,
 			default: true,
+		},
+		/**
+		 * Use this prop to set position of the popover in relation to the element that triggers it
+		 */
+		position: {
+			type: String,
+			validator( value: string ): boolean {
+				return Object.values( PopoverPositions ).includes( value as PopoverPositions );
+			},
+			default: 'top',
 		},
 	},
 	methods: {
@@ -154,6 +166,7 @@ $base: '.wikit-Popover';
 		width: math.hypot($wikit-Popover-pointer-width/2, $wikit-Popover-pointer-height);
 		height: math.hypot($wikit-Popover-pointer-width/2, $wikit-Popover-pointer-height);
 		transform: rotate(45deg);
+		transform-origin: bottom left;
 		border: $wikit-Popover-border-width $wikit-Popover-border-style $wikit-Popover-border-color;
 		background: $wikit-Popover-background-color;
 		box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


vue-components/src/components/Popover.vue
```diff
@@ -9,7 +10,7 @@
 			<!-- @slot Target should always be a button, as we will listen to its click and hover events -->
 			<slot name="target" />
 		</span>
-		<div class="wikit-Popover__content-wrapper" v-if="isContentShown">
+		<div class="wikit-Popover__content-wrapper" v-if="true || isContentShown">
```
bereket-WMDE, line 13: Ignore this line. I was just testing some stuff :)

vue-components/src/components/Popover.vue
```diff
@@ -159,7 +180,7 @@ $base: '.wikit-Popover';
 		box-sizing: border-box;
 	}
 
-	&--bottom {
+	&--main-wrapper {
```
micgro42, line 183: I'm not sure what the idea behind this is.
Shouldn't all those values be different for at least one the directions?
If there are some declaration that are the same for all, then why not add them to their respective classes above?

vue-components/src/components/Popover.vue
```diff
@@ -179,5 +200,127 @@ $base: '.wikit-Popover';
...
+	&--bottom-left {
+		#{$base}__content-wrapper {
+			transform: translateX(-100%);
+		}
+		#{$base}__pointer {
+			inset-inline-start: calc(100% - (#{$wikit-Popover-pointer-width} + #{$wikit-Popover-pointer-width} / 2));
+		}
+	}
+
+	&--top-center {
```
micgro42, line 226: I think this can just be `--top`, (and `--right`, etc.).

vue-components/src/components/Popover.vue
```diff
@@ -179,5 +200,127 @@ $base: '.wikit-Popover';
...
+		}
+	}
+	&--right-bottom{
+		#{$base}__content-wrapper {
+			transform: translate(0, -50%);
+			margin-inline-start: $wikit-Popover-distance + $wikit-Popover-pointer-height;
+			inset-inline-start: 100%;
+		}
+		#{$base}__pointer {
+			width: $wikit-Popover-pointer-width / 2;
```
micgro42, line 297: No. This requires you knowing what the actual ratio between width and height is. Rather, when it is rotated by 90 degrees, then the width becomes the height and vice versa, and we should use the values accordingly.

micgro42 (commented):
Currently, in storybook the pointer is not below the center of the target.

As it is shown in Figma:
image


Also, you can make your life a bit easier in storybook if you extend the basic story with the position prop.

So the template would then look like
```
<Popover :is-shown="isShown" :react-to-hover="reactToHover" :position="position">
```
and the `argTypes` could be extended to

```
basic.argTypes = {
	position: {
		control: {
			type: 'inline-radio',
			options: [ 'bottom-left', 'bottom', 'right-bottom', 'right' ], // add all eight options here
		},
	},
	content: {
		control: {
			type: 'text',
		},
	},
};
```

Also, the `all` story will have to contain a visible (but small) popover for each direction

vue-components/src/components/Popover.vue
micgro42: (I know that it is stated differently in the task, but after looking at https://bootstrap-vue.org/docs/components/popover I agree that having just `top`, `right`, `bottom`, `left` for the center positions is good enough)

bereket-WMDE:
@micgro42 thanks for the feedback.
I updated your change requests. However, I'm having some issues with the story template.
but is not a big deal. we can discuss it tomorrow.

bereket-WMDE:
Updated the `all` template

vue-components/stories/Popover.stories.ts
```diff
@@ -48,26 +58,27 @@ basic.argTypes = {
...
+						<Popover :is-shown="true" :react-to-hover="false" :position="value">
 							<template v-slot:target>
 								<Button :iconOnly="true" aria-label="show hint">
 									<Icon type="info-outlined" size="medium" style="vertical-align: top;"/>
 								</Button>
 							</template>
 							<template v-slot:default>
 								<div>
-									Foo bar
+								current position <b><em>{{ value }}</em></b>. You can modify it in the "Controls" section.
```
micgro42, line 80: It probably shouldn't claim twelve times that I can modify the position in the controls, when that is actually false. :)

vue-components/src/components/Popover.vue
```diff
@@ -179,5 +194,239 @@ $base: '.wikit-Popover';
...
+		#{$base}__pointer {
+			width: $wikit-Popover-pointer-width;
+			height: $wikit-Popover-pointer-height;
+			inset-block-start: -$wikit-Popover-pointer-height;
+			inset-inline-start: $wikit-Popover-pointer-width;
+		}
+
+		#{$base}__pointer::before {
+			inset-block-end: 0;
+			transform-origin: bottom left;
```
micgro42, line 215: I think this `transform-origin: bottom left;` is identically for every position, right? Then it can move out of the positioning rules, into the `&__pointer::before` rule above, next to the transform rule that is already there.

vue-components/src/components/Popover.vue
```diff
@@ -52,6 +53,20 @@ export default Vue.extend( {
...
 			default: true,
 		},
+		/**
+		 * Use this prop to set position of the popover in relation to the element that triggers it
+		 */
+		position: {
+			type: String,
+			validator( value: string ): boolean {
+				return value === null ||
+					[ 'top', 'right', 'bottom', 'left',
```
micgro42, line 63: those are repeated three times, once here and twice in the story. Let's extract them into an enum. Have a look at the `iconProps.ts` file for an example

micgro42 (changes requested):
Looks good and it is almost done! There three small things that we should improve and then it can move forward.

vue-components/stories/Popover.stories.ts
bereket-WMDE: ha, correct :)

vue-components/src/components/Popover.vue
bereket-WMDE: thanks. it was a silly oversight

micgro42:
Not sure why exactly, but that last change broke the off-center positions

vue-components/stories/Popover.stories.ts
```diff
@@ -34,10 +35,17 @@ export function basic( args ): Component {
...
 	reactToHover: true,
+	position: 'bottom',
 	content: 'Here is some <em>content</em>. You can modify it in the "Controls" section.',
 };
 
 basic.argTypes = {
+	position: {
+		control: {
+			type: 'inline-radio',
+			options: PopoverPositions,
```
micgro42, line 46: Nitpick: I think if you make this `Object.values(PopoverPositions)` then the options might not be in all-caps in the controls section

micgro42 (changes requested):
Better, but the positions left-top, left-bottom and right-top still need work. That becomes especially apparent when you add a bit more text to the Popover.

SaiSan-WMDE:
I'm noticing the vertical position issues too, plus the pointer looks off when the popover is placed on the left (Chrome/Mac):

Screenshot 2021-02-25 at 19 40 53

Kinda related: Would it be possible to make available the `position` values in a dropdown in the Controls, instead of multiple radio buttons?

bereket-WMDE:
> I'm noticing the vertical position issues too, plus the pointer looks off when the popover is placed on the left (Chrome/Mac):
> 
> Screenshot 2021-02-25 at 19 40 53
> 
> Kinda related: Would it be possible to make available the `position` values in a dropdown in the Controls, instead of providing multiple radio buttons?


Thanks for the feedback. the dropdown is now added in the controls. we also worked on fixing the position of the pointer

vue-components/src/components/Popover.vue
```diff
@@ -154,30 +166,202 @@ $base: '.wikit-Popover';
...
+		}
+
+		#{$base}__pointer::before {
+			inset-block-start: calc(-1 * (50% + #{$wikit-Popover-pointer-width} / 2));
+		}
+	}
+
+	@mixin right {
+		#{$base}__content-wrapper {
+			margin-inline-start: $wikit-Popover-distance + ($wikit-Popover-pointer-height);
```
micgro42, line 209: Those parentheses are not needed

vue-components/src/components/Popover.vue
```diff
@@ -154,30 +166,202 @@ $base: '.wikit-Popover';
...
+				inset-block-start: 50%;
+			}
+
+			@if $subposition == top {
+				inset-block-end: 0;
+				margin-block-end: $wikit-Popover-pointer-margin-vertical;
+			}
+
+			@if $subposition == bottom {
+				inset-block-start: $wikit-Popover-pointer-height;
```
micgro42, line 302: Needs to be `$wikit-Popover-pointer-margin-vertical`

vue-components/src/components/Popover.vue
```diff
@@ -154,30 +166,202 @@ $base: '.wikit-Popover';
...
+		#{$base}__pointer {
+			width: $wikit-Popover-pointer-width;
+			height: $wikit-Popover-pointer-height;
+
+			@if $subposition == center {
+				inset-inline-start: calc(50% - #{$wikit-Popover-pointer-width} / 2);
+			}
+
+			@if $subposition == right {
+				inset-inline-start: $wikit-Popover-pointer-width / 2;
```
micgro42, line 264: Needs to be `$wikit-Popover-pointer-margin-horizontal`

vue-components/src/components/Popover.vue
```diff
@@ -154,30 +166,202 @@ $base: '.wikit-Popover';
...
+			}
+
+			@if $subposition == left {
+				inset-inline-end: $wikit-Popover-pointer-width / 2;
```
micgro42, line 268: Needs to be `$wikit-Popover-pointer-margin-horizontal`

micgro42 (changes requested):
Some places are still not using the correct token, otherwise it looks like things are coming together :+1:

bereket-WMDE:
> Some places are still not using the correct token, otherwise it looks like things are coming together +1

Thanks for pointing them out.

micgro42 (approved):
Looks good to me. I think this branch can now be merged into the PopoverComponent branch.

bereket-WMDE:
@micgro42 Thanks for reviewing it.

PR #381, opened 2021-02-25
Link component

Ladsgroup:
Issue: [T272958](https://phabricator.wikimedia.org/T272958)

### DIFF
--- a/vue-components/src/components/Link.vue
+++ b/vue-components/src/components/Link.vue
@@ -1,10 +1,12 @@
 <template>
-	<span
+	<a
 		:class="[
 			'wikit',
 			'wikit-Link',
 			(underlined == true) ? `wikit-Link--underlined` : ''
 		]"
+		:href="href"
+		:target="target"
 	>
 		<Icon
 			v-if="icon === 'before'"
@@ -21,7 +23,7 @@
 			color="progressive"
 			size="xsmall"
 		/>
-	</span>
+	</a>
 </template>
 
 <script lang="ts">
@@ -44,6 +46,14 @@ export default Vue.extend( {
 		underlined: {
 			type: Boolean,
 			default: false,
+		},
+		href: {
+			type: String,
+			default: '',
+		},
+		target: {
+			type: String,
+			default: '_self',
 		},
 	},
 	components: {
--- a/vue-components/src/components/Link.vue
+++ b/vue-components/src/components/Link.vue
@@ -1,0 +1,79 @@
+<template>
+	<span
+		:class="[
+			'wikit',
+			'wikit-Link',
+			(underlined == true) ? `wikit-Link--underlined` : ''
+		]"
+	>
+		<Icon
+			v-if="icon === 'before'"
+			class="wikit-Link__icon wikit-Link__icon--before"
+			type="link"
+			color="progressive"
+			size="large"
+		/>
+		<span class="wikit-Link__content"><slot /></span>
+		<Icon
+			v-if="icon === 'after'"
+			class="wikit-Link__icon wikit-Link__icon--after"
+			type="newwindow"
+			color="progressive"
+			size="large"
+		/>
+	</span>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import Icon from './Icon.vue';
+
+/**
+ * Uses the following components internally: Icon
+ */
+export default Vue.extend( {
+	name: 'Link',
+	props: {
+		icon: {
+			type: String,
+			default: 'none',
+			validator( value: string ): boolean {
+				return [ 'before', 'after', 'none' ].includes( value );
+			},
+		},
+		underlined: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	components: {
+		Icon,
+	},
+} );
+</script>
+
+<style lang="scss">
+.wikit-Link {
+	@include Link();
+
+	display: flex;
+	align-items: center;
+	width: max-content;
+
+	&--underlined {
+		text-decoration: underline;
+	}
+
+	&__icon {
+		color: $wikit-Link-icon-color;
+
+		&--before {
+			padding-inline-end: $wikit-Link-icon-spacing;
+		}
+
+		&--after {
+			padding-inline-start: $wikit-Link-icon-spacing;
+		}
+	}
+}
+</style>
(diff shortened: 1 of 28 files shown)


vue-components/src/components/Link.vue
```diff
@@ -0,0 +1,79 @@
...
+		/>
+		<span class="wikit-Link__content"><slot /></span>
+		<Icon
+			v-if="icon === 'after'"
+			class="wikit-Link__icon wikit-Link__icon--after"
+			type="newwindow"
+			color="progressive"
+			size="large"
+		/>
+	</span>
```
bereket-WMDE, line 24: maybe the Icon element could just be one. 
something like this, maybe
`<Icon
        class="wikit-Link__icon"
        :class="[ `wikit-Popover--${icon}` ]"
        v-if="icon !== 'none'"
        type="setLink()"
        color="progressive"
         size="large"
/>`
and you can set the type based on the `Icon` type...

its just a thought. doesn't have to be implemented.

bereket-WMDE:
no tests ?

vue-components/src/components/Link.vue
Ladsgroup: Thanks but I don't understand how that place the icon before and after the slot. Am I missing something obvious?

Ladsgroup:
> no tests ?

It doesn't have any behavior to test. The look gets tested through chromatic. I don't see the point of testing for example hover though.

vue-components/stories/Link.stories.ts
```diff
@@ -0,0 +1,58 @@
...
+			<div>
+			<p><Link :underlined="underlined"><span v-html="content" /></Link></p>
+			</div>
+		`,
+    };
+}
+
+basic.args = {
+    underlined: false,
+    content: 'Click here. You can modify it in the "Controls" section.',
```
SaiSan-WMDE, line 27:
```suggestion
    content: 'Click here',
```

Ladsgroup: Done.

SaiSan-WMDE: What about this?

Ladsgroup: Sure. Now it doesn't let me commit it.

Ladsgroup: Done.

vue-components/src/components/Link.vue
```diff
@@ -0,0 +1,79 @@
...
+			color="progressive"
+			size="large"
+		/>
+		<span class="wikit-Link__content"><slot /></span>
+		<Icon
+			v-if="icon === 'after'"
+			class="wikit-Link__icon wikit-Link__icon--after"
+			type="newwindow"
+			color="progressive"
+			size="large"
```
SaiSan-WMDE, line 22:
```suggestion
			size="xsmall"
```

Ladsgroup: Figma says it should be large. Am I missing something really obvious? https://www.figma.com/file/gi3JYwSUQYhKS0mHwK25cQwi/WiKit?node-id=3922%3A5

SaiSan-WMDE: Mind if we use the xsmall size in this story? 

I've been hesitating to pass the UI tests only because of this sizing dilemma, but I don't think it should be a blocker at this point: should we apply a different, smaller size to icons placed after text elements just because of aesthetic reasons? We have at least two cases: labels and links, but it keeps feeling arbitrary.

SaiSan-WMDE: I only documented the --before icon in Figma :-/

vue-components/stories/Link.stories.ts
```diff
@@ -0,0 +1,58 @@
...
+			<div>
+			<p><Link :underlined="underlined"><span v-html="content" /></Link></p>
+			</div>
+		`,
+    };
+}
+
+basic.args = {
+    underlined: false,
+    content: 'Controllable link.',
```
SaiSan-WMDE, line 27:
```suggestion
    content: 'Controllable link',
```

vue-components/src/components/Link.vue
SaiSan-WMDE: Hey, @Ladsgroup. After further discussion with Charlie, we agreed on the fact that (like with Labels) the suffix icon shouldn't be so big, as it gets distracting while its purpose is providing secondary information rather than making the link recognizable/communicating meaning. So I committed some changes to:

- Introduce the sizes 'small' and 'xsmall' of the Icon component
- Separate the size tokens of the prefix and suffix in Link.json
- Apply a 'xsmall' size to the --after icon

Hope that's fine!

vue-components/src/components/Link.vue
bereket-WMDE: you're right. I missed the icons

SaiSan-WMDE:
I thought it would be nice to display the visited state on click (and remove it on page reload). But I'm not sure where would I add an href attribute?

micgro42:
I'm confused. This seems to be missing the actual link, i.e., the `<a>` tag?

Ladsgroup:
That's a valid concern, I went with span since I want to let implementation handle it. For examples I have seen cases that look like a link but they are not <a> for various reasons (javascript handling the click for example) I don't have a preference though.

micgro42:
> That's a valid concern, I went with span since I want to let implementation handle it. For examples I have seen cases that look like a link but they are not for various reasons (javascript handling the click for example) I don't have a preference though.

True, but that is in almost all cases the wrong way to do it. We have buttons for this. Let's go with a native link here.

Ladsgroup:
Cool. Done now.

SaiSan-WMDE:
As mentioned during the daily this morning, it would be great if links could go back to their default state instead of remaining visited (e.g. on reload?).

Ladsgroup:
> As mentioned during the daily this morning, it would be great if links could go back to their default state instead of remaining visited (e.g. on reload?).

Done.

SaiSan-WMDE:
Thanks, @Ladsgroup! Hopefully one last thing: the "underlined" prop seems to not be working (it's applied by default, and the component doesn't react to a change in the Controls setting). It might have something to do with using the <a> element 🤔

Ladsgroup:
> Thanks, @Ladsgroup! Hopefully one last thing: the "underlined" prop seems to not be working (it's applied by default, and the component doesn't react to a change in the Controls setting). It might have something to do with using the `<a>` element thinking

Done. It broke because of changing it to `<a>` which has underline. It should be fixed now.

SaiSan-WMDE (approved):
Looking good!

PR #382, opened 2021-02-25
feat: Return parsed value from ExtendedNumberInput

micgro42:
This component needs to provide and use the parsed input string.

Added test examples from the reference implementation at
https://github.com/wmde/Number/blob/master/tests/ValueParsers/QuantityParserTest.php


micgro42:
@SaiSan-WMDE This PR slightly changes the controllable story by explicitly showing the values that the internal validation extracted from the text that the user entered. Could you have a look on whether you're fine with it?

There seem to be also several unrelated changes there, I'm not sure what caused them.

SaiSan-WMDE:
> @SaiSan-WMDE This PR slightly changes the controllable story by explicitly showing the values that the internal validation extracted from the text that the user entered. Could you have a look on whether you're fine with it?

I think it's quite a nice thing to have. Thanks, Michael. I'll quickly adjust the style of these value indicators to make them match the other stories, if that's ok.

vue-components/tests/unit/components/ExtendedNumberInput.spec.ts
```diff
@@ -1,36 +1,119 @@
...
+	} );
+
+	it( 'set different value after mount', async () => {
+		const wrapper = mount( ExtendedNumberInput );
+
+		const inputElement = wrapper.find( 'input' );
+		expect( ( inputElement.element as HTMLFormElement ).value ).toBe( '' );
+
+		wrapper.setProps( { value: { number: 123, precision: 0.5 } } );
+		await Vue.nextTick();
```
bereket-WMDE, line 42: taking from Amir's suggestion [here](https://github.com/wmde/query-builder/pull/214) , 
we can get rid of these `Vue.nexttrick()` things. shouldn't be too much effort

micgro42: Good point, fixed

bereket-WMDE:
It looks good from my side. Unless other's have further comments.

bereket-WMDE:
HI @micgro42 
This ticket has been open for  awhile now. Can we close it ?

bereket-WMDE:
This ticket is currently "on hold". currently this components is not consumed by anything (since the QB has a different solution ).
when we need a Wikidata numeric input, this can be re-opened

PR #383, opened 2021-02-26
Add InputExtender tokens

SaiSan-WMDE:
This hands over (hopefully all) the tokens needed to style the InputExtender, a menu-like element that will support the provision of data in WD/WB-specific inputs.

Issue: [T275758](https://phabricator.wikimedia.org/T275758)


Ladsgroup (approved):
Approving the tokens

PR #386, opened 2021-03-02
add_named_slot_for_icon_to_design_system_components_with_label

bereket-WMDE:
Bug: [T274346](https://phabricator.wikimedia.org/T274346)


vue-components/src/components/Checkbox.vue
```diff
@@ -53,12 +64,26 @@ export default Vue.extend( {
...
+		}
+
+		&--rtl {
+			margin-inline-start: $dimension-layout-xxsmall;
+		}
+
+		.wikit-Icon {
+			vertical-align: middle;
+		}
+
```
bereket-WMDE, line 86: could be a mixin ..

vue-components/src/components/Checkbox.vue
```diff
@@ -9,10 +9,13 @@
 			:checked="checked"
 			:disabled="disabled"
 		>
+		<slot v-if="this.textDirection == 'rtl'" name="suffix" />
 		<label
 			class="wikit-checkbox__label"
+			:class="[ `wikit-checkbox__label`, `wikit-checkbox__label--${textDirection}` ]"
 			:for="id"
 		>{{ label }}</label>
+		<slot v-if="this.textDirection == 'ltr'" name="suffix" />
```
bereket-WMDE, line 18: In the case of the checkbox, I put the `slot` outside of the `label`, since we don't want the checkbox to be checked when the user clicks on the icon (we want the popover to show). the `label` is tied to the `Checkbox` (`:for="id"`)

vue-components/src/components/Checkbox.vue
```diff
@@ -9,10 +9,13 @@
 			:checked="checked"
 			:disabled="disabled"
 		>
+		<slot v-if="this.textDirection == 'rtl'" name="suffix" />
```
micgro42, line 12: No. We don't want to explicitly specify with props whether this is `ltr` or `rtl`. There are HTML5 technologies that do this for us. The easiest might be flexbox.

Have a look at this commit: https://github.com/wmde/wikit/commit/667cb5b30306ddf5cf2bf23c3ceb64ed4f461c4d

micgro42 (changes requested):

SaiSan-WMDE:
Finally leaving some comments here after some discussion and trying some things out. I already had a chat with Bereket about all this, but:

- That "info outline" icon does definitely not look crisp. I tried some things locally (like replacing the svg path by that of the original WMF svg just in case, or applying `shape-rendering="crispEdges"`) but nothing worked, so I guess that the design of this particular icon does not adjust properly to this small size (the same can be seen [in OOUI](https://doc.wikimedia.org/oojs-ui/master/demos/?page=widgets&theme=wikimediaui&direction=ltr&platform=desktop#TextInputWidget-with-access-key))

- It would be ideal if the icon could be aligned with the last line of the label in case they wrap (instead of being vertically centered). But I'm aware this might not be easy to achieve, so this is not a strict requirement.

Image Pasted at 2021-3-3 12-38

- I noticed the following behavior where long labels take over a wider width, pushing the icon away. 

Screenshot 2021-03-03 at 18 43 28

bereket-WMDE:
>     * It would be ideal if the icon could be aligned with the last line of the label in case they wrap (instead of being vertically centered). But I'm aware this might not be easy to achieve, so this is not a strict requirement.
> 

As far as i know, the Icon has to be part of the label, in order to to achieve the desired result. but we don't want that, since the Icon has a different purpose (opening the popover component). right now, clicking the label would check the checkbox.  But that could be my limited styling knowledge :) I am happy to hear ideas from other devs.

bereket-WMDE:
>     * I noticed the following behavior where long labels take over a wider width, pushing the icon away.
> 
> 
> Screenshot 2021-03-03 at 18 43 28

Fixed in the new commit.

vue-components/src/components/Checkbox.vue
bereket-WMDE: looks much cleaner

vue-components/src/styles/mixins/Label.scss
```diff
@@ -5,8 +5,8 @@
     font-weight: $wikit-Label-font-weight;
     line-height: $wikit-Label-line-height;
     display: $displayType;
-     
-    @if $displayType == block {
+
+    @if $displayType == block or $displayType == inline-flex {
```
micgro42, line 9: Why do we need this change?

vue-components/src/components/Dropdown.vue
```diff
@@ -253,8 +254,15 @@ $base: '.wikit-Dropdown';
...
 
+	&__label-wrapper {
+		display: flex;
+		align-items: center;
+		gap: $dimension-layout-xxsmall;
+	}
+
 	&__label {
 		@include Label( block );
+		display: inine;
```
micgro42, line 265: typo: this should probably be `inline`? Similarly in the other files.

However, **this line can just be removed**, it is obviously not needed as it is currently ignored and the every looks fine.

micgro42 (changes requested):
Looks already pretty good, but there is some cleanup still needed.

vue-components/src/components/Checkbox.vue
```diff
@@ -53,11 +51,14 @@ export default Vue.extend( {
...
 .wikit-checkbox {
+	display: flex;
+
 	&__label {
 		@include Label;
-
 		display: flex;
 		align-items: center;
+		word-break: break-all;
+		margin-inline-end: $dimension-layout-xxsmall;
```
SaiSan-WMDE, line 61:
```suggestion
		margin-inline-end: $dimension-spacing-small;
```

SaiSan-WMDE: Agreeing with @micgro42's comment, since the spacing is applied at a component level.

vue-components/src/components/Dropdown.vue
```diff
@@ -253,8 +254,15 @@ $base: '.wikit-Dropdown';
 		position: absolute;
 	}
 
+	&__label-wrapper {
+		display: flex;
+		align-items: center;
+		gap: $dimension-layout-xxsmall;
```
SaiSan-WMDE, line 260:
```suggestion
		gap: $dimension-spacing-small;
```

vue-components/src/components/Lookup.vue
```diff
@@ -105,8 +109,15 @@ export default defineComponent( {
 
 <style lang="scss">
 .wikit-Lookup {
+	&__label-wrapper {
+		display: flex;
+		align-items: center;
+		gap: $dimension-layout-xxsmall;
```
SaiSan-WMDE, line 115:
```suggestion
		gap: $dimension-spacing-small;
```

vue-components/src/components/TextInput.vue
```diff
@@ -84,8 +87,15 @@ export default defineComponent( {
 
 <style lang="scss">
 .wikit-TextInput {
+	&__label-wrapper {
+		display: flex;
+		align-items: center;
+		gap: $dimension-layout-xxsmall;
```
SaiSan-WMDE, line 93:
```suggestion
		gap: $dimension-spacing-small;
```

vue-components/stories/Checkbox.stories.ts
```diff
@@ -32,16 +34,32 @@ export function basic( args ): Component {
...
-				<div><Checkbox label="Checked disabled" :checked="true" disabled="disabled"/></div>
+				<div>
+					<Checkbox label="Unchecked disabled" :checked="false" disabled="disabled"></Checkbox>
+				</div>
 				<br>
-				<div><Checkbox label="Unchecked disabled" :checked="false" disabled="disabled"/></div>
+				<div dir="ltr">
+					<Checkbox label="Unchecked with suffix icon" :checked="false">
+						<template v-slot:suffix>
+							<Icon type="info-outlined" size="small" color="base"/>
```
SaiSan-WMDE, line 59:
```suggestion
							<Icon type="info-outlined" size="medium" color="base"/>
```

vue-components/stories/Dropdown.stories.ts
```diff
@@ -143,7 +148,21 @@ export function all(): Component {
...
+				<p dir="ltr">
+					<Dropdown
+						label="Label with suffix icon"
+						:menu-items="menuItems"
+						v-model="selectedItem"
+						placeholder="Select an option"
+						:error="error"
+					>
+					<template v-slot:suffix>
+						<Icon type="info-outlined" size="small" color="base"/>
```
SaiSan-WMDE, line 163:
```suggestion
						<Icon type="info-outlined" size="medium" color="base"/>
```

vue-components/stories/Lookup.stories.ts
```diff
@@ -128,6 +129,33 @@ export function all(): Component {
...
+					<Lookup
+						:search-input.sync="search"
+						label="Label with suffix icon"
+						v-model="selectedItem"
+						:menu-items="menuItems"
+						placeholder="Placeholder"
+						@scroll="onScroll"
+					>
+						<template v-slot:suffix>
+							<Icon type="info-outlined" size="small" color="base"/>
```
SaiSan-WMDE, line 142:
```suggestion
							<Icon type="info-outlined" size="medium" color="base"/>
```

vue-components/stories/TextInput.stories.ts
```diff
@@ -9,13 +10,28 @@ export default {
...
+				<p>
+					<TextInput label="Label" :error="{ type: 'error', message: 'Inline error message' }" placeholder="Placeholder"></TextInput>
+				</p>
+				<p>
+					<TextInput label="Label" :error="{ type: 'warning', message: 'Warning message' }" placeholder="Placeholder"></TextInput>
+				</p>
+				<p dir="ltr">
+					<TextInput label="Label with icon suffix" placeholder="Placeholder">
+						<template v-slot:suffix>
+							<Icon type="info-outlined" size="small" color="base"/>
```
SaiSan-WMDE, line 31:
```suggestion
							<Icon type="info-outlined" size="medium" color="base"/>
```

vue-components/src/styles/mixins/Label.scss
bereket-WMDE: we don't. I forgot to remove that from a previous commit. 
removed now

vue-components/src/components/Dropdown.vue
bereket-WMDE: ah.. ofc this line is useless. thanks

micgro42 (commented):
Good from my side but it looks like chromatic needs another poke? Though the diff seems to be only a anti-aliasing issue

micgro42 (approved):

SaiSan-WMDE:
Approving this PR so we can move forward, although at some point (probably while working on [this Phab ticket](https://phabricator.wikimedia.org/T272609)? although the change does not only apply to checkboxes) we should come up with a way to align the icon with the top or the baseline of the first line of wrapping labels, instead of centering it in relation to the label text. This would prevent misalignment like the following:

Screenshot 2021-03-10 at 12 49 14

I'll document the fix in the linked ticket.

bereket-WMDE:
> Approving this PR so we can move forward, although at some point (probably while working on [this Phab ticket](https://phabricator.wikimedia.org/T272609)? Though the change does not only apply to checkboxes) we should come up with a way to align the icon with the top or the baseline of the first line of wrapping labels, instead of centering it in relation to the label text. This would prevent misalignment like the following:
> 
> Screenshot 2021-03-10 at 12 49 14
> 
> I'll document the fix in the linked ticket.

Yes, i agree with the approach

PR #387, opened 2021-03-04
Introduce LookupWithExtender

Ladsgroup:


vue-components/src/components/InputWithExtender.vue
```diff
@@ -0,0 +1,91 @@
...
+			type: String,
+			default: null,
+		},
+	},
+	methods: {
+		onInput( value: string ): void {
+			this.showExtension = true;
+			this.$emit( 'input', value );
+		},
+		onEsc(): void {
```
micgro42, line 64: This is unused, but it is functionality that we want.

micgro42 (commented):
Good start, but probably needs more work. Content inside the extender must be focusable. From the current implementation I would guess that it isn't.

Could you add a link to the content of one of the storybook components so that it can be tried out?

vue-components/src/components/InputWithExtender.vue
micgro42: Looks good but doesn't work :upside_down_face: 

I think you may have to emit that event inside the TextInput as well

micgro42 (approved):
Good from my side, but needs the ok from UX to be published as-is for now until we start implementing the DateInput

SaiSan-WMDE:
Looks good but unfortunately the placement of the extender menu of the validated components is off:

Screenshot 2021-03-10 at 15 57 30

Ladsgroup:
> Looks good but unfortunately the placement of the extender menu of the validated components is off:
> 
> Screenshot 2021-03-10 at 15 57 30

yeah, the validation message is a part of the input though. Meaning unless I make css magic to give negative positioning, I don't see an easy way to fix it...

SaiSan-WMDE:
> > Screenshot 2021-03-10 at 15 57 30
> 
> yeah, the validation message is a part of the input though. Meaning unless I make css magic to give negative positioning, I don't see an easy way to fix it...

I cannot quickly try it myself because I'm finding merge conflicts when trying to pull this but locally, but `z-index: 1` might do it? In the dropdown component, the menu also has `position: static`

micgro42:
> > Looks good but unfortunately the placement of the extender menu of the validated components is off:
> > Screenshot 2021-03-10 at 15 57 30
> 
> yeah, the validation message is a part of the input though. Meaning unless I make css magic to give negative positioning, I don't see an easy way to fix it...

The brute force way out would probably be to stop using the TextInput and to start using the Input and ValidationMessage directly. That would mean we wouldn't have to re-emit those events in TextInput.

The trouble with positioning will be that it will be unsure how much it needs to be repositioned. That error or warning message could have multiple lines. Maybe it would be possible to reposition it from the top with `top: $label-height + $label-margin + $input-height;` but that feels also very hacky and brittle

SaiSan-WMDE:
> Content inside the extender must be focusable. From the current implementation I would guess that it isn't.
> Could you add a link to the content of one of the storybook components so that it can be tried out?

If we suspect that not including a focusable element in the menu now could involve having to redo any parts of this core component while working on the DateInput (I'm not sure if that's the case), then I would probably try that out now? If the suggestion was only for demonstration purposes, then I wouldn't say this a requirement at the moment.

vue-components/src/components/InputWithExtender.vue
SaiSan-WMDE: just a heads-up that this seems to not be working any more

vue-components/src/components/InputWithExtender.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+	<div
+		:class="[ 'wikit', 'wikit-InputWithExtender' ]"
+	>
+		<label class="wikit-InputWithExtender__label" :for="id">{{ label }}</label>
```
micgro42, line 5: This is going to need the suffix slot as well, like the ones added in #386.

(We probably should create Label component or something like that at some point :thinking:)

vue-components/src/components/InputWithExtender.vue
```diff
@@ -0,0 +1,106 @@
...
+		<Input
+			:id="id"
+			:value="value"
+			@focus.native="showExtension = true"
+			@blur.native="showExtension = false"
+			@esc.native="showExtension = false"
+			@input="onInput"
+			:placeholder="placeholder"
+			:disabled="disabled"
+		/>
```
micgro42, line 15: This is missing the feedbackType prop, so the border around the input with an error message are currently not red. See the TextInput for an example.

vue-components/src/components/InputWithExtender.vue
```diff
@@ -0,0 +1,106 @@
...
+	<div
+		:class="[ 'wikit', 'wikit-InputWithExtender' ]"
+	>
+		<label class="wikit-InputWithExtender__label" :for="id">{{ label }}</label>
+		<Input
+			:id="id"
+			:value="value"
+			@focus.native="showExtension = true"
+			@blur.native="showExtension = false"
+			@esc.native="showExtension = false"
```
micgro42, line 11: That is actually not working, because there is no `@esc` event in vuejs. It needs to use the `@keydown` event with the `.esc` modifier: `@keydown.esc="..."`, see: https://vuejs.org/v2/guide/events.html#Key-Modifiers

vue-components/src/components/InputWithExtender.vue
Ladsgroup: Done

vue-components/src/components/InputWithExtender.vue
Ladsgroup: Done

vue-components/src/components/InputWithExtender.vue
Ladsgroup: Done

Ladsgroup: keydown.esc didn't work on its own, needed native too but worked after it

SaiSan-WMDE (approved):
Looks good!

giphy (2)

vue-components/src/components/InputWithExtender.vue
micgro42: Almost, this needs a [wrapping `<span>`](https://github.com/wmde/wikit/pull/386/files#diff-2ea7a99f61da845846aec597ca9e2163ad2b2b0bf9e46b769e93860331d250aaR6-R9) and [some styles](https://github.com/wmde/wikit/pull/386/files#diff-2ea7a99f61da845846aec597ca9e2163ad2b2b0bf9e46b769e93860331d250aaR257-R261) so that it works nicely with rtl as well.

micgro42 (changes requested):
One small thing is left, then this is good to go

Ladsgroup:
Done

micgro42 (approved):
\o/

PR #391, opened 2021-03-10
Create bouncingDots Loader Component

bereket-WMDE:
Bug: [T267010](https://phabricator.wikimedia.org/T267010)

- creates the bouncing dot loader component
- creates story documentation


vue-components/stories/BouncingDots.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+		`,
+    };
+}
+
+basic.args = {
+    size: 'small',
+};
+
+basic.argTypes = {
+    position: {
```
SaiSan-WMDE, line 33: was this intended for "size"?

SaiSan-WMDE: (I think a selector control for size is the best choice)

vue-components/stories/BouncingDots.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+				<div v-for="size in bouncingDotsSizes">
+                    {{ size }}<br><br>
+                    <BouncingDots :size="size"></BouncingDots><br><br>
+				</div>
+            </div>
+		`,
+    };
+}
+
+basic.args = {
```
SaiSan-WMDE, line 28: I think that using a selector control for size is the best choice, so the options are visible for users.

vue-components/stories/BouncingDots.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+    size: 'small',
```
SaiSan-WMDE, line 29: I think that using a selector control for size is the best choice. Like that, the options are visible for users.

vue-components/stories/BouncingDots.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+};
+
+basic.argTypes = {
+    position: {
+		control: {
```
SaiSan-WMDE, line 34: we can remove the controls, since both sizes are visible on the page

vue-components/stories/BouncingDots.stories.ts
```diff
@@ -0,0 +1,39 @@
+import BouncingDots from '@/components/BouncingDots';
+import { bouncingDotsSizes } from '@/components/bouncingDots';
+import { Component } from 'vue';
+
+export default {
+    component: BouncingDots,
+    title: '/Loader/BouncingDots',
+};
+
+export function basic( args: { content: string; underlined: boolean } ): Component {
```
SaiSan-WMDE, line 10: I believe this page can be called "All", since it contains all size variants

vue-components/stories/BouncingDots.stories.ts
```diff
@@ -0,0 +1,39 @@
...
+    return {
+        data(): object {
+            return { args, bouncingDotsSizes };
+        },
+        components: { BouncingDots },
+        props: Object.keys( args ),
+        template: `
+			<div>
+				<div v-for="size in bouncingDotsSizes">
```
SaiSan-WMDE, line 19:
```suggestion
				<div v-for="size in bouncingDotsSizes" style="font-family: sans-serif; color: #202122">
```

SaiSan-WMDE: I'm hoping this is the way to modify the text style of the size indicator. 

I was trying to do this tinny change locally, but I'm getting the following error (probably irrelevant):

Screenshot 2021-03-11 at 10 17 47

vue-components/src/components/bouncingDots.ts
```diff
@@ -0,0 +1 @@
+export const bouncingDotsSizes = [ 'small', 'medium' ];
```
micgro42, line 1: It might be that things get confused by having the same filename that differs only by extension and capitalization. Better to rename this file to `bouncingDotsProps.ts` or something like that

vue-components/stories/BouncingDots.stories.ts
micgro42: And since this is supposed to not contain controls, we con drop the `args` parameter and every place where it is used.

vue-components/stories/BouncingDots.stories.ts
micgro42: This configuration can then be replaced with something like there:

https://github.com/wmde/wikit/blob/65074e872c86613c4326a5332aa8a5fcbee76e55/vue-components/stories/QuantityInput.stories.ts#L199-L201

See also https://storybook.js.org/docs/react/essentials/controls#hide-nocontrols-warning
(and the entire page might be of interest in general)

bereket-WMDE: @SaiSan-WMDE thank you for the update
@micgro42 thanks for the resource

vue-components/stories/BouncingDots.stories.ts
bereket-WMDE: removed unnecessary `args` parameter

vue-components/src/components/bouncingDots.ts
bereket-WMDE: yea makes sense to change the name

SaiSan-WMDE (approved):
LGTM!

giphy

micgro42 (approved):

PR #395, opened 2021-03-17
Emit scroll event in QuantityInput

guergana:


Ladsgroup:
Not against this PR per se but scroll event is quite a ticking bomb. Basically it emits scroll events on desktop with pace of 30/s and in mobile it's 100/s (which brought down Twitter once: https://johnresig.com/blog/learning-from-twitter/). I have been going around deleting onscroll handlers in our codebase these days.
Can you explain why this is needed and for what?

bereket-WMDE:
> Not against this PR per se but scroll event is quite a ticking bomb. Basically it emits scroll events on desktop with pace of 30/s and in mobile it's 100/s (which brought down Twitter once: https://johnresig.com/blog/learning-from-twitter/). I have been going around deleting onscroll handlers in our codebase these days.
> Can you explain why this is needed and for what?


This scroll event is needed to load more results upon scroll. (just like what we've in the lookup component). normally, only 12 results are loaded first time and only the first six results are shown. then when the user scrolls down, we make a request to the API and load more results. 

and the scroll event is only tied to the lookup part (not the entire page). so i think we're fine....for now

guergana:
> Not against this PR per se but scroll event is quite a ticking bomb. Basically it emits scroll events on desktop with pace of 30/s and in mobile it's 100/s (which brought down Twitter once: https://johnresig.com/blog/learning-from-twitter/). I have been going around deleting onscroll handlers in our codebase these days.
> Can you explain why this is needed and for what?

This is usually debounced by the consumer application.

Ladsgroup:
okay then. Please debounce and throttle (if making API requests).

Ladsgroup (approved):

bereket-WMDE (approved):

PR #398, opened 2021-03-18
Start DateInput

Ladsgroup:
Bug: [T273002](https://phabricator.wikimedia.org/T273002)

### DIFF
--- a/vue-components/src/components/DateInput.vue
+++ b/vue-components/src/components/DateInput.vue
@@ -20,7 +20,7 @@
 				<slot v-if="!value" name="prompt" />
 				<BouncingDots v-if="!parsedValue && value" type="small" />
 				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
-					<Icon type="info" size="small" />
+					<Icon color="inherit" type="info" size="small" />
 					{{ calendarNotice }}
 				</div>
 			</template>
@@ -85,6 +85,7 @@ export default defineComponent( {
 <style lang="scss">
 .wikit-DateInput {
 	&__IntroText {
+		padding-block-end: $wikit-InputExtender-content-padding-vertical;
 		font-family: $font-family-style-label;
 		font-weight: $font-weight-style-label;
 		font-size: $font-size-style-label;
@@ -93,6 +94,7 @@ export default defineComponent( {
 	}
 
 	&__ParsedValue {
+		padding-block: $wikit-InputExtender-content-padding-vertical;
 		font-family: $font-family-style-label;
 		font-weight: $font-weight-bold;
 		font-size: $font-size-style-label;
@@ -101,6 +103,7 @@ export default defineComponent( {
 	}
 
 	&__CalendarNotice {
+		padding-block-start: $wikit-InputExtender-content-padding-vertical;
 		font-family: $font-family-style-description;
 		font-weight: $font-weight-style-description;
 		font-size: $font-size-style-description;
--- a/vue-components/src/components/DateInput.vue
+++ b/vue-components/src/components/DateInput.vue
@@ -17,7 +17,9 @@
 				<p class="wikit-DateInput__ParsedValue" v-if="parsedValue">
 					<b>{{ parsedValue }}</b>
 				</p>
-				<slot v-if="!value" name="prompt" />
+				<p class="wikit-DateInput__Prompt" v-if="promptText && !value">
+					{{ promptText }}
+				</p>
 				<BouncingDots v-if="!parsedValue && value" type="small" />
 				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
 					<Icon color="inherit" type="info" size="small" />
@@ -70,6 +72,10 @@ export default defineComponent( {
 			type: String,
 			default: null,
 		},
+		promptText: {
+			type: String,
+			default: null,
+		},
 		value: {
 			type: String,
 			default: null,
@@ -108,6 +114,14 @@ export default defineComponent( {
 		font-weight: $font-weight-style-description;
 		font-size: $font-size-style-description;
 		line-height: $font-line-height-style-description;
+		color: $font-color-subtle;
+	}
+
+	&__Prompt {
+		font-family: $font-family-style-placeholder;
+		font-size: $font-size-style-placeholder;
+		font-weight: $font-weight-style-placeholder;
+		line-height: $font-line-height-style-placeholder;
 		color: $font-color-subtle;
 	}
 }
--- a/vue-components/src/components/DateInput.vue
+++ b/vue-components/src/components/DateInput.vue
@@ -20,7 +20,7 @@
 				<slot v-if="!value" name="prompt" />
 				<BouncingDots v-if="!parsedValue && value" type="small" />
 				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
-					<Icon type="info-outlined" size="small" />
+					<Icon type="info-filled" size="small" />
 					{{ calendarNotice }}
 				</div>
 			</template>
--- a/vue-components/src/components/DateInput.vue
+++ b/vue-components/src/components/DateInput.vue
@@ -1,0 +1,111 @@
+<template>
+	<div
+		:class="[ 'wikit', 'wikit-DateInput' ]"
+	>
+		<InputWithExtender
+			:label="label"
+			:value="value"
+			:error="error"
+			:placeholder="placeholder"
+			:disabled="disabled"
+			@input="$emit( 'input', $event )"
+		>
+			<template v-slot:default>
+				<p class="wikit-DateInput__IntroText">
+					{{ resultsIntroText }}
+				</p>
+				<p class="wikit-DateInput__ParsedValue" v-if="parsedValue">
+					<b>{{ parsedValue }}</b>
+				</p>
+				<slot v-if="!value" name="prompt" />
+				<BouncingDots v-if="!parsedValue && value" type="small" />
+				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
+					<Icon type="info-outlined" size="small" />
+					{{ calendarNotice }}
+				</div>
+			</template>
+			<template v-slot:suffix>
+				<slot name="suffix" />
+			</template>
+		</InputWithExtender>
+	</div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import VueCompositionAPI, { defineComponent } from '@vue/composition-api';
+import { errorProp } from '@/compositions/validatable';
+import Icon from '@/components/Icon.vue';
+import InputWithExtender from '@/components/InputWithExtender.vue';
+import BouncingDots from '@/components/BouncingDots.vue';
+
+Vue.use( VueCompositionAPI );
+
+export default defineComponent( {
+	name: 'DateInput',
+	components: {
+		BouncingDots,
+		Icon,
+		InputWithExtender,
+	},
+	props: {
+		error: errorProp,
+		disabled: {
+			type: Boolean,
+			default: false,
+		},
+		placeholder: {
+			type: String,
+			default: '',
+		},
+		label: {
+			type: String,
+			default: null,
+		},
+		parsedValue: {
+			type: String,
+			default: null,
+		},
+		resultsIntroText: {
+			type: String,
+			default: null,
+		},
+		value: {
+			type: String,
+			default: null,
+		},
+		calendarNotice: {
+			type: String,
+			default: null,
+		},
+	},
+} );
+</script>
+
+<style lang="scss">
+.wikit-DateInput {
+	&__IntroText {
+		font-family: $font-family-style-label;
+		font-weight: $font-weight-style-label;
+		font-size: $font-size-style-label;
+		line-height: $font-line-height-style-label;
+		color: $font-color-base;
+	}
+
+	&__ParsedValue {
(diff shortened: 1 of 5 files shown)


vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,113 @@
...
+	>
+		<InputWithExtender
+			:label="label"
+			:value="value"
+			:feedback-type="feedbackType"
+			:placeholder="placeholder"
+			:disabled="disabled"
+			@input="$emit( 'input', $event )"
+		>
+			<template v-slot:default>
```
micgro42, line 13: This also has to reproduce the suffix slot

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,113 @@
...
+				<slot v-if="!value" name="prompt" />
+				<p class="wikit-DateInput--IntroText" v-else>
+					{{ resultsIntroText }}
+				</p>
+				<p class="wikit-DateInput--ParsedValue" v-if="parsedValue && value">
+					<b>{{ parsedValue }}</b>
+				</p>
+				<BouncingDots v-if="!parsedValue && value" type="small" />
+				<div class="wikit-DateInput--CalendarNotice" v-if="calendarNotice && parsedValue">
```
micgro42, line 22: I think this class should be `wikit-DateInput__CalendarNotice` - it is an element, not a modifier, right?

micgro42 (commented):
looks good so far :+1:

vue-components/src/components/DateInput.vue
Ladsgroup: Done.

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,117 @@
+<template>
+	<div
+		:class="[ 'wikit', 'wikit-DateInput' ]"
+	>
+		<InputWithExtender
+			:label="label"
+			:value="value"
+			:error="error"
+			:feedback-type="feedbackType"
```
micgro42, line 9: InputWithExtender doesn't have a prop "feedback" type. The error prop is good enough

micgro42 (commented):
Solid job! just one little detail to fix and it can have my approval :+1:

vue-components/src/components/DateInput.vue
Ladsgroup: oh nice. Thanks. done now.

SaiSan-WMDE:
This is really great! 💯 

There's a Reference error that prevents the Basic story to be displayed in Chromatic. Adding some comments here after checking the components in the All story:

- For some reason, the IntroText style is not being applied to the default prompt:
 
Screenshot 2021-03-19 at 12 59 12

- The variants in the All page should have some more vertical space between them (see the All story of the ExtendedNumberInput for reference). I believe that the spacing is achieved by containing the variants in paragraphs.

- In the design specification, it was indicated that text elements should be contained in 24px high "bounding boxes" spaced by 4px. Checking the current implementation, I believe that probably that's not necessary, but that definitely the vertical spacing between elements should be increased to 8px (dimension.spacing.small.value). I wasn't sure where to suggest the change

SaiSan-WMDE:
Charlie made a good point regarding the default active state of the DateInput component that made me think the interaction flow could be improved in the following way:

Current: Once the user focuses on the DateInput, the extender menu opens to display the prompt text.
Desired: Once the user focuses on the DateInput, instead of just the prompt, users also see the resultsIntroText. In the following way:

Screenshot 2021-03-30 at 10 39 48

Since the component is still in the last stages of development, I thought we still are on time to make this improvement 🙏

Ladsgroup:
> This is really great! 100
> 
> There's a Reference error that prevents the _Basic_ story from being displayed in Chromatic. Adding some comments here after checking the components in the _All_ story:
> 
>     * For some reason, the IntroText style is not being applied to the default prompt in Chrome & Safari on MacOs (the style looks good in Firefox/Linux, though):
> 
> 
> Screenshot 2021-03-19 at 12 59 12
> 
>     * The variants in the All page should have some more vertical space between them (see the All story of the ExtendedNumberInput for reference). I believe that the spacing is achieved by containing the variants in paragraphs (which is probably also not ideal).
> 

Added `<br>` in between.

>     * In the design specification, it was indicated that text elements inside the Extender menu should be contained in 24px high "bounding boxes" (aka divs) verticall-spaced by 4px. Checking the current implementation, I'm not sure if that's applied or even if it's necessary, but definitely the vertical spacing between elements should be increased, to probably 8px (dimension.spacing.small.value). I wasn't sure where to suggest the change

Isn't it `$wikit-InputExtender-padding-vertical`?

SaiSan-WMDE:
Thanks for all the changes and fixes, @Ladsgroup! 

> > * For some reason, the IntroText style is not being applied to the default prompt in Chrome & Safari on MacOs (the style looks good in Firefox/Linux, though):

This error still persists: the text-styles applied to the prompt are ignored in MacOS (checked Safari, Firefox and Chrome)

> Isn't it `$wikit-InputExtender-padding-vertical`?

Yes, that token is defining the internal padding of the menu, but I believe that the elements are unaffected because we are using <p>s.

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+				<p class="wikit-DateInput__IntroText">
+					{{ resultsIntroText }}
+				</p>
+				<p class="wikit-DateInput__ParsedValue" v-if="parsedValue">
+					<b>{{ parsedValue }}</b>
+				</p>
+				<slot v-if="!value" name="prompt" />
+				<BouncingDots v-if="!parsedValue && value" type="small" />
+				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
+					<Icon type="info-outlined" size="small" />
```
SaiSan-WMDE, line 23:
```suggestion
					<Icon type="info-filled" size="small" />
```

SaiSan-WMDE: my bad! I realized we should be using the icon that accompanies notice messages.

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+				<p class="wikit-DateInput__IntroText">
+					{{ resultsIntroText }}
+				</p>
+				<p class="wikit-DateInput__ParsedValue" v-if="parsedValue">
+					<b>{{ parsedValue }}</b>
+				</p>
+				<slot v-if="!value" name="prompt" />
+				<BouncingDots v-if="!parsedValue && value" type="small" />
+				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
+					<Icon type="info-filled" size="small" />
```
SaiSan-WMDE, line 23:
```suggestion
					<Icon type="info" size="small" />
```

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+			default: null,
+		},
+	},
+} );
+</script>
+
+<style lang="scss">
+.wikit-DateInput {
+	&__IntroText {
+		font-family: $font-family-style-label;
```
SaiSan-WMDE, line 88:
```suggestion
		padding-block: $wikit-InputExtender-content-padding-vertical;
		font-family: $font-family-style-label;
```

SaiSan-WMDE, line 88:
```suggestion
		padding-block: $wikit-InputExtender-content-padding-vertical;
		font-family: $font-family-style-label;
```

SaiSan-WMDE: should be added to each one of these, right? 

sorry, I'd usually try this out locally, but I find lots of conflicts that I'm not sure how to resolve when I try to work on this branch locally

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+				<p class="wikit-DateInput__IntroText">
+					{{ resultsIntroText }}
+				</p>
+				<p class="wikit-DateInput__ParsedValue" v-if="parsedValue">
+					<b>{{ parsedValue }}</b>
+				</p>
+				<slot v-if="!value" name="prompt" />
+				<BouncingDots v-if="!parsedValue && value" type="small" />
+				<div class="wikit-DateInput__CalendarNotice" v-if="calendarNotice && parsedValue">
+					<Icon type="info" size="small" />
```
SaiSan-WMDE, line 23: can we apply `color="subtle"` to the icon or make it inherit the color of the text it accompanies? Any of the options are fine

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+			type: String,
+			default: null,
+		},
+	},
+} );
+</script>
+
+<style lang="scss">
+.wikit-DateInput {
+	&__IntroText {
```
SaiSan-WMDE, line 87: I guess we should add `padding-bottom: $wikit-InputExtender-content-padding-vertical;` here

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+		font-family: $font-family-style-label;
+		font-weight: $font-weight-style-label;
+		font-size: $font-size-style-label;
+		line-height: $font-line-height-style-label;
+		color: $font-color-base;
+	}
+
+	&__ParsedValue {
```
SaiSan-WMDE, line 95: And `padding-bottom: $wikit-InputExtender-content-padding-vertical;` & `padding-top: $wikit-InputExtender-content-padding-vertical;` here

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,111 @@
...
+		font-family: $font-family-style-label;
+		font-weight: $font-weight-bold;
+		font-size: $font-size-style-label;
+		line-height: $font-line-height-style-label;
+		color: $font-color-base;
+	}
+
+	&__CalendarNotice {
```
SaiSan-WMDE, line 103: And `padding-top: $wikit-InputExtender-content-padding-vertical;` here

vue-components/src/components/DateInput.vue
Ladsgroup: Done

vue-components/src/components/DateInput.vue
Ladsgroup: Done

vue-components/src/components/DateInput.vue
Ladsgroup: Done

Ladsgroup: We don't have "subtle" color type. The ones are `'base', 'warning', 'error', 'notice', 'success', 'inherit'`. I change it to inherit.

SaiSan-WMDE:
yay! looks so good! I would approve right away if it weren't (again) for that missing style:

Screenshot 2021-03-30 at 17 58 14

Is there anything we can try there? Should we create a class for it too? The style literally matches that of the input [placeholder](https://www.figma.com/file/gi3JYwSUQYhKS0mHwK25cQwi/WiKit?node-id=7000%3A9060).

vue-components/src/components/DateInput.vue
SaiSan-WMDE: thanks! inherit makes sense. I'm opening a small PR to add those missing colors (I was convinced they were there, but maybe mixed up colors and sizes), but no further change is needed here.

vue-components/src/components/DateInput.vue
```diff
@@ -0,0 +1,114 @@
...
+			@input="$emit( 'input', $event )"
+		>
+			<template v-slot:default>
+				<p class="wikit-DateInput__IntroText">
+					{{ resultsIntroText }}
+				</p>
+				<p class="wikit-DateInput__ParsedValue" v-if="parsedValue">
+					<b>{{ parsedValue }}</b>
+				</p>
+				<slot v-if="!value" name="prompt" />
```
micgro42, line 20: Not blocking this PR: But since we now show the intro text always, do we want the prompt to actually still be a slot? That is where the application can inject whatever markup it wants. Or would we not prefer for this now to be a text-only prop? @SaiSan-WMDE 

Though, this is not blocking the merge and can still be adjusted later on.

micgro42:
This still needs a line in the CHANGELOG.md, I think :)

vue-components/src/components/DateInput.vue
SaiSan-WMDE: @Ladsgroup and I just discussed this and originally agreed to keep the slot. But you're right, this is some static text that will always be displayed, so using a text-only prop like in the other cases makes more sense and is more consistent?

micgro42: You tell me 🙃
This really depends on what you envision might be inserted there by the application.
If it is only ever going to be plain text that is always supposed to be styled the same way, then a prop makes more sense.
On the other hand, if you want to keep the possibility open that the application might want to provide some copy that includes a "learn more" link, then we should keep the slot.

If you two discussed this and decided that to keep the slot, then I'm confident that this is the right decision and am very happy to approve this as soon as we have the note in the CHANGELOG.md :)

SaiSan-WMDE: I believe this should be a text slot for now, since I cannot foresee any use cases that justify the use of a slot at the moment. A slot could be introduced later if that changed.

micgro42: Implemented as requested.

SaiSan-WMDE (approved):
Just gorgeous. Thanks, Michael and Amir! 💯 

giphy (4)

micgro42 (approved):

PR #401, opened 2021-03-31
Add missing color classes and props to icon component

SaiSan-WMDE:
This should allow applying the subtle, emphasized and inverted colors to the icon component.

This is what happens when you keep forgetting that you can run tests locally 🙈


Ladsgroup (changes requested):
This needs more work, the icons need to be added to the icon colors ts file, otherwise the validation fails.

Ladsgroup (approved):
Thanks!

PR #405, opened 2021-04-01
Create default placeholder class for Extender

SaiSan-WMDE:
Adding an extra class to (hopefully) style the text included in the Extender menu by default.

### DIFF
--- a/tokens/properties/components/InputExtender.json
+++ b/tokens/properties/components/InputExtender.json
@@ -39,6 +39,23 @@
                 "vertical": {
                     "value": "{dimension.spacing.xsmall.value}"
                 }
+            },
+            "placeholder": {
+                "font-family": {
+                    "value": "{font.family.style.placeholder.value}"
+                },
+                "font-size": {
+                    "value": "{font.size.style.placeholder.value}"
+                },
+                "font-weight": {
+                    "value": "{font.weight.style.placeholder.value}"
+                },
+                "line-height": {
+                    "value": "{font.line-height.style.placeholder.value}"
+                },
+                "color": {
+                    "value": "{font.color.subtle.value}"
+                }
             }
         }
     }
--- a/vue-components/src/components/InputWithExtender.vue
+++ b/vue-components/src/components/InputWithExtender.vue
@@ -17,7 +17,7 @@
 			:placeholder="placeholder"
 			:disabled="disabled"
 		/>
-		<div v-if="showExtension" class="wikit-InputWithExtender__extension">
+		<div v-if="showExtension" class="wikit-InputWithExtender__extension wikit-InputWithExtender__placeholder">
 			<slot />
 		</div>
 		<ValidationMessage
@@ -107,6 +107,14 @@ export default defineComponent( {
 		border: $wikit-InputExtender-border-width $wikit-InputExtender-border-style $wikit-InputExtender-border-color;
 		box-shadow: $wikit-InputExtender-box-shadow;
 		border-radius: $wikit-InputExtender-border-radius;
+	}
+
+	&__placeholder {
+		font-family: $wikit-InputExtender-content-placeholder-font-family;
+		font-size: $wikit-InputExtender-content-placeholder-font-size;
+		font-weight: $wikit-InputExtender-content-placeholder-font-weight;
+		line-height: $wikit-InputExtender-content-placeholder-line-height;
+		color: $wikit-InputExtender-content-placeholder-color;
 	}
 
 }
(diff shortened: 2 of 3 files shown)


vue-components/src/components/InputWithExtender.vue
```diff
@@ -109,6 +109,14 @@ export default defineComponent( {
 		border-radius: $wikit-InputExtender-border-radius;
 	}
 
+	&__placeholder {
```
micgro42, line 112: What is the motivation behind this second class instead of adding those styles to `wikit-InputWithExtender__extension`?

SaiSan-WMDE, line 112:
```suggestion
	&__content-placeholder {
```

vue-components/src/components/InputWithExtender.vue
```diff
@@ -17,7 +17,7 @@
 			:placeholder="placeholder"
 			:disabled="disabled"
 		/>
-		<div v-if="showExtension" class="wikit-InputWithExtender__extension">
+		<div v-if="showExtension" class="wikit-InputWithExtender__extension wikit-InputWithExtender__placeholder">
```
SaiSan-WMDE, line 20:
```suggestion
		<div v-if="showExtension" class="wikit-InputWithExtender__extension wikit-InputWithExtender__content-placeholder">
```

vue-components/src/components/InputWithExtender.vue
```diff
@@ -17,7 +17,7 @@
 			:placeholder="placeholder"
 			:disabled="disabled"
 		/>
-		<div v-if="showExtension" class="wikit-InputWithExtender__extension">
+		<div v-if="showExtension" class="wikit-InputWithExtender__extension wikit-InputWithExtender__content-placeholder">
```
SaiSan-WMDE, line 20:
```suggestion
		<div v-if="showExtension" class="wikit-InputWithExtender__extension 
		wikit-InputWithExtender__content-placeholder">
```
```suggestion
		<div v-if="showExtension" class="wikit-InputWithExtender__extension 
		wikit-InputWithExtender__content-placeholder">
```

SaiSan-WMDE, line 20:
```suggestion
		<div v-if="showExtension" class="wikit-InputWithExtender__extension 
		wikit-InputWithExtender__content-placeholder">
```

vue-components/src/components/InputWithExtender.vue
```diff
@@ -109,6 +109,14 @@ export default defineComponent( {
 		border-radius: $wikit-InputExtender-border-radius;
 	}
 
+	&__content-default {
```
micgro42, line 112: I'm confused. What is the motivation to not add those styles to the `wikit-InputWithExtender__extension` class?

SaiSan-WMDE: that's what I tried originally, but the styles were ignored 🤔  I'll try again with a fresh look after this while

SaiSan-WMDE: done. not sure what made me create a different class originally 🤷‍♀️

tokens/properties/components/InputExtender.json
```diff
@@ -39,6 +39,23 @@
                 "vertical": {
                     "value": "{dimension.spacing.xsmall.value}"
                 }
+            },
+            "placeholder": {
```
micgro42, line 43: I'm not sure `placeholder` is the ideal term here as I feel that this is already a fixed term that refers to the literal placeholder in `<input>` HTML elements. (Also given that this component _does_ include such a input element, even though just as a sub-component)

How about calling this `content` instead, as these styles are intentionally applied to the content of the extender? (even though they are overridable by the application)

micgro42 (commented):
I like it and am willing to approve it from a technical perspective. However,  I still have one thought on a naming issue and would like to check-in on that one first and hear your thoughts.

tokens/properties/components/InputExtender.json
SaiSan-WMDE: TL;DR: I'll remove the placeholder level and adjust the tokens to make them reflect the default text style of the menu's content.

I hesitated to use the term "placeholder" for the reason you're mentioning. I thought that the object structure ("content-placeholder") would clarify the meaning, but I agree that the name of the variables can still be confusing in their final context (the InputWithExtender.vue file). Probably the whole placeholder approach was wrong, I think we should look at these as the default text styles of the menu content, instead of as some temporary default.

micgro42 (approved):
Looks great!

PR #410, opened 2021-04-12
Avoid emitting 'update:isShown' when not needed

Ladsgroup:
Currently, doing any debug shows a lot of these events being emitted
making debugging much harder (and also has performance costs). Let's
just not do it.


micgro42 (approved):
I wonder if the `clickOutsideHandler` should attached and detached more dynamically, but your solution is a good and easy first fix :+1:

PR #411, opened 2021-04-12
Add preventDefault to keydown handlers of menu items and dropdown

Ladsgroup:
Similar to what ooui has done:
https://doc.wikimedia.org/oojs-ui/master/js/source/DropdownWidget.html#OO-ui-DropdownWidget

This prevents the whole screen to scroll when you use keyboard to
navigate through items.

Bug: T277759


micgro42 (approved):
Tested and it works! :+1:

PR #413, opened 2021-04-14
Fix label word wrap

SaiSan-WMDE:
This should prevent the words in any label text from breaking arbitrarily whenever their content wraps. Hyphenation is also introduced to prevent creating disproportionate white space between lines when long words are present.

Before:

Screenshot 2021-04-14 at 19 29 09

After:

Screenshot 2021-04-14 at 19 29 42


Ladsgroup (approved):

Ladsgroup:
We probably should fix the indentation of the scss file but that's for later.

PR #418, opened 2021-04-29
Install jest-axe and write a11y tests

tzhelyazkova:
The tests caught one a11y issue which this change fixes


micgro42 (approved):
Looks good to me :+1:

itamargiv (approved):
I like what I see! I wonder if there's more to the API then just "toHaveNoViolations"

tzhelyazkova:
> I like what I see! I wonder if there's more to the API then just "toHaveNoViolations"

No, it's just that according to the README, and can't see anything else in the code either.

PR #428, opened 2021-05-17
Update postcss plugin to be compatible with postcss 8

micgro42:
This makes our postcss plugin compatible with postcss 8 based on what is apparently the official migration guide: https://evilmartians.com/chronicles/postcss-8-plugin-migration

This became an issue due to the automatic dependabot pull request for postcss 8 in #427.

However, this currently doesn't work since our dependency of postcss is not what is actually run when building Wikit. What is actually run is the postcss as specified by vue-cli, which still depends on postcss 7. The plugin as written in this PR is not compatible with postcss 7.

That being said, the vulnerability due to which dependabot created this PR #427 does't affect us. As stated in https://github.com/postcss/postcss/commit/8682b1e4e328432ba692bed52326e84439cec9e4#commitcomment-49809613:

> [...] it affects only use cases when:
>
>     1. User send you CSS
>
>     2. You compile this CSS on your servers
>
>
> In this case, the user can generate special CSS which will take seconds or minutes to compile. An attacker can use it to DoS your servers.
>
> If you can’t update PostCSS, you can add timeout for CSS processing.

Though, this patch can, and maybe should, be applied when we upgrade vue-cli or otherwise are able to actually make use of postcss v8 as it may represent a performance improvement.

### DIFF
--- a/vue-components/build/postcssWikitScope.js
+++ b/vue-components/build/postcssWikitScope.js
@@ -1,4 +1,3 @@
-const postcss = require( 'postcss' );
 const selectorParser = require( 'postcss-selector-parser' );
 
 // copied & modified from https://github.com/pazams/postcss-scopify/blob/master/index.js#L86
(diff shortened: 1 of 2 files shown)


micgro42:
Directly closing this again as it is only here for future reference.

PR #429, opened 2021-05-18
Disabled components should display disabled labels

Ladsgroup:
Note that "disabled" attribute doesn't work on <label> so it needs
class.

Bug: T268842

### DIFF
--- a/vue-components/src/styles/mixins/Label.scss
+++ b/vue-components/src/styles/mixins/Label.scss
@@ -8,6 +8,10 @@
   overflow-wrap: break-word;
   hyphens: auto;
 
+  &--disabled {
+    color: $wikit-Label-font-color-disabled;
+  }
+
   @if $displayType == block {
     padding-block-end: $wikit-Label-padding-block-end;
   }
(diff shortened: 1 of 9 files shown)


vue-components/src/styles/mixins/Label.scss
```diff
@@ -8,6 +8,10 @@
   overflow-wrap: break-word;
   hyphens: auto;
 
+  &--disabled {
```
itamargiv, line 11: mixins can accept parameters, as seen in line 1, so IMO it would be easier to add the logic regarding the label color into the mixin itself, rather than in each and every component.

itamargiv (changes requested):

vue-components/src/styles/mixins/Label.scss
Ladsgroup: It does accept parameter but not dynamically so it can't change it on the fly (in case a label gets disabled by a user interaction). Right?

itamargiv: hmm, that is very correct, I didn't think of that use case for some reason :sweat_smile:

itamargiv (approved):

Ladsgroup:
I'm the laziest person you can imagine, trust me I'd make a smaller patch if I could :P

PR #431, opened 2021-05-21
Fix docs about path to variables.css

1-Byte:
The underscores are not needed when one wants to include *variables.css*.


tzhelyazkova (approved):
Thanks! 

Edit: I can't merge this until we figure out how we want to run GitHub actions on PRs opened from a fork.

PR #434, opened 2021-06-07
Remove unnecessary outline of dropdowns in Safari

SaiSan-WMDE:
Issue: [T275052](https://phabricator.wikimedia.org/T275052)

### DIFF
--- a/vue-components/src/components/Dropdown.vue
+++ b/vue-components/src/components/Dropdown.vue
@@ -227,7 +227,7 @@ $base: '.wikit-Dropdown';
 			border-color: $wikit-Dropdown-active-border-color;
 			background-color: $wikit-Dropdown-active-background-color;
 			box-shadow: $wikit-Dropdown-active-box-shadow;
-			outline: none;
+			outline: none; // this removes an unwanted outline added by Safari on focus
 		}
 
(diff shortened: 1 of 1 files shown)


vue-components/src/components/Dropdown.vue
```diff
@@ -227,6 +227,7 @@ $base: '.wikit-Dropdown';
 			border-color: $wikit-Dropdown-active-border-color;
 			background-color: $wikit-Dropdown-active-background-color;
 			box-shadow: $wikit-Dropdown-active-box-shadow;
+			outline: none;
```
tzhelyazkova, line 230: We should probably comment inline that this affects only Safari's behavior.
I can see how someone can remove it when they see it doesn't do anything on Chrome/Firefox.

tzhelyazkova (approved):
Tested on Safari 14.1 and it works 💯

vue-components/src/components/Dropdown.vue
SaiSan-WMDE, line 230:
```suggestion
			outline: none; // this removes an unwanted outline added by Safari on focus
```

tzhelyazkova (approved):

PR #436, opened 2021-06-14
Check for overflow in Popover

lucaswerkmeister:
See the individual commit messages for details.

Bug: [T274261](https://phabricator.wikimedia.org/T274261)

### DIFF
--- a/vue-components/src/components/Popover.vue
+++ b/vue-components/src/components/Popover.vue
@@ -1,7 +1,7 @@
 <template>
 	<div
 		class="wikit wikit-Popover"
-		:class="[ `wikit-Popover`, `wikit-Popover--${position}` ]"
+		:class="[ `wikit-Popover--${adjustedPosition || position}`, { 'wikit-Popover--flush-edges': flushEdges } ]"
 		@mouseenter="startHover"
 		@mouseleave="endHover"
 		v-detect-click-outside="clickOutsideHandler"
@@ -10,7 +10,7 @@
 			<!-- @slot Target should always be a button, as we will listen to its click and hover events -->
 			<slot name="target" />
 		</span>
-		<div class="wikit-Popover__content-wrapper" v-if="isContentShown">
+		<div class="wikit-Popover__content-wrapper" v-if="isContentShown" ref="wrapper">
 			<div class="wikit-Popover__pointer" />
 			<div class="wikit-Popover__content">
 				<!-- @slot The content of the Popover goes into the default slot. -->
@@ -32,8 +32,11 @@ export default Vue.extend( {
 	data() {
 		return {
 			isContentShown: false,
+			adjustedPosition: null as string | null,
+			flushEdges: false,
 			showContentTimeoutID: null as number | null,
 			hideContentTimeoutID: null as number | null,
+			windowResizeEventHandler: null as ( () => void ) | null,
 		};
 	},
 	directives: {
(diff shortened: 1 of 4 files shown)


lucaswerkmeister:
Oops, I probably shouldn’t have deleted that GitHub Actions comment. Here’s the netlify link:

https://55fd8dc219a10a237dd1c583f5f427bf190a2d05--wmde-wikit.netlify.app/

lucaswerkmeister:
TODOs include:

- check overflow on more events – window resize, content change, position change
- …tests? idk

Possible further changes:
- a computed property for `adjustedPosition || position` might be useful (`effectivePosition`?)
- rewrite `checkOverflow()` to async/await syntax?

micgro42:
The changing of the position seems fine, but I'm having trouble getting the shrinking to work in Storybook?

vue-components/src/components/Popover.vue
```diff
@@ -264,6 +440,19 @@ $pointer-edge-length: math.hypot($wikit-Popover-pointer-width/2, $wikit-Popover-
 			}
 		}
 
+		&#{$base}--flush-edges #{$base}__content-wrapper {
+
+			@if $subposition == start {
+				inset-inline-end: min(50% - #{$wikit-Popover-pointer-width} / 2 -
+				#{$wikit-Popover-pointer-margin-horizontal}, 0px);
```
micgro42, line 447: I'm still struggling to fully wrap my head around this. Why are we doing this calculation, instead of just setting it 0px directly?

lucaswerkmeister: You can’t see it in the storybook (which is maybe a shortcoming that should be addressed anyways), but on very narrow buttons (such as icon buttons), this length becomes negative, and that’s necessary to ensure that the pointer is centered over the button.

With original `calc`:
Screenshot 2021-06-16 at 13-08-48  wmde query-builder
With `0`:
Screenshot 2021-06-16 at 13-08-14  wmde query-builder

vue-components/src/components/Popover.vue
```diff
@@ -264,6 +440,19 @@ $pointer-edge-length: math.hypot($wikit-Popover-pointer-width/2, $wikit-Popover-
 			}
 		}
 
+		&#{$base}--flush-edges #{$base}__content-wrapper {
+
+			@if $subposition == start {
+				inset-inline-end: min(50% - #{$wikit-Popover-pointer-width} / 2 -
```
lucaswerkmeister, line 446: TODO: Amend the commit message to note that

1. `min()` isn’t supported in IE11, but
2. that’s acceptable (I hope) because that just means it falls back to the non-flush-edges `inset-inline-end`, effectively disabling `avoidOverflowByMakingEdgesFlush()` (so we continue with `avoidOverflowBySettingMaxWidth()`).

vue-components/src/components/Popover.vue
micgro42: ah, gotcha, thanks

vue-components/src/components/Popover.vue
```diff
@@ -112,14 +129,173 @@ export default Vue.extend( {
...
+		 *
+		 * By default, for the non-center horizontal subpositions,
+		 * both the pointer and the edge of the popover are close to the center of the button.
+		 * We can gain some space by moving the edge of the popover to the edge of the button
+		 * (assuming that the button has some substantial width:
+		 * for small buttons, especially icon buttons, this makes no difference),
+		 * at the cost of also moving the pointer along with it.
+		 * The movement of the pointer away from the button’s center is unwanted,
+		 * but unavoidable given the current DOM structure
+		 * (where it’s positioned relative to the popover, not the button).
```
lucaswerkmeister, line 251: Demonstration of the difference from this function, including the unwanted pointer movement:
center
flush
The unwanted pointer movement is the reason that this adjustment is relatively late in the list of adjustments – we try jiggling the position first, and if that works, then the pointer can stay centered on the button.

micgro42 (approved):
This will make the tooltips in the query builder much nicer on mobile 🙌

PR #439, opened 2021-06-29
ci: run on pull requests

lucaswerkmeister:
Run CI on pull requests, including external ones. To avoid double CI runs, this means we no longer run CI on pushes, except to the master branch (after a pull request was merged). The deploy job is skipped on external pull requests, since the secrets won’t be available; this should cause the browser-tests job to be skipped as well.

Bug: [T283911](https://phabricator.wikimedia.org/T283911)

### DIFF
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -5,7 +5,11 @@
 #
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
 name: Test and Deploy
 jobs:
 
@@ -108,6 +112,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     needs: setup
+    if: ${{ github.repository_owner == 'wikit' }}
     outputs:
       deploy_url: ${{ steps.netlify_deploy.outputs.url }}
     steps:


Ladsgroup (approved):

.github/workflows/test-and-demo.yaml
```diff
@@ -108,6 +112,7 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     needs: setup
+    if: ${{ github.repository_owner == 'wikit' }}
```
micgro42, line 115: I think this broke our CI. It should have run in #443, but it didn't.

lucaswerkmeister: Wait, shouldn’t this be `wmde`? `wikit` is the repository _name_, not _owner.

lucaswerkmeister: Let’s see if #444 fixes it.

.github/workflows/test-and-demo.yaml
```diff
@@ -5,7 +5,11 @@
 #
 # For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
 
-on: push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
micgro42, line 12: I'm not entirely sure yet, but I suspect this has also somehow effectively broken the Netlify link that is commented into the PRs. I suspect that now `github.ref` in `--alias $(echo '${{ github.ref }}' | shasum | awk '{print $1}')` in the deploy step below refers no longer to the branch name but to something like `refs/pull/439/merge`.

I haven't looked closer into how to fix this yet. Might do so tomorrow.

PR #442, opened 2021-07-01
Use node12

Ladsgroup:
Bug: T285670


Ladsgroup:
Not 100% sure it works in all aspects. Double checking tomorrow.

micgro42:
You selected Node12, i.e. Maintenance LTS, would it also work with Node14, i.e. Active LTS?

I think using the nodejs docker images from DockerHub would be fine. Those releng images aren't meant for dev environments anyway.

micgro42:
After talking about this a bit more in the daily today:

As long as the (major) node version in the developer image and CI is consistent, we should be fine to increase this to 14. Applications that use this library via npm will always use the already compiled code, not the dev dependencies of Wikit.

That being said, we will also want to update the dev images of Query Builder, Wikidata Bridge and others. But those should probably be consistent with is used in their CI and on WMF Jenkins that is Node.js v12.

That being said, we _can_ do that in two steps/commits and merge this PR from v10 to v12 now and open another one for making the step from v12 to v14.

micgro42 (dismissed):
If we end up sticking to this way of providing Node.js v12, then I'd recommend renaming the service in docker-compose.yml from `node` to `npm`, because that will let us run commands like `docker-compose run --rm --user=$(id -u):$(id -g) npm audit` which is much less confusing than the messing with entry point in this patch.

But that should probably be a follow-up. Because if we switch to Node.js v14 after all, then we do not want that.

Ladsgroup:
> If we end up sticking to this way of providing Node.js v12, then I'd recommend renaming the service in docker-compose.yml from `node` to `npm`, because that will let us run commands like `docker-compose run --rm --user=$(id -u):$(id -g) npm audit` which is much less confusing than the messing with entry point in this patch.
> 
> But that should probably be a follow-up. Because if we switch to Node.js v14 after all, then we do not want that.

All of these gymnastics were needed because the releng docker file was different than the production one, I just reverted to use the dockerhub image and the changes are pretty minimal here.

micgro42:
> > If we end up sticking to this way of providing Node.js v12, then I'd recommend renaming the service in docker-compose.yml from `node` to `npm`, because that will let us run commands like `docker-compose run --rm --user=$(id -u):$(id -g) npm audit` which is much less confusing than the messing with entry point in this patch.
> > But that should probably be a follow-up. Because if we switch to Node.js v14 after all, then we do not want that.
> 
> All of these gymnastics were needed because the releng docker file was different than the production one, I just reverted to use the dockerhub image and the changes are pretty minimal here.

Looks good, but now the `package-lock.json` needs to be regenerated, because this Node.js v12 image contains npm 6 again.

Ladsgroup:
> Looks good, but now the `package-lock.json` needs to be regenerated, because this Node.js v12 image contains npm 6 again.

Thanks. Done now.

micgro42 (approved):

PR #443, opened 2021-07-07
Use node14

Ladsgroup:


micgro42:
Could you _touch_ a file in the vue-components repo for this PR? Because currently half the CI is not running, and it would probably be really useful to see them running and succeeding with Node.js v14 :)

Ladsgroup:
> Could you _touch_ a file in the vue-components repo for this PR? Because currently, half the CI is not running, and it would probably be useful to see them running and succeeding with Node.js v14 :)

git didn't let me push that file so I made a change to README file, did it fix it?

micgro42:
> > Could you _touch_ a file in the vue-components repo for this PR? Because currently, half the CI is not running, and it would probably be useful to see them running and succeeding with Node.js v14 :)
> 
> git didn't let me push that file so I made a change to README file, did it fix it?

It should have, but it seems the `deploy` step didn't work, probably related to #439? Even though the check added there _looks_ correct.

lucaswerkmeister:
#444 hopefully fixes CI.

lucaswerkmeister:
Rebased onto #444, let’s see what CI looks like now.

Ladsgroup:
It passes now.

docker-compose.yml
```diff
@@ -3,7 +3,7 @@ version: '2'
 services:
     node:
         # Once Wikimedia has node12 in production docker-registry, use that.
```
lucaswerkmeister, line 5: This comment should probably be `node14` now?

lucaswerkmeister:
Could this cause any issues when using Wikit in Wikimedia? E.g. is it possible that we start using node14-only constructs in Wikit, and then building downstream projects (like Data Bridge) one Node 12 wouldn’t work when they use Wikit?

Ladsgroup:
> Could this cause any issues when using Wikit in Wikimedia? E.g. is it possible that we start using node14-only constructs in Wikit, and then building downstream projects (like Data Bridge) one Node 12 wouldn’t work when they use Wikit?

I can't give you a good answer on that TBH but wikit directly won't be used in production and where it's used on projects, it will be an npm dependency so tests would fail there. And node version is much looser dependency than for example ES version.

The docker-compose file here is only for development and building a dev server to storybooks and such, nothing to core of wikit.

micgro42:
> Could this cause any issues when using Wikit in Wikimedia? E.g. is it possible that we start using node14-only constructs in Wikit, and then building downstream projects (like Data Bridge) one Node 12 wouldn’t work when they use Wikit?

I don't see how. Wikit is not a node application. When we make a release, we compile the components into es5 and that es5 code is what we install with npm, e.g. in Query Builder.

But if I'm missing something or maybe talking complete rubbish, @tzhelyazkova would know and be able to correct me. :)

docker-compose.yml
lucaswerkmeister: Okay, so maybe we should just remove the comment? If it’s just a development image, it shouldn’t be so important which Node image we use, right?

tzhelyazkova:
> > Could this cause any issues when using Wikit in Wikimedia? E.g. is it possible that we start using node14-only constructs in Wikit, and then building downstream projects (like Data Bridge) one Node 12 wouldn’t work when they use Wikit?
> 
> I don't see how. Wikit is not a node application. When we make a release, we compile the components into es5 and that es5 code is what we install with npm, e.g. in Query Builder.
> 
> But if I'm missing something or maybe talking complete rubbish, @tzhelyazkova would know and be able to correct me. :)

I think that's about right. Node is needed for the build and development tools, it doesn't affect the content of WiKit's `vue-components` and `tokens` packages.

Ladsgroup:
I removed the note from the docker-compose

lucaswerkmeister (approved):

tzhelyazkova (approved):

PR #444, opened 2021-07-08
Fix repository owner comparison in CI

lucaswerkmeister:
The repository owner is wmde, not wikit.


micgro42 (approved):
the URL commented into the PR is still wrong. But that seems to be a different problem. At least the deploy step runs again :+1:

PR #281, opened 2020-11-18
uptade wikimedia-ui-base dependency to v0.18

SaiSan-WMDE:
A couple of new tokens were introduced in order to align our styles with wikimedia-ui-base. No values were modified in this update.

Issue: https://phabricator.wikimedia.org/T268032

PR #445, opened 2021-07-09
Update wikimedia base UI

zonzon1804:
Update wikimedia base UI from 0.18.0 to 0.19.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ development)_
 - Initial version of the Lookup component
 - Additional icons to the Icon component
 - Additional sizes to the Icon component
-- New tokens based on wikimedia-ui-base v 0.18.0
+- New tokens based on wikimedia-ui-base v 0.19.0
 - New Dropdown Component
 - Quiet button variant
 - Icon only button variant
--- a/tokens/package-lock.json
+++ b/tokens/package-lock.json
@@ -7161,9 +7161,9 @@
 			"dev": true
 		},
 		"wikimedia-ui-base": {
-			"version": "0.18.0",
-			"resolved": "https://registry.npmjs.org/wikimedia-ui-base/-/wikimedia-ui-base-0.18.0.tgz",
-			"integrity": "sha512-K4Ym0lQpxE7dATeSbd+15skfsB7wmzDkR9F+iEKZkodnB8Cb3l2D7kurEjsiXFpqwnwvXC/5qs6ZBlWY2Jj4hA==",
+			"version": "0.19.0",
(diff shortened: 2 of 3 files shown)


tzhelyazkova (changes requested):
The package `wikimedia-ui-base` provides most of the values for WiKit's design tokens. Therefore we depend on it only for the `tokens` package.

For inspiration, here is the PR #281 where the last update of that package happened.
For more info about the different packages within WiKit see https://wmde.github.io/wikit/?path=/story/documentation-decisions-and-adrs-overview--page#packages
For more info on how this package relates to our design tokens see this ADR https://wmde.github.io/wikit/?path=/story/documentation-decisions-and-adrs-adrs-3-design-token-agnosticism-and-tooling--page#3-design-token-agnosticism-and-tooling
For more info on how we manage the different packages within this repo, see this ADR https://wmde.github.io/wikit/?path=/story/documentation-decisions-and-adrs-adrs-4-use-a-monorepo--page

lucaswerkmeister:
The latest version is still adding the package to the root `package.json`, as far as I can tell, which probably shouldn’t happen. (Also, there’s a strange unrelated diff in the `vue-components` tests?)

zonzon1804:
saw that... didn't notice. Right away

lucaswerkmeister (changes requested):
It still looks like the change does more than it’s supposed to:

- `docs/package-lock.json` has changes that look unrelated
- `package-lock.json` still has wikimedia-ui-base
- `tokens/package-lock.json` updates wikimedia-ui-base but also has other changes that look unrelated
- `vue-components/package-lock.json` has a huge diff that GitHub won’t even render, but that I assume must also be unrelated?
- `vue-components/tests/e2e/specks/Lookup.test.js` has changes that also look unrelated

Ladsgroup:
Maybe I'm wrong but if you don't use the right npm version, it'll mess up the package lock files. My suggestion is to use the docker commands mentioned in the README. I'm sorry if you followed it and this is not the problem.

zonzon1804:
it could be, npm is stuck with 6.14.4. Lemme check it out

lucaswerkmeister (approved):

zonzon1804:
Anyone knows why the browser-tests is skipped? Would it be just the beauty of running stuff on a cloud computer or something wrong with the code?

tzhelyazkova:
> Anyone knows why the browser-tests is skipped? Would it be just the beauty of running stuff on a cloud computer or something wrong with the code?

The browser tests are configured to run only on changes in the package `vue-components` since it doesn't make sense to run for `tokens` and `docs` as there are no components there.

tzhelyazkova (approved):

PR #448, opened 2021-07-14
Release all packages of WiKit under v2.0.0

dati18:
Release `docs`, `tokens`, `vue-components` under Wikit 2.0.0


package.json
```diff
@@ -21,7 +21,7 @@
     "publish-tokens-vue": "lerna publish from-package --yes"
   },
   "devDependencies": {
-    "lerna": "^3.22.1",
+    "lerna": "^4.0.0",
```
lucaswerkmeister, line 24: I assume this is the reason for the 14k lines of changes in the `package-lock.json`; is it really necessary?

zonzon1804: I got suggested by npm to update lerna... otherwise it shows me 28 vulnerabilities. Should I?

lucaswerkmeister: Let’s at least do it in a separate commit.

Ladsgroup: I think we should update lerna first (in separate commit) before releasing 2.0.0

Ladsgroup: mostly to reduce npm vulnerabilities and surface attacks (supply chain attack)

lerna.json
```diff
@@ -4,5 +4,5 @@
     "vue-components",
     "docs"
   ],
-  "version": "2.0.0-alpha.12"
+  "version": "2.0.0"
```
lucaswerkmeister, line 7: This looks like it belongs in the next commit.

zonzon1804:
What happened to Storybook? It passed the ui-review test yesterday... I haven't changed anything dramatic...

package-lock.json
```diff
@@ -1,1253 +1,9219 @@
 {
   "name": "wikit",
+  "lockfileVersion": 2,
```
micgro42, line 3: Sorry for being nit-picky here, but it seems you ran `npm install` here not with docker, but with some local npm? Please regenerate the package-lock.json for this commit with `docker-compose run --rm --user=$(id -u):$(id -g) node npm install --package-lock-only`

micgro42: The reasoning behind my request is that this would otherwise cause local changes to the `package-lock.json` whenever runs `npm install` with the docker command locally. That isn't a great developer experience.

zonzon1804: I think you got a point... I did run `npm install` locally. Lemme fix that

micgro42:
> What happened to Storybook? It passed the ui-review test yesterday... I haven't changed anything dramatic...

I think it is that the border radius was changed ever so slightly in #446 and chromatic isn't good at understanding git and wrongly thinks that this is related to this pull request. I.e. this is a false positive.

tzhelyazkova (commented):
I think this PR should also amend the `CHANGELOG.md` file by renaming `Unreleased` to `2.0.0  - <date of release>` and introducing a new `Unreleased` empty section above.

package.json
lucaswerkmeister: Eh, I forgot that this repo only allows squash-merge, so requiring separate commits is kinda pointless…

CHANGELOG.md
```diff
@@ -9,9 +9,9 @@ _**Important:** This file contains a draft to the official release notes, and is
 not a stable documentation of the upcoming version (as it will frequently change during
 development)_
 
-## [Unreleased]
+## [2.0.0 - <16.07.2021>]
```
lucaswerkmeister, line 12: The date is outdated now, unfortunately.

Also, I think there should still be an “unreleased” section, and the “expected next version” doesn’t belong below the 2.0.0 heading in my opinion – I would prefer adding the 2.0.0 heading between that line and the “added” heading. (But maybe Tonina disagrees?)

dati18: Yep, I tried to release it last week but it was kinda late

dati18: @lucaswerkmeister ok I've done it, please review one more time

CHANGELOG.md
```diff
@@ -11,7 +11,7 @@ development)_
 
 ## [Unreleased]
 
-**Expected next version:** 2.0.0
+## [2.0.0] - 20-07.2021
```
lucaswerkmeister, line 14: Ok but now the date is a bit strange…

- format should be either 20.07.2021 or 2021-07-20 IMHO (I’d prefer 2021-07-20)
- today is the 19th – do you want to wait with the release until tomorrow?

lucaswerkmeister (approved):

lucaswerkmeister:
I’ve noticed this too late, but `vue-components/package.json` still has this line which should probably have been updated:

```json
    "@wmde/wikit-tokens": "^2.0.0-alpha.12",
```

Hopefully we can fix that for the next release.

PR #449, opened 2021-07-20
Remove the Button browser test

tzhelyazkova:
This test only checks if the Button renders and it was added as an example of how to write browser tests with Nightwatch.

Now there are browser tests for the Lookup component, therefore this is no longer needed.

Bug: T268765


jakobw (approved):
:hammer:

PR #451, opened 2021-08-03
Update "Developing Components" doc

tzhelyazkova:
We removed the example button browser test because now there are
more browser tests written for other components, therefore we don't
need an example one anymore.


SaiSan-WMDE (approved):
Thanks for the update! 😁

PR #455, opened 2021-08-13
Convert hard coded calculation into a token value

itamargiv:
This change converts a hard-coded calculation value in  the tokens source into a token value, as the hard-coded calculation was generating code that was both unusable in the CSS variable sheets (CSS does not understand `/` without `calc`) and was causing the sass compiler to emit deprecation warnings.

Bug: [T288809](https://phabricator.wikimedia.org/T288809)

### DIFF
--- a/tokens/properties/mixins/BouncingDots.json
+++ b/tokens/properties/mixins/BouncingDots.json
@@ -7,12 +7,12 @@
             "value": "{border.radius.circle.value}"
         },
         "margin-left": {
-            "value": "{dimension.dynamic.size.100.value}/4",
-            "comment": "not sure the operation is going to work"
+            "value": "{dimension.dynamic.size.25.value}",
+            "comment": "Encoding this value as 100/4 = 25"
         },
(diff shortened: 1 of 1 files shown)


tokens/properties/mixins/BouncingDots.json
```diff
@@ -7,12 +7,12 @@
             "value": "{border.radius.circle.value}"
         },
         "margin-left": {
-            "value": "{dimension.dynamic.size.100.value}/4",
-            "comment": "not sure the operation is going to work"
+            "value": "{dimension.dynamic.size.25.value}",
```
jakobw, line 10: I was pondering whether it makes more sense to use a hardcoded value here instead of a token just in case the ration between these two token changes, but it seems safe in this case.

jakobw (approved):
Math checks out :abacus: :nerd_face:

itamargiv:
> Math checks out abacus nerd_face

Thank you Sire :pray:

PR #458, opened 2021-08-20
Add inline progress bar tokens

SaiSan-WMDE:
This makes available a first draft of the tokens needed to style the inline progress bars (both determinate and indeterminate). The assumption (TBD) is that this inline loader will be implemented as a mixin with two types. 
This inline bar is displayed within certain components (e.g. text area, dialog, table) to indicate their loading state to users.

Issue: [T289058](https://phabricator.wikimedia.org/T289058)

See [this Codepen ](https://codepen.io/saraisanchez/pen/abJeNrm)to get a feel of the determinate and indeterminate inline loading bar animation behaviors.


tokens/properties/mixins/InlineProgressBar.json
```diff
@@ -0,0 +1,39 @@
...
+        },
+        "height": {
+            "value": "{dimension.static.size.50.value}"
+        },
+        "animation": {
+            "duration": {
+                "value": "{animation.loader.duration.slow.value}"
+            }
+        },
+        "track": {
```
SaiSan-WMDE, line 14: The "track" is the wrapper that contains both inline progress bar typologies (determinate and indeterminate). The width of the track is 100% by default, and therefore is ultimately determined by its own contextual wrapper. 
Note: It might make more sense to let the progress bars inherit the height of the track instead of creating two tokens with the same value. Suggestions welcome.

tokens/properties/mixins/InlineProgressBar.json
```diff
@@ -0,0 +1,39 @@
+{
+    "wikit-InlineProgressBar": {
+        "background-color": {
+            "value": "{background.color.progressive.default.value}"
+        },
+        "height": {
+            "value": "{dimension.static.size.50.value}"
+        },
+        "animation": {
```
SaiSan-WMDE, line 9: I don't think we need to create tokens for the animations' timing-functions. After all, if WiKit were to continue, it would be ideal to get rid of the tokens that document CSS values, and use mixins  (with duration tokens) to apply transitions and animations instead.

itamargiv (approved):
NICE! thank you!

PR #460, opened 2021-08-24
Add textarea story

itamargiv:
_**Note: This PR depends on #459**_

This change adds documentation for the initial text area component, and a story to display it in WiKit's storybook. Additionally, it adds the `value` props to the TextArea component, which was missed (yet unneeded) by the Mismatch Finder implementation.

Bug: [T289141](https://phabricator.wikimedia.org/T289141)

### DIFF
--- a/vue-components/src/components/TextArea.vue
+++ b/vue-components/src/components/TextArea.vue
@@ -15,6 +15,7 @@
 				'wikit-TextArea__textarea',
 				`wikit-TextArea__textarea--${resizeType}`
 			]"
+			:value="value"
 			:rows="rows"
 			:placeholder="placeholder"
 			label=""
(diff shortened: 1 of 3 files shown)


vue-components/src/components/TextArea.vue
```diff
@@ -11,10 +11,12 @@
 			</label>
 		</span>
 		<textarea
+			:id="id"
 			:class="[
 				'wikit-TextArea__textarea',
 				`wikit-TextArea__textarea--${resizeType}`
 			]"
+			:value="value"
```
Silvan-WMDE, line 19: Is this textarea value (as well as the corresponding test) actually a change that should go into #459? Or is it somehow related to the story?

Silvan-WMDE (commented):
just a question, while this is still a draft

vue-components/src/components/TextArea.vue
itamargiv: I don't think it really relates to any of them, I only noticed it since I was looking at the parent story while working where it is mentioned but not mentioned by any of the sub tickets we had to implement. I can separate this to a followup smaller PR though if you think that's better.

Silvan-WMDE: no, if it doesn't obviously belong in #459, I think you can just as well leave it in here.

PR #456, opened 2021-08-20
Create textare.json and missing tokens

SaiSan-WMDE:

PR #464, opened 2021-09-02
TextArea: Add read only and loading states

itamargiv:
_**Important:** This PR cannot be merged until both #456 and #458 are merged as well. Once they are on the main branch, this PR must be rebased_


itamargiv:
Abandoned for split branches

PR #467, opened 2021-09-02
Add body-component line-height token

SaiSan-WMDE:
This removes the label style (which is not part of the Design Style Guide) and adds a token to reduce the line-height of "body", so this base style can be used as the default by all components without altering their heights.

This is prep for the creation of the table component tokens. 
Issue: [T289053](https://phabricator.wikimedia.org/T289053)


vue-components/src/components/DateInput.vue
```diff
@@ -92,19 +92,19 @@ export default defineComponent( {
 .wikit-DateInput {
 	&__IntroText {
 		padding-block-end: $wikit-InputExtender-content-padding-vertical;
-		font-family: $font-family-style-label;
-		font-weight: $font-weight-style-label;
-		font-size: $font-size-style-label;
-		line-height: $font-line-height-style-label;
+		font-family: $font-family-style-body;
```
SaiSan-WMDE, line 95: Probably worth replacing by component tokens (new json file should be created) to prevent the same issue from happening in a hypothetical future.

tokens/properties/global.json
```diff
@@ -150,7 +150,7 @@
                 "value": "none"
             },
             "small": {
-                "value": "{wmf-line-height-heading.value}"
+                "value": "1.3"
```
SaiSan-WMDE, line 153:
```suggestion
                "value": "1.25"
```

vue-components/src/components/DateInput.vue
SaiSan-WMDE: done in [23c5fe6](https://github.com/wmde/wikit/pull/467/commits/23c5fe64ee7d183a8f1627c636404fecc0a3e059)

tzhelyazkova (approved):
Yay for `DateInput` tokens 💯

SaiSan-WMDE:
> Yay for `DateInput` tokens 💯

Thanks! Really wishing we could use mixins there instead, though 😅

PR #468, opened 2021-09-02
Use rem as the unit of layout tokens

SaiSan-WMDE:
Using values in _em_  with layout tokens complicates having control over the vertical rhythm of pages: the selected token will be adjusted in context, potentially contradicting what was intended/stated by the design specifications (e.g. our intended 16px (1em) spacing, will translate to 32px if the token is applied to size a margin below a h1, or to 18px if applied to space a h4).

This change implies having to do some adjustments to the vertical spacing applied to the Query builder. Issue: https://phabricator.wikimedia.org/T288931


tokens/properties/global.json
```diff
@@ -233,6 +233,23 @@
                     "value": "2.500em"
                 }
             },
+            "size-rem": {
```
SaiSan-WMDE, line 236: Should we better do:

A. 

`"dimension": {
        "dynamic": {
            "size-em": {
               [...]
                },
            "size-rem": {
               [...]
                },`

B.

`"dimension": {
        "dynamic": {
            "em": {
               [...]
               },
            "rem": {
               [...]
               },`

C.

`"dimension": {
        "dynamic": {
            "size": {
              "em": {
                 }
               },
            "rem": {
               [...]
                },`

jakobw: I vote for option B. Since they're in the "dimension" category they're all size related in some way so the extra "size" prefix doesn't tell us anything new IMO.

itamargiv (changes requested):
Looks good to me! The only comment I have is that we probably want to add to the change log the fact that the layout tokens are now rem base rather than em based, since those are used externally.

SaiSan-WMDE:
> Looks good to me! The only comment I have is that we probably want to add to the change log the fact that the layout tokens are now rem based rather than em based, since those are used externally.

Agh! Of course 😅  Thanks for the heads-up!  Adding that rn

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ development)_
 - Disabled components will display disabled labels to make it more clear to users
 - Dropdown will not be activated when toggling the tooltip via enter
 - Keyboard-focusable components display a focused error and warning states
+- Dimension-layout tokens are now based on _rem_ instead of _em_
```
tzhelyazkova, line 57: This should be in the unreleased section.

SaiSan-WMDE: Done. I didn't remember the structure of the file had change! Thanks (again), señorita 🙇‍♀️

itamargiv (approved):
Thank you!

tzhelyazkova (approved):

PR #471, opened 2021-09-03
Text area improvements

SaiSan-WMDE:
- Add disabled state of text area  ✅
- Allow `both` as resize value  ✅   
- `Rows` prop value should be >=2 ✅  (Tiny validator included. Unit test in #477)
- Prevent unlimited horizontal minimization ✅ (Since the control type was changed to "range", a high placeholder limit of 20 rows has been defined for now)
- Set min-height to prevent excessive vertical resizing ✅ 
- Correct line-height to make value text more readable ✅ 

Storybook link: https://5efdb3b5f65950002286285d-yrpuwrrkzh.chromatic.com/


vue-components/src/components/TextArea.vue
```diff
@@ -3,7 +3,8 @@
 		<span class="wikit-TextArea__label-wrapper">
 			<label
 				:class="[
-					'wikit-TextArea__label'
+					'wikit-TextArea__label',
+					disabled ? `wikit-TextInput__label--disabled` : ''
```
SaiSan-WMDE, line 7:
```suggestion
					disabled ? `wikit-TextArea__label--disabled` : ''
```

jakobw (dismissed):
This looks pretty great! :tada: 

> Rows prop value should be >=2

I see the change in storybook, but do you think this should always be enforced for the component? In that case we could add a prop validator checking the value.

SaiSan-WMDE:
> > Rows prop value should be >=2
> 
> I see the change in storybook, but do you think this should always be enforced for the component? In that case we could add a prop validator checking the value.

Thanks so much for taking a look, @jakobw!  I thought that indicating the value as a "min" for the `Rows`prop [here](https://github.com/wmde/wikit/pull/471/files#diff-a132c794d1cfbf1351c9fafbf3c97375b5ad1b387c34f416428740fef366835aR62) was enforcing enough. Would you recommend including some sort of validation?

vue-components/src/components/TextArea.vue
```diff
@@ -96,6 +98,9 @@ export default defineComponent( {
 		rows: {
 			type: Number,
 			default: 2,
+			validator( value: number ): boolean {
```
SaiSan-WMDE, line 101: Is this too silly? 😅

vue-components/src/components/TextArea.vue
```diff
@@ -96,6 +98,9 @@ export default defineComponent( {
...
+				return value >= 2;
```
SaiSan-WMDE, line 102: Is this too silly? 😅

jakobw: Nope, that looks good. I uploaded #477 with a small unit test for it so we don't accidentally break it in the future.

jakobw (dismissed):
Looks good to me! :tada:

jakobw (approved):
:shipit:

PR #479, opened 2021-09-09
Adjust indeterminate inline loader width to reduce its speed

SaiSan-WMDE:
Tiny change to reduce the speed of the inline indeterminate loader and get it closer to the [WMF live demo example.](https://di-searchland-2.web.app/)

Related but not a blocker to [T290151](https://phabricator.wikimedia.org/T290151).


itamargiv (approved):
Sweet! Thank you!

PR #481, opened 2021-09-09
TextArea fixes

SaiSan-WMDE:
Two missing fixes:

- The TextArea story file was missing a comma (for sure this is collateral damage from conflict solving), and tests kept complaining
- A calc needed to be introduced to adjust the min-height of the text area

### DIFF
--- a/vue-components/stories/TextArea.stories.ts
+++ b/vue-components/stories/TextArea.stories.ts
@@ -7,18 +7,18 @@ export default {
 };
 
 export function basic( args: object ): Component {
-	return {
-		data(): object {
-			return {
-				// Binding to the currentValue prevents the value from resetting
-				// on the re-render when changing the read-only control from
-				// false to true
-				currentValue: '',
-			};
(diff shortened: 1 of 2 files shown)


vue-components/stories/TextArea.stories.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
tzhelyazkova, line 1: why do we need to disable the max-len eslint rule for this file?

tzhelyazkova (commented):
Are the indentation changes made on purpose? Is it something your IDE did or something the `fix` npm script did?

vue-components/stories/TextArea.stories.ts
SaiSan-WMDE: the IDE made me do it. Let me rewind and try to undo this and the indentation.

tzhelyazkova (approved):

PR #484, opened 2021-09-20
Move basic Table Component from Mismatch Finder to Wikit

guergana:
Table component with basic states from Mismatch Finder. Task [T289843](https://phabricator.wikimedia.org/T289843)

Bug: [T291074](https://phabricator.wikimedia.org/T291074)

### DIFF
--- a/vue-components/src/components/Table.vue
+++ b/vue-components/src/components/Table.vue
@@ -1,0 +1,187 @@
+<template>
+	<table
+		:class="[
+			'wikit',
+			'wikit-Table',
+			`wikit-Table--linear-${breakpoint}`
+		]"
+	>
+		<slot />
+	</table>
+</template>
+
+<script lang="ts">
+import Vue, { PropType } from 'vue';
+import { Breakpoint, validateBreakpoint } from './Breakpoint';
+
+export default Vue.extend( {
+	name: 'Table',
+	props: {
+		linearize: {
+			type: String as PropType<Breakpoint>,
+			validator( value: Breakpoint ): boolean {
+				return validateBreakpoint( value );
+			},
+			default: Breakpoint.Tablet,
+		},
+	},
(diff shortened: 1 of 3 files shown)


vue-components/src/components/Table.vue
```diff
@@ -0,0 +1,187 @@
...
+		* lets us add text before the content to provide
+		* the missing context.
+		*
+		* Markup:
+		*   <td data-title="Column Header">Content Here</td>
+		*
+		* Display:
+		*   Column Header: Content Here
+		*/
+		/* stylelint-disable selector-no-qualifying-type  */
```
itamargiv, line 79: I think disabling this line from linting here is a good idea, since the motivation for this rule is usually to prevent more general style usages of qualifying types, that is - a tagname followed by a qualifier such as a class, id or other attribute selector.

However in this case we want to ensure that we are indeed only targeting `td`, and `th` with the `data-header` attribute (and not `trs` or any other element someone might put in that slot) so I think this usage is justified. Thank you!

vue-components/tests/unit/components/Table.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+
+		expect( wrapper.props().linearize ).toBe( Breakpoint.Desktop );
+		expect( wrapper.find( 'table' ).classes() ).toContain( 'wikit-Table--linear-desktop' );
+	} );
+
+	it( 'ignores invalid breakpoint values', () => {
+		expect( () => mount( Table, {
+			propsData: { linearize: 'nonsense' },
+		} ) ).toThrow( 'Invalid prop: custom validator check failed for prop "linearize"' );
+	} );
```
itamargiv, line 21: Well done :100:

itamargiv (approved):
Great work @guergana ! Thank you!

PR #485, opened 2021-09-21
Add a Build step to the setup job of the Test and Deploy workflow

chukarave:
The build step is invoked in case of a cache hit.
https://phabricator.wikimedia.org/T290426

### DIFF
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -52,6 +52,9 @@ jobs:
         path: |
           */dist
         key: ${{ runner.os }}-${{ hashFiles('*/dist') }}
+    - name: Build
+      if: steps.cache.outputs.cache-hit == 'true'
+      run: lerna bootstrap
     - name: Install
       if: steps.cache.outputs.cache-hit != 'true'
       run: npm ci
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -8,7 +8,7 @@
 on:
   push:
     branches:
-      - master
+      - T290426-test-workflow-to-include-tsc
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-demo.yaml
```diff
@@ -52,6 +52,9 @@ jobs:
         path: |
           */dist
         key: ${{ runner.os }}-${{ hashFiles('*/dist') }}
+    - name: Build
+      if: steps.cache.outputs.cache-hit == 'true'
```
toban, line 56: Why don't we want this to run every time?

lucaswerkmeister: I assume it’s because `lerna bootstrap` is already in the `postinstall` script in `package.json`, so it should run as part of the `npm ci` immediately below; we only need to run it separately if `npm ci` is skipped because `node_modules/` was restored from cache.

chukarave: Yes, this is the gist of it. Based on the info I got from @itamargiv we use the cache to begin with in order to avoid installing every time (to save time + resources). When we do end up installing, the build runs automatically afterwards which makes the separate Build step redundant.

toban: ok

lucaswerkmeister: Can we add a comment explaining this? Maybe something like ``# (normally runs as part of `npm ci`)``.

(To me this would also feel more natural if this block was after the Install block instead of before it. `npm ci` is the standard/straightforward case, and `lerna bootstrap` becomes necessary if that was skipped, due to the cache.)

lucaswerkmeister (approved):

PR #492, opened 2021-10-14
vue-components: Add slots for icons in Button.vue

tzhelyazkova:
There are now three slots in Button:
- default one for the label
- prefix one for an icon positioned before the label
- suffix one for an icon positioned after the label

Bug: T270327

### DIFF
--- a/vue-components/src/components/Button.vue
+++ b/vue-components/src/components/Button.vue
@@ -9,7 +9,21 @@
 		]"
 		:type="nativeType"
 	>
-		<slot />
+		<div class="wikit-Button--content">
+			<!-- @slot Use this slot to pass an icon to appear before the label-->
+			<slot name="prefix" />
+			<div
+				:class="[
+					hasPrefixSlot ? `wikit-Button--content--start` : '',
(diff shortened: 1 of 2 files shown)


tzhelyazkova:
Requesting a UX review from @sai-san first. The only fail in CI is Chromatic, due to the newly introduced button story
https://www.chromatic.com/build?appId=5efdb3b5f65950002286285d&number=2353

tzhelyazkova:
> This is awesome 100 100 100
> 
> Just a couple of comments regarding the story display:
> 
> I'd say that **we don't need to document the button with suffix at story level**, just the prefixed version is enough. The reason being that the common UX pattern is to include prefix icons to support the meaning of the button labels before these. We needed to include a suffix to make the `wikit-Button` component technically reusable to create a different component: button dropdowns, which communicate the presence of a dropdown menu via a down arrow suffix icon.
> 
> I'd also say that **including the button with prefix in each one of the typology stories** would allow for a more organized consultation. (Whether we should do the same about the `icon-Only` is material for a separate discussion thinking )

Alright, the stories have been shuffled https://www.chromatic.com/build?appId=5efdb3b5f65950002286285d&number=2355

PR #500, opened 2021-10-26
Fix wikit-Dialog tokens

sai-san:
A couple of fixes needed to be applied to the Dialog tokens file:

1. The value of `box-shadow-divider-bottom` was corrected, which means that `$wikit-Dialog-header-box-shadow` can now be applied to the Dialog's header (as intended) instead of to the body.

2. The missing token `$wikit-Dialog-border-color` was created: it should replace the alias `$border-color-base-subtle` being used to style the component at the moment.


itamargiv (approved):
Looking great! Will change in Mismatch Finder after the next pre-release

PR #497, opened 2021-10-21
Add 'arrownext' and 'arrowprevious' icons

sai-san:
This adds the missing 'arrowprevious' icon, to be used by a button instance in the Mismatch Finder UI. It also includes its pair 'arrownext'.

In addition to the missing icon components, we added a 'dir' prop and  the controls to modify it.

Issues: [T290864](https://phabricator.wikimedia.org/T290864)  and [T294844](https://phabricator.wikimedia.org/T294844)

PR #503, opened 2021-11-09
Show dir control only if icon is flippable

jakobw:
https://26ce474ceb014e3726daf62e2cdeec10225359db--wmde-wikit.netlify.app/?path=/docs/vue_icon--basic

### DIFF
--- a/vue-components/stories/Icon.stories.ts
+++ b/vue-components/stories/Icon.stories.ts
@@ -1,5 +1,5 @@
 import Icon from '@/components/Icon';
-import { iconSizes, IconDirection, iconColors, IconTypes } from '@/components/iconProps';
+import { iconSizes, IconDirection, flippable, IconTypes } from '@/components/iconProps';
 import { Component } from 'vue';
 
 export default {
@@ -8,7 +8,9 @@ export default {
 };
 
 const typeValues = Object.values( IconTypes );
-export function basic( args: object ): Component {
+export function basic( args: { dir: string; type: string }, { argTypes } ): Component {
+	argTypes.dir.control.disable = !flippable.includes( args.type );
+
 	return {
 		components: { Icon },
 		props: Object.keys( args ),


vue-components/stories/Icon.stories.ts
```diff
@@ -8,7 +8,9 @@ export default {
 };
 
 const typeValues = Object.values( IconTypes );
-export function basic( args: object ): Component {
+export function basic( args: { dir: string; type: string }, { argTypes } ): Component {
+	argTypes.dir.control.disable = !flippable.includes( args.type );
```
itamargiv, line 12: I think this might need to exist in the component's `data` or as a computed property in order for it to be responsive to changes in the canvas view. I had a similar problem in TextArea. See https://github.com/wmde/wikit/blob/master/vue-components/stories/TextArea.stories.ts#L11-L18

jakobw:
Whoops, sorry. I somehow assumed that this issue went away because #497 was merged. Reopening, and will rebase the branch.

vue-components/stories/Icon.stories.ts
jakobw: Thanks for the comment! Sadly I'm not sure I understand the suggestion. The component reacts fine to any changes coming from the controls. It's the other way around that seems to get lost somehow. Setting `argTypes.dir.control.disable` doesn't reach the controls. The line we're commenting on gets executed at the right time with the right values, it just doesn't have the same effect as it does in the "Docs" tab.

In the "Canvas" tab, the component sits on a separate page that's embedded through an iframe, in "Docs" they're on the same page. I sadly don't know storybook well enough to know if and how it's possible to make the two talk both ways in the former setup.

Feel free to tinker with this and build on top or try a different approach. I probably won't have time to look into this again this week, but would of course be happy to see it done :smile: 

Ideally this wouldn't require a control -> component story -> control roundtrip at all. This would be easy if one control could inform another. That's https://github.com/storybookjs/storybook/issues/11486 I guess.

itamargiv: Yeah, I guess I misunderstood how this works. Now that I tried it, I see it's not what I thought :facepalm:  

In any case, it might be too much of an overhead to try and shoehorn an unsupported feature into storybook for a component that is probably going to be migrated to codex (and with that into vitepress) probably sooner rather than later.

sai-san: Ok. Considering Storybook's limitations and the effort required to only potentially overcome them, I'd say we should close this PR after all? Displaying the desired behavior only in Docs might make the improvement less discoverable, and the inconsistency between both Controls panels could generate confusion.
Thanks, Jakob, for getting so close to a solution! 💯 And thanks Itamar for your support 🙌

jakobw: It has all the thumbs up, and I absolutely agree that this isn't worth the effort and/or potential for confusion. Thanks for looking into it though! :sparkles:

jakobw:
Closing per the last few comments!

PR #507, opened 2021-11-30
Add Story & Storybook Documentation for Dialog Component

guergana:
Bug: [T295875](https://phabricator.wikimedia.org/T295875)

### DIFF
--- a/vue-components/src/components/Dialog.vue
+++ b/vue-components/src/components/Dialog.vue
@@ -196,7 +196,8 @@ export default Vue.extend( {
 		},
 		dismiss(): void {
 			/** 
-			 * Closes / dismisses the dialog
+			 * Emitted whenever the dialog is dismissed by clicking the **X** button, 
+			 * clicking the page overlay or pressing the <kbd>Esc</kbd> key.
 			 */
 			this.$emit( 'dismissed' );
 			this.hide();
--- a/vue-components/src/components/Dialog.vue
+++ b/vue-components/src/components/Dialog.vue
@@ -179,7 +179,11 @@ export default Vue.extend( {
 			document.removeEventListener( 'keydown', this._handleKeydown );
 			this.open = false;
 			/**
-			 * is called whenever the visible state of the dialog changes
+			 * Emitted whenever the visible state of the dialog changes. Includes a boolean to represent 
+			 * the new state of the dialog as payload.
+			 * 
+			 * Payload:
+			 * `open` - `true` when the dialog is opened, `false` when it is closed 
 			 */
 			this.$emit( 'update:visible', this.open );
 		},
--- a/vue-components/src/components/Dialog.vue
+++ b/vue-components/src/components/Dialog.vue
@@ -203,8 +203,15 @@ export default Vue.extend( {
 		},
 		_dispatch( namespace: string ): void {
 			/**
-			 * generic action that will be performed by the primary 
-			 * or secondary buttons of the dialog component
+			 * Emitted when one of the Dialog's action buttons are pressed. Includes the action's `namespace` 
+			 * as defined in the `actions` prop and an instance of the dialog itself as payload.
+			 *
+			 * Payload:
+			 * `namespace` - A string, specifying the namespace of the action that was triggered, can be 
+			 *  used to differentiate between the primary and secondary actions of the Dialog.
+			 *
+			 * `dialog` - An instance of the dialog the action is triggered from. Can be used to close the dialog 
+			 * with `dialog.hide()` after executing code associated with the action.
 			 */
 			this.$emit( 'action', namespace, this );
 		},
--- a/vue-components/src/components/Dialog.vue
+++ b/vue-components/src/components/Dialog.vue
@@ -188,11 +188,11 @@ export default Vue.extend( {
 			document.removeEventListener( 'keydown', this._handleKeydown );
 			this.open = false;
 			/**
-			 * Emitted whenever the visible state of the dialog changes. Includes a boolean to represent 
+			 * Emitted whenever the visible state of the dialog changes. Includes a boolean to represent
 			 * the new state of the dialog as payload.
-			 * 
+			 *
 			 * Payload:
-			 * `open` - `true` when the dialog is opened, `false` when it is closed 
+			 * `open` - `true` when the dialog is opened, `false` when it is closed
 			 */
 			this.$emit( 'update:visible', this.open );
 		},
@@ -204,8 +204,8 @@ export default Vue.extend( {
 			this.$nextTick( this._resetScroll );
 		},
 		dismiss(): void {
-			/** 
-			 * Emitted whenever the dialog is dismissed by clicking the **X** button, 
+			/**
+			 * Emitted whenever the dialog is dismissed by clicking the **X** button,
 			 * clicking the page overlay or pressing the <kbd>Esc</kbd> key.
 			 */
 			this.$emit( 'dismissed' );
@@ -213,14 +213,14 @@ export default Vue.extend( {
 		},
 		_dispatch( namespace: string ): void {
 			/**
-			 * Emitted when one of the Dialog's action buttons are pressed. Includes the action's `namespace` 
+			 * Emitted when one of the Dialog's action buttons are pressed. Includes the action's `namespace`
 			 * as defined in the `actions` prop and an instance of the dialog itself as payload.
 			 *
 			 * Payload:
-			 * `namespace` - A string, specifying the namespace of the action that was triggered, can be 
+			 * `namespace` - A string, specifying the namespace of the action that was triggered, can be
 			 *  used to differentiate between the primary and secondary actions of the Dialog.
 			 *
-			 * `dialog` - An instance of the dialog the action is triggered from. Can be used to close the dialog 
+			 * `dialog` - An instance of the dialog the action is triggered from. Can be used to close the dialog
 			 * with `dialog.hide()` after executing code associated with the action.
 			 */
 			this.$emit( 'action', namespace, this );
--- a/vue-components/src/components/Dialog.vue
+++ b/vue-components/src/components/Dialog.vue
@@ -178,6 +178,9 @@ export default Vue.extend( {
 		hide(): void {
 			document.removeEventListener( 'keydown', this._handleKeydown );
 			this.open = false;
+			/**
+			 * is called whenever the visible state of the dialog changes
+			 */
 			this.$emit( 'update:visible', this.open );
 		},
 		show(): void {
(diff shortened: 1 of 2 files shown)


vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,125 @@
...
+					variant="primary" type="progressive"
+				>
+					Open Basic Dialog
+				</Button>
+				<Dialog
+					:title="title"
+					ref="simple"
+					:actions="actions"
+					style="transform: translate(50%, 50%)"
+					@action="(_, dialog) => dialog.hide()"
```
guergana, line 27: I took this from the MMF. should this be the generic action?

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+					:dismiss-button="dismissButton"
+					:visible="visible"
+				>
+					<p style="margin-bottom:10px">Hello, I am a simple dialog</p>
+					{{ dialogText }}
```
guergana, line 32: tbh i am not sure about adding this here. it seems storybook doesnt accept html so it doesn't render optimally, if we want paragraphs and a title we need to add a separate control for each of them?

vue-components/src/components/Dialog.vue
```diff
@@ -188,10 +191,17 @@ export default Vue.extend( {
...
 		},
 		dismiss(): void {
+			/** 
+			 * Closes / dismisses the dialog
+			 */
 			this.$emit( 'dismissed' );
 			this.hide();
 		},
 		_dispatch( namespace: string ): void {
+			/**
```
guergana, line 201: Prob needs a better explanation. I don't know how to write it because i am not sure I even understand how it works 100%.

vue-components/stories/Dialog.stories.ts
itamargiv: I am not sure if I understand the question? If you mean to say should the dialog just close on any button click then yes, it should

vue-components/src/components/Dialog.vue
```diff
@@ -188,10 +191,17 @@ export default Vue.extend( {
...
+			 * generic action that will be performed by the primary 
+			 * or secondary buttons of the dialog component
+			 */
```
itamargiv, line 204:
```suggestion
			/**
			 * Emitted when one of the Dialog's action buttons are pressed. Includes the action's `namespace` 
			 * as defined in the `actions` prop and an instance of the dialog itself as payload.
			 *
			 * Payload:
			 * `namespace` - A string, specifying the namespace of the action that was triggered, can be 
			 *  used to differentiate between the primary and secondary actions of the Dialog.
			 *
			 * `dialog` - An instance of the dialog the action is triggered from. Can be used to close the dialog 
			 * with `dialog.hide()` after executing code associated with the action.
			 */
```

vue-components/stories/Dialog.stories.ts
itamargiv: I am not sure about this either. I definitely don't think we should add more than one control for the body of the dialog, but I'm not sure how valuable this is if we can't actually change the HTML. @sai-san what do you think?

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+    title: {
+        control: {
+            type: 'text',
+        },
+    },
+	actions: {
+        control: {
+            type: 'object',
+        },
+    },
```
itamargiv, line 72: Actually, the type for this is array, but it is currently documented as `DialogAction[]` which is not necessarily understandable to implementations outside WiKit. So we can override this by adding a type summary description:
```suggestion
    actions: {
        control: {
            type: 'object',
        },
        table: {
            type: {
                summary: 'array',
            },
        },
    },
```

itamargiv (changes requested):
Looks good altogether, there are still missing descriptions for each of the props in the controls and docs tab of the story. We can add those as docbloc comments above the definition of each property in the component's vue file.

Additionally, we have an issue with the overly not showing up. Probably due to the changes @Silvan-WMDE and I made by converting the `top: 0, left: 0, right: 0, bottom: 0` css positions of the overlay in the vue file to the `inset` shorthand (which might not work as expected).

vue-components/stories/Dialog.stories.ts
sai-san: I agree that the controls are quite limited. Let's just include a single editable paragraph in the body, that's fine.

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+	return {
+		components: { Button, Dialog },
+		props: Object.keys( args ),
+		template: `
+			<div>
+				<Button 
+					@click.native="$refs.simple.show()" 
+					variant="primary" type="progressive"
+				>
+					Open Basic Dialog
```
sai-san, line 20:
```suggestion
					  Open complex dialog
```

sai-san, line 20:
```suggestion
					Open complex dialog
```

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+					<p style="margin-bottom:10px">Hello, I am a simple dialog</p>
+					{{ dialogText }}
+				</Dialog>
+			</div>
+		`,
+	};
+}
+
+complex.args = {
+	title: 'I am a basic dialog',
```
sai-san, line 40:
```suggestion
	title: 'I am a complex dialog',
```

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+	dismissButton: 'true',
+	actions: [
+		{
+			label: 'Close me',
```
sai-san, line 44:
```suggestion
			label: 'Primary action',
```

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+			namespace: 'close'
+		},
+		{
+			label: 'Secondary button',
```
sai-san, line 48:
```suggestion
			label: 'Secondary action',
```

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,83 @@
...
+			namespace: 'secondary'
+		}
+	],
+	visible: true,
+	dialogText: `Lorem ipsum dolor sit amet, consectetur adipiscing elit.
```
sai-san, line 53: Heads-up that this should be replaced by the copy linked from the ticket. [See copy in Figma](https://www.figma.com/file/gi3JYwSUQYhKS0mHwK25cQwi/WiKit-Design-System?node-id=10349%3A13224).

vue-components/stories/Dialog.stories.ts
guergana: right now there is only one action, yes, so both buttons close the window. is this desired = should i just leave it like this? :)

vue-components/stories/Dialog.stories.ts
guergana: oki :)

vue-components/stories/Dialog.stories.ts
guergana: ah yes true, forgot to update it :)

vue-components/stories/Dialog.stories.ts
guergana: XD sorry

sai-san: hahaha no prob! I like this Alice in Wonderland kind of copy, actually ✨

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,86 @@
...
+		{
+			label: 'Secondary action',
+			namespace: 'secondary'
+		}
+	],
+	visible: true,
+	dialogText: `Complex dialogs can display extensive information and contain all kinds of interactive elements
+	(such as inputs, tables, tabs, lists) that may allow users to submit and edit information.
+
+	All dialogs require a title. This should be a short, one line overview of the purpose of the dialog.
```
sai-san, line 55:
```suggestion
```

sai-san: Just removing this sentence makes it look less crowded.

sai-san:
A couple of issues:

- The overllay is still not displayed behing the dialog (it's positioned to the bottom right)
Screenshot 2021-12-01 at 17 26 00

- Is there a way to increase the size of the canvas in the Docs page so the Dialog is not displayed on top of the props table? (This was done with the [Popover component](https://wmde.github.io/wikit/?path=/docs/vue_popover--basic))

A request for a change that I could apply myself in a separate PR if that sounds better:

- The size of the icon in the close button should be `large`. (This would basically require changing line 41 in the Dialog.vue file to <icon type="close" size="large" />)

Thank you! C'mon we're almost there 🙌

vue-components/stories/Dialog.stories.ts
```diff
@@ -0,0 +1,86 @@
...
+			label: 'Primary action',
+			namespace: 'close'
+		},
+		{
+			label: 'Secondary action',
+			namespace: 'secondary'
+		}
+	],
+	visible: true,
+	dialogText: `Complex dialogs can display extensive information and contain all kinds of interactive elements
```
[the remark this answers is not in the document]
guergana: ah yes, good point

guergana:
> A couple of issues:
> 
>     1. The **overllay is still not displayed behind** the dialog (it's positioned to the bottom right)
> 
> 
> Screenshot 2021-12-01 at 17 26 00
> 
>     2. Is there a way to **increase the size of the canvas in the Docs page** so the Dialog is not displayed on top of the props table? (see screenshot above too)(This was done with the [Popover component](https://wmde.github.io/wikit/?path=/docs/vue_popover--basic))
> 
> 
> A request for a change that I could apply myself in a separate PR if that sounds better:
> 
>     3. The **size of the icon in the close button** should be `large`. (This would basically require changing line 41 in the Dialog.vue file to )
> 
> 
> Thank you! C'mon we're almost there raised_hands

1. Yes, the canvas can be resized, no problem. The button I would suggest yes making a separate PR for that since that's not part of the story itself, but it's a fix to the the dialog component, right?

vue-components/stories/Dialog.stories.ts
```diff
@@ -12,7 +12,7 @@ export function complex( args: { title: string, actions: string, dismissButton:
 		components: { Button, Dialog },
 		props: Object.keys( args ),
 		template: `
-			<div>
+			<div style="height: 70vh;">
```
itamargiv, line 15: what does this fix?

sai-san: It increases the height of the canvas section in the Docs page, so the Dialog does not overlap with the props table (see my comment above).

itamargiv: oh, yeah, just saw it now

itamargiv (approved):
Approving for now, as the overlay will be fixed in another PR.

itamargiv (approved):
Approving again with the same comment as before

PR #511, opened 2021-12-09
Lint all the things that should be linted

jakobw:
`vue-cli-service lint` (as of recently??) [lints only the `src/` and `tests/` directories](https://cli.vuejs.org/core-plugins/eslint.html#injected-commands). This meant that the whole `stories/` directory wasn't being linted at all. When explicitly targeting the whole directory, it doesn't apply the `.eslintignore`. It's unclear to me what the proper way is to lint all directories, but running eslint directly without the vue-cli wrapper does the right thing, so let's just do that.


vue-components/stories/Link.stories.ts
```diff
@@ -2,60 +2,60 @@ import Link from '@/components/Link';
...
+	return {
+		data(): object {
+			return { args };
+		},
+		components: { Link },
+		props: Object.keys( args ),
+		template: `
 			<div>
 			<p><Link
                 :underlined="underlined"
```
micgro42, line 20: Somehow the indenting with spaces here is not raising an eslint issue? 🙃

vue-components/package.json
```diff
@@ -17,11 +17,11 @@
     "e2e:saucelabs": "env DATE=\"$(date)\" nightwatch --config nightwatch.config.js --env sauceChrome,sauceFirefox,sauceIE,sauceEdge,sauceSafari",
     "build:extract-stories": "npx sb extract storybook-static storybook-static/stories.json",
     "build:storybook": "build-storybook",
-    "fix": "vue-cli-service lint",
+    "fix": "eslint --ext .js,.ts,.vue --fix .",
     "fix:css": "stylelint '**/*.vue' --fix",
     "storybook": "start-storybook",
     "test": "npm-run-all test:*",
-    "test:lint": "vue-cli-service lint --no-fix --max-warnings 0",
+    "test:lint": "eslint --ext .js,.ts,.vue .",
```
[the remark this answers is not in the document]
jakobw: Nope, good catch! Fixed with 969c799.

vue-components/stories/Link.stories.ts
jakobw: Meh, template literals are special, I guess. I won't attempt to fix that in this PR.

micgro42: Fair enough. But let's keep it in mind in case we ever come across a rule that addresses it :)

PR #512, opened 2021-12-10
Force component imports from main.ts in stories

jakobw:
By forcing components to be imported from `main.ts` instead of directly importing it from the source file, we make it harder for ourselves to forget exporting a component from the vue-components package (e.g. 2c6f608ce7bccbcf7ca06ff6a770cf3c5ba715a4).

I initially thought that the number of `// eslint-disable-line no-restricted-imports` I had to add is something that should be worked around as part of this PR, however, I believe this actually shows a problem worth pointing out: whatever is useful for the component stories is likely also useful for consuming applications. Instead of making it more convenient to ignore those, we should probably think about ways to export these types/enums/constants as well!

### DIFF
--- a/vue-components/stories/.eslintrc.js
+++ b/vue-components/stories/.eslintrc.js
@@ -1,5 +1,9 @@
 module.exports = {
 	rules: {
 		'max-len': 'off', // lots of long lines due to inline templates
+		'no-restricted-imports': [ // force component imports from src/main.ts to remember exporting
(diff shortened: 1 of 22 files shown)


jakobw:
Ok, looks like this doesn't actually work well after all because we also show internal components in storybook as well, and they aren't distinguishable from ones that actually get exported. Dammit :D

PR #515, opened 2021-12-20
Prevent dropdown from emitting an 'input' event if the selected option is selected again

sai-san:
This prevents the wikit-Dropdown component from emitting an 'input' event in case users click the selected option from the menu again.

Issue: [T298020](https://phabricator.wikimedia.org/T298020)

### DIFF
--- a/vue-components/src/components/Dropdown.vue
+++ b/vue-components/src/components/Dropdown.vue
@@ -158,8 +158,8 @@ export default defineComponent( {
 
 			// the following comment generates the event's description for the docs tab in storybook
 			/**
-			 * This event is emitted whenever an item is selected on the Dropdown.
-			 * As long as the selected item is different from the current selection.
+			 * This event is emitted whenever an item is selected on the Dropdown,
+			 * as long as the selected item is different from the current selection.
 			 *
 			 * @property {MenuItem|null} The event payload contains the whole MenuItem object.
 			 *                           The payload is null when no item is selected or the item is deselected.
--- a/vue-components/src/components/Dropdown.vue
+++ b/vue-components/src/components/Dropdown.vue
@@ -159,11 +159,14 @@ export default defineComponent( {
 			// the following comment generates the event's description for the docs tab in storybook
 			/**
 			 * This event is emitted whenever an item is selected on the Dropdown.
+			 * As long as the selected item is different from the current selection.
 			 *
 			 * @property {MenuItem|null} The event payload contains the whole MenuItem object.
 			 *                           The payload is null when no item is selected or the item is deselected.
 			 */
-			this.$emit( 'input', menuItem );
+			if ( menuItem !== this.value ) {
+				this.$emit( 'input', menuItem );
+			}
 		},
 		onClick(): void {
 			this.startShowingTheMenu();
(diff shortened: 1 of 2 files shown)


vue-components/tests/unit/components/Dropdown.spec.ts
```diff
@@ -123,6 +125,21 @@ describe( 'Dropdown', () => {
 		expect( wrapper.emitted( 'input' )![ 0 ] ).toEqual( [ menuItems[ selectedItem ] ] );
 	} );
 
+	it( 'doesn\'t emit an `input` event if currently selected item is selected again', async () => {
```
jakobw, line 128: Nice!

vue-components/src/components/Dropdown.vue
```diff
@@ -159,11 +159,14 @@ export default defineComponent( {
 			// the following comment generates the event's description for the docs tab in storybook
 			/**
 			 * This event is emitted whenever an item is selected on the Dropdown.
+			 * As long as the selected item is different from the current selection.
```
jakobw, line 162: Super nitpicky, but IMO this would read better if it was part of the previous sentence, i.e. "This event is emitted whenever an item is selected on the Dropdown as long as the selected item is different from the current selection."

vue-components/tests/unit/components/Dropdown.spec.ts
```diff
@@ -5,9 +5,11 @@ import OptionsMenu from '@/components/OptionsMenu.vue';
 import ValidationMessage from '@/components/ValidationMessage.vue';
 import { MenuItem } from '@/components/MenuItem';
 
-async function createDropdownWrapperWithExpandedMenu( menuItems: MenuItem[] ): Promise<Wrapper<Dropdown>> {
+async function createDropdownWrapperWithExpandedMenu( menuItems: MenuItem[],
+	value?: Record<string, any> | null ): Promise<Wrapper<Dropdown>> {
 	const wrapper = mount( Dropdown, { propsData: {
 		menuItems,
+		value: value || null,
```
jakobw, line 12: This can be simplified a tiny bit:
```suggestion
	value?: MenuItem ): Promise<Wrapper<Dropdown>> {
	const wrapper = mount( Dropdown, { propsData: {
		menuItems,
		value,
```

We don't need to explicitly pass `null` because the component prop will default to `null` on its own if it's undefined. We can also use `MenuItem` instead of a `Record` because that's exactly what the Dropdown expects.

jakobw (changes requested):
Found some tiny things to nitpick about, but otherwise this looks great! It's also consistent with how native elements behave, i.e. a `<select>` doesn't emit events if the current selection is reselected. Makes a lot of sense! \o/

vue-components/tests/unit/components/Dropdown.spec.ts
sai-san: Makes sense! Committing. Thanks, Jakob 💯

vue-components/src/components/Dropdown.vue
sai-san: replaced full stop by a coma, but kept the comment in two lines to avoid surpassing the character limit. Is that ok, or should I disable max-len?

jakobw (approved):

Silvan-WMDE (approved):
Looks good, tested locally, works. Thank you - great job! :100:

PR #516, opened 2021-12-20
Add browser tests to Dialog component

guergana:
This PR adds the following browser tests to the Dialog component:

- Traps page focus, so that only visually focused elements are tab-able
- Prevents underlying page from scrolling when opened and on initial render
- Reset dialog scroll bars to top when closed and reopened

Some bugs have been discovered in IE11 and Safari while writing this PR so the test that involves the tabs have been disabled in this PR. The test will be reenabled on those browsers once the issues have been tackled.

Bug: [T295876](https://phabricator.wikimedia.org/T295876)


vue-components/tests/e2e/specs/Dialog.test.js
```diff
@@ -0,0 +1,117 @@
...
+	} );
+
+	it( 'traps page focus, so that only visually focused elements are tab-able', function ( client ) {
+
+		const currentBrowser = client.options.desiredCapabilities.browserName;
+
+		// exclude internet explorer for now because there is a known bug in the .keys
+		// and .sendKeys functions
+		// ie: https://github.com/nightwatchjs/nightwatch/issues/2536
+		// firefox: https://github.com/nightwatchjs/nightwatch/issues/2468
```
jakobw, line 35: This comment isn't up to date, right? We worked around the `.keys` problem to some extent. It should probably instead mention whatever new problem we ran into, and that safari still didn't work despite all workarounds and despite looking fine in manual tests.

vue-components/tests/e2e/specs/Dialog.test.js
```diff
@@ -0,0 +1,117 @@
...
+						client.assert.equal( elementText2, 'Secondary action' );
+					} );
+				} )
+				// different browsers are counting the combinations differently.
+				// firefox starts in the previous selected element,
+				// while chrome starts the tab count on each elementActive call
+				// chrome: makes sure it goes back to the first element after four tabs
+				.sendKeys( 'body', currentBrowser !== 'firefox' ?
+					client.Keys.TAB + client.Keys.TAB + client.Keys.TAB + client.Keys.TAB
+					: client.Keys.TAB + client.Keys.TAB )
```
jakobw, line 65: Same as above

vue-components/tests/e2e/specs/Dialog.test.js
```diff
@@ -0,0 +1,117 @@
...
+	} );
+
+	it( 'resets dialog scroll bars to top when closed and reopened', function ( client ) {
+		if ( client.options.desiredCapabilities.browserName !== 'internet explorer' ) {
+			client
+				.click( '.wikit-Button' )
+				.pause( 500 )
+				.waitForElementPresent( '.wikit-Dialog' )
+				// substract 100px from the dialog's original height so scrollbars are shown
+				// TODO: see why this is failing in IE11 in saucelabs
```
jakobw, line 100: We know that it fails because the whole Dialog doesn't load, right?

vue-components/tests/e2e/specs/Dialog.test.js
```diff
@@ -0,0 +1,117 @@
...
+
+	it( 'traps page focus, so that only visually focused elements are tab-able', function ( client ) {
+
+		const currentBrowser = client.options.desiredCapabilities.browserName;
+
+		// exclude internet explorer for now because there is a known bug in the .keys
+		// and .sendKeys functions
+		// ie: https://github.com/nightwatchjs/nightwatch/issues/2536
+		// firefox: https://github.com/nightwatchjs/nightwatch/issues/2468
+		if ( currentBrowser !== 'internet explorer' && currentBrowser !== 'safari' ) {
```
jakobw, line 36: If possible I'd prefer using [guard clauses](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) here instead of nesting the whole test, i.e.
```
if ( currentBrowser === 'internet explorer' || currentBrowser === 'safari' ) {
  return;
}
// actual test
```

Even cleaner could probably be using [test tags](https://nightwatchjs.org/guide/running-tests/test-tags.html#using-test-tags) but if I understood it right this would only work per file, so that would probably not work nicely.

vue-components/tests/e2e/specs/Dialog.test.js
guergana: yep. forgot to check the comments. :D

vue-components/tests/e2e/specs/Dialog.test.js
guergana: I agree that test tags would probably not work in this case. At least not as shown in the linked documentation.

guergana: addressed

vue-components/tests/e2e/specs/Dialog.test.js
guergana: Yes, agree :) Addressed this too.

vue-components/tests/e2e/specs/Dialog.test.js
```diff
@@ -0,0 +1,117 @@
...
+					client.elementIdText( getWebElementIdFromActiveElementResult( result ), ( element ) => {
+						const elementText1 = element.value.replace( /\s+/g, ' ' ).trim();
+						client.assert.equal( elementText1, 'Primary action' );
+					} );
+				} )
+				// sending multiple keys inside array will fail in Firefox.
+				// See: https://github.com/nightwatchjs/nightwatch/issues/2468#issuecomment-738089219
+				.sendKeys( 'body',
+					currentBrowser !== 'firefox' ?
+						client.Keys.TAB + client.Keys.TAB : client.Keys.TAB )
```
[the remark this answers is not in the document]
guergana: I am not sure this is solving the same issue though. What happens here is that firefox and non-firefox browsers need a different number of tabs.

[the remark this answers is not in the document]
guergana: We would still need to do the conditional within the function. I am also not sure what `sendKeysToActiveElement` is doing. Why does it need the classes?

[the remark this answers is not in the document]
jakobw: > I am not sure this is solving the same issue though. What happens here is that firefox and non-firefox browsers need a different number of tabs, because they are being counted differently by the browser it seems.

Are you sure? They seem to behave the same when I try it manually in chromium and firefox. My guess is that this is a subtle difference in how the browsers behave when doing `.sendKeys( 'body', ... )`. This `sendKeysToActiveElement` I made is basically a poor man's reimplementation of `.keys()`. Instead of tabbing twice from the `body` element, it finds the currently active element and sends the key to that. Looks like #520 works, so I'll add a clarifying comment to the custom command.

jakobw (approved):
WOOOO :tada:

PR #517, opened 2021-12-21
browsertests: use custom `keys` command

jakobw:


guergana:
We have decided to just skip IE for now since the whole Dialog component is not displaying in that browser

PR #522, opened 2022-01-13
revert arrow function to traditional js to support IE11

chukarave:
Phab Task: https://phabricator.wikimedia.org/T298238

IE11 does not support es6 syntax. function adjusted to traditional js seems

### DIFF
--- a/vue-components/stories/Dialog.stories.ts
+++ b/vue-components/stories/Dialog.stories.ts
@@ -23,7 +23,7 @@ export function complex( args: { title: string; actions: string; dismissButton: 
 					:title="title"
 					ref="simple"
 					:actions="actions"
-					@action="function ( _, dialog ){ return dialog.hide() }"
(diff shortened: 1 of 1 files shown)


vue-components/stories/Dialog.stories.ts
```diff
@@ -23,7 +23,7 @@ export function complex( args: { title: string; actions: string; dismissButton:
 					:title="title"
 					ref="simple"
 					:actions="actions"
-					@action="(_, dialog) => dialog.hide()"
+					@action="function ( _, dialog ){ return dialog.hide() }"
```
lucaswerkmeister, line 26: This should also work without the `return`, right? I don’t think we care about the return value here, only about the effect of calling the function.

chukarave: yep!

lucaswerkmeister (approved):

PR #528, opened 2022-02-07
Add additional build for Vue 3 compatibility mode

lucaswerkmeister:
This adds an extra build to dist, named `wikit-vue-components-vue3compat`, which includes some changes to make Wikit compatible with the Vue 3 migration build:

- the `@vue/composition-api` dependency is externalized, because it’s incompatible with Vue 3
- components that use an HTML element with the same name (e.g. `Button`/`<button>`) internally use a non-conflicting name (but are still imported and used under their original name elsewhere), because otherwise there are sometimes infinite loops when mounting them

The intention is that this additional build can be used in two ways:

1. with Vue 2, where `vue` is `vue@^2` and `@vue/composition-api` is `@vue/composition-api@^1`
2. with Vue 3, where `vue` is `@vue/compat@^3` and `@vue/composition-api` is an alias for the same module

Note that even in option 1, `@vue/composition-api` must now be available in the environment, whereas in the regular (non-vue3compat) build it is built into the bundle and only Vue itself is required.

Note that this only makes Wikit compatible with the migration build, not with pure Vue 3: we still use Vue 2 APIs, this just includes the bare minimum of changes to make the library work under the migration build.

### DIFF
--- a/vue-components/README.md
+++ b/vue-components/README.md
@@ -4,6 +4,8 @@ The component implementation of [WiKit](https://github.com/wmde/wikit) in vue
 
 ## Installation
 
+## Vue 2
+
 **The library assumes the consuming application uses Vue.js v2.6+.**
 
 #### 1. Install via npm
@@ -24,6 +26,16 @@ or in your entry component, e.g `App.vue`, add
 
 `import { Button } from '@wmde/wikit-vue-components';`
 
+### Vue 3 migration build
+
+There is an experimental build that works with the Vue 3 migration build (`@vue3/compat`).
+To use it, change your imports to
+`@wmde/wikit-vue-components/dist/wikit-vue-components-vue3compat.common.js`
+(or configure `@wmde/wikit-vue-components` to be an alias for that),
+then configure your build so that `vue` and `@vue/composition-api` are both aliases for `@vue/compat`.
+
+Keep in mind that this build still uses Vue 2 APIs,
+so it is only compatible with the migration build, not with pure Vue 3.
 
 ## Development
 
--- a/vue-components/package.json
+++ b/vue-components/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "serve": "vue-cli-service serve",
-    "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json && tsc --emitDeclarationOnly",
+    "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json && tsc --emitDeclarationOnly && VUE_APP_VUE3COMPAT=true vue-cli-service build --target lib --formats commonjs --name wikit-vue-components-vue3compat --no-clean src/main.ts",
     "test:unit": "vue-cli-service test:unit",
     "test:a11y": "jest --testMatch=\"**/tests/a11y/**\"",
     "e2e": "vue-cli-service test:e2e",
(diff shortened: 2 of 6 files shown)


lucaswerkmeister:
Not sure what to do about the failing test, I don’t really see how it could be caused by this PR…

```
TEST FAILURE: 1 error during execution; 0 assertions failed, 100 passed (1m 59s)

 ✖ Dialog.test
 – traps page focus, so that only visually focused elements are tab-able (1m 6s)

…

  TypeError: Error while running "elementActive" command: Cannot convert undefined or null to object
       at Function.values (<anonymous>)
       at getWebElementIdFromActiveElementResult (/home/runner/work/***/***/vue-components/tests/e2e/specs/Dialog.test.js:11:16)
       at NightwatchAPI.<anonymous> (/home/runner/work/***/***/vue-components/tests/e2e/specs/Dialog.test.js:40:27)
       at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

micgro42:
I've triggered a re-run. Let's see if it fails consistently.

vue-components/README.md
```diff
@@ -24,6 +26,16 @@ or in your entry component, e.g `App.vue`, add
 
 `import { Button } from '@wmde/wikit-vue-components';`
 
+### Vue 3 migration build
+
+There is an experimental build that works with the Vue 3 migration build (`@vue3/compat`).
+To use it, change your imports to
+`@wmde/wikit-vue-components/dist/wikit-vue-components-vue3compat.common.js`
+(or configure `@wmde/wikit-vue-components` to be an alias for that),
```
lucaswerkmeister, line 34: The following Vite config snippet seems to work for that (in one of the PoC changes):

```js
resolve: {
    alias: {
        '@wmde/wikit-vue-components/dist/wikit-vue-components.css': path.resolve( __dirname, 'node_modules/@wmde/wikit-vue-components/dist/wikit-vue-components.css' ),
        '@wmde/wikit-vue-components': path.resolve( __dirname, 'node_modules/@wmde/wikit-vue-components/dist/wikit-vue-components-vue3compat.common.js' ),
    },
},
```

lucaswerkmeister:
> I've triggered a re-run. Let's see if it fails consistently.

Looks like it passed this time.

vue-components/package.json
```diff
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "serve": "vue-cli-service serve",
-    "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json && tsc --emitDeclarationOnly",
+    "build": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json && tsc --emitDeclarationOnly && VUE_APP_VUE3COMPAT=true vue-cli-service build --target lib --formats commonjs --name wikit-vue-components-vue3compat --no-clean src/main.ts",
```
itamargiv, line 13: This command is a bit hefty, this is really non-blocking though. I think in a followup, we should split this into two script entries and combine them with `npm-run-all -s`

itamargiv (approved):
Looking good :sunglasses: on tiny suggestion for a followup, but it's not blocking, so feel free to merge, and we can follow up later.

vue-components/package.json
lucaswerkmeister: Hmm, but it can’t be `npm-run-all build:*`, because we already have some `build:` tasks that shouldn’t be part of `build` (as I just noticed after implementing this suggestion locally ^^)

lucaswerkmeister: #529

PR #533, opened 2022-02-07
Stop building and testing for IE11

micgro42:
This branch is going to target Vue 3 which no longer support IE11 and
Safari older than 10. See also wikimedia/browserslist-config-wikimedia#6

This should make future builds more robust and slightly faster.

Bug: T301155

### DIFF
--- a/vue-components/.browserslistrc
+++ b/vue-components/.browserslistrc
@@ -2,8 +2,7 @@ last 2 Chrome versions
 last 2 Firefox versions
 last 2 Opera versions
 last 2 Edge versions
-IE >=11
-Safari >=5.1
-iOS >=6.1
-Android >=4.1
-defaults
+
+Safari >=10
+iOS >=10
+last 2 Android versions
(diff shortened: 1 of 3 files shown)


vue-components/.browserslistrc
```diff
@@ -2,8 +2,7 @@ last 2 Chrome versions
 last 2 Firefox versions
 last 2 Opera versions
 last 2 Edge versions
-IE >=11
-Safari >=5.1
-iOS >=6.1
-Android >=4.1
-defaults
+
```
lucaswerkmeister, line 5: Intentional blank line?

lucaswerkmeister (approved):
Looks good to me, no other IE(11) references found outside of ADRs.

vue-components/.browserslistrc
micgro42: Yes :/. But I rather not add another commit, the browser tests are green for once. That will be fixed once wikimedia/browserslist-config-wikimedia#6 is merged and this file is replaced by just referencing the modern-es6 config.

PR #534, opened 2022-02-08
Remove components that will not yet be migrated to vue 3

micgro42:
This should reduce the cognitive overhead and also make the tests run a
bit faster and more stable.

These components can be readded if and when we need to migrate them to
vue 3.

Bug: T301155


lucaswerkmeister (changes requested):
I think there are a few more things that can be removed:

- the `validateExtendedNumberInput` util function and its test
- the `InlineProgressBar` mixin
- the `Link` mixin
- the `PopoverPositions` enum

micgro42:
> I think there are a few more things that can be removed:
> 
>     * the `validateExtendedNumberInput` util function and its test
> 
>     * the `InlineProgressBar` mixin
> 
>     * the `Link` mixin
> 
>     * the `PopoverPositions` enum

Mh, that's a bit bigger than the inteded scope, but sure, it doesn't hurt. I will update this PR.

lucaswerkmeister (approved):

micgro42:
I feel I want a check-list somewhere for how often I have to restart and wait for the browser tests to finish

PR #538, opened 2022-02-11
Apply extra attributes to the input tag of the TextInput component

micgro42:
This allows setting attributes such as required or maxlength.


vue-components/src/components/TextInput.vue
```diff
@@ -19,6 +19,7 @@
 			:feedback-type="feedbackType"
 			:placeholder="placeholder"
 			:disabled="disabled"
+			v-bind="otherAttributes"
```
itamargiv, line 22: Classic usage of `v-bind` magic. Thanks for fixing up!

itamargiv (approved):
Looks good, tested locally, attributes end up on the correct elements :+1:

lucaswerkmeister (approved):
Looks good; do you know if there’s any documentation for the setup function and its `context` argument? The [`inheritAttrs` docs](https://v2.vuejs.org/v2/api/#inheritAttrs) I found only mention `this.$attrs`.

micgro42:
> Looks good; do you know if there’s any documentation for the setup function and its `context` argument? The [`inheritAttrs` docs](https://v2.vuejs.org/v2/api/#inheritAttrs) I found only mention `this.$attrs`.

I think this is the official docs: https://vuejs.org/api/composition-api-setup.html#setup-context

lucaswerkmeister:
> > Looks good; do you know if there’s any documentation for the setup function and its `context` argument? The [`inheritAttrs` docs](https://v2.vuejs.org/v2/api/#inheritAttrs) I found only mention `this.$attrs`.
> 
> I think this is the official docs: https://vuejs.org/api/composition-api-setup.html#setup-context

Those are the Vue 3 docs, we’re still targeting Vue 2 in this branch… does that mean that the feature already exists in Vue 2, just without being documented?

micgro42:
> > > Looks good; do you know if there’s any documentation for the setup function and its `context` argument? The [`inheritAttrs` docs](https://v2.vuejs.org/v2/api/#inheritAttrs) I found only mention `this.$attrs`.
> > 
> > 
> > I think this is the official docs: [vuejs.org/api/composition-api-setup.html#setup-context](https://vuejs.org/api/composition-api-setup.html#setup-context)
> 
> Those are the Vue 3 docs, we’re still targeting Vue 2 in this branch… does that mean that the feature already exists in Vue 2, just without being documented?

We're using the composition-api plugin and it was my understanding that it provides `setup()` mostly as it is in vue3. But maybe I need to double check

lucaswerkmeister:
Ah, I think that’s the part I was missing. Thanks!

PR #543, opened 2022-02-14
Upgrade vue-cli to 5.0.0-rc3

micgro42:
This is more or less the last step before switching to vue3

`vue-loader` and `webpack` needed to be added to deal with some dependency resolution issues and also building seemed to have strange errors due to unaligned loader versions.


micgro42:
Some tests are still failing locally. I'll have a look at what is going tomorrow.

PR #546, opened 2022-02-21
Replace stylelint-use-logical with stylelint-use-logical-spec

lucaswerkmeister:
The original stylelint-use-logical seems to be effectively unmaintained, and is causing problems with its old stylelint peer dependency; its fork stylelint-use-logical-spec seems reasonably active, has a comparable number of downloads, and is compatible with newer stylelint.

The new version also complains about width/height (wanting them replaced with inline/block-size), but at least one of these replacements is not trivial (in OptionsMenu, the max-width involves the vw unit), so disable this for now.

---

Inspired by 55169482695961c64aa5ceef2a1b7252a398c291 from #543.

### DIFF
--- a/vue-components/package-lock.json
+++ b/vue-components/package-lock.json
@@ -65,7 +65,7 @@
 				"stylelint": "^13.13.1",
 				"stylelint-config-sass-guidelines": "^7.1.0",
 				"stylelint-config-standard": "^20.0.0",
-				"stylelint-use-logical": "^1.1.0",
+				"stylelint-use-logical-spec": "^3.2.2",
 				"typescript": "~4.5.0",
 				"vue-template-compiler": "^2.6.12"
 			}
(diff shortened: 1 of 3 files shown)


lucaswerkmeister:
This could also be done on the master branch, I suppose.

lucaswerkmeister:
> The new version also complains about width/height (wanting them replaced
with inline/block-size), but at least one of these replacements is not
trivial (in OptionsMenu, the max-width involves the vw unit), so disable
this for now.

I’m not sure how I feel about this part of the change, to be honest… but if we want to replace width/height as well, I suppose we could at least do that separately, and still merge this.

micgro42:
> > The new version also complains about width/height (wanting them replaced
> > with inline/block-size), but at least one of these replacements is not
> > trivial (in OptionsMenu, the max-width involves the vw unit), so disable
> > this for now.
> 
> I’m not sure how I feel about this part of the change, to be honest… but if we want to replace width/height as well, I suppose we could at least do that separately, and still merge this.

No, I think this is a good call. Overall, I'd like to limit the work we put into wikit to what is necessary for our immediate use case. Codex seems to be getting stable _fast_ and so there is probably no point in doing long-term focused maintenance work on wikit.

micgro42 (approved):

PR #548, opened 2022-03-01
Update vue-cli to version 5

micgro42:
To be honest, most of this is poking things until they turned green. In particular, I have no idea why it now needs to be `.trigger( 'focus.native' );`, but that `.native` is going to go away with the switch to vue 3 anyway, so I'm not particularly invested in fully figuring out this mystery.


lucaswerkmeister (approved):
Looks alright to me and seems to work in CI. Do you think any of it can be pulled out into a smaller PR? Maybe b84cb1eb6d7b277fe6866076c2ffd5791b1ea165, if the new nightwatch assertions were already available in the previous version?

I also added 5c99f02889c2f42e8cf01dead518b28b6a8475fb at the end since I noticed the broken URL while looking at the config; I didn’t realize that the URL didn’t actually belong to the PR, and has been in there for who knows how long (I can’t be bothered to check the `git blame`). Feel free to discard that commit.

vue-components/package-lock.json
```diff
@@ -62,6 +62,7 @@
 				"postcss": "^7.0.35",
 				"postcss-dir-pseudo-class": "^5.0.0",
 				"postcss-inline-svg": "^4.1.0",
+				"postcss-loader": "^6.2.1",
```
lucaswerkmeister, line 65: I’m confused by this change (and the webpack removal below) – isn’t this part of the package-lock.json supposed to be the same as the (dev)dependencies of package.json? But the package.json isn’t touched in this commit.

micgro42:
> Looks good to me if it’s going to be squashed.

wikit only allows squashes, so yes :)

vue-components/package-lock.json
micgro42: yes, npm is incredibly weird. as I understand it, `package-lock.json` is only updated iff `node_modules` is updated. So if something is changed to `package.json` and that change did not result in a change being need in `node_modules`, then `package-lock.json` is also not updated. Even though some dependencies might have nominally changed in the top-level `package.json`. I'm not yet 100% sure for how this plays with `npm install --package-lock.only`.

PR #551, opened 2022-03-04
Jump to Vue 3

micgro42:
The components build now successfully, tests pass and storybook works.

The `Message` component is still missing. It was accidentally removed as part of #534 and will be added back in an extra PR after this one is merged.


micgro42:
This is now green, except for the visual regression tests, but strangely, they seem to have caught something somewhat real? The little whitespace that used to be between components is now gone. o.O

micgro42:
Let me know if there are any questions that are not answered in the commit message.

chukarave (approved):

micgro42:
the two `any` types in production code will be replaced in a follow-up PR

PR #554, opened 2022-03-10
Replace `any` types in production code

micgro42:
The `any` used for the `extraStyles` prop does not actually cause an
error when omitted and building wikit. However it does show up in
VSCode, probably because VSCode uses Volar internally which uses a more
advanced parser for vue files than @vue/cli. Still, it is unclear what
the right way forward here would be. So far, it seemed to not be
possible to type it as `StyleValue`, but maybe I just didn't find the
right place for it.

### DIFF
--- a/vue-components/src/components/OptionsMenu.vue
+++ b/vue-components/src/components/OptionsMenu.vue
@@ -43,7 +43,7 @@
 </template>
 
 <script lang="ts">
-import { defineComponent, PropType } from 'vue';
+import { ComponentPublicInstance, defineComponent, PropType } from 'vue';
 import debounce from 'lodash/debounce';
 import { MenuItem } from './MenuItem';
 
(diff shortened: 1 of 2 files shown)


lucaswerkmeister (dismissed):

lucaswerkmeister:
> So far, it seemed to not be possible to type it as `StyleValue`, but maybe I just didn't find the right place for it.

Did you try this place? It doesn’t seem to produce any build errors, at least (I can’t say if it helps in VSCode).

```patch
diff --git a/vue-components/src/components/TextInput.vue b/vue-components/src/components/TextInput.vue
index 39fa43d8..d44e582e 100644
--- a/vue-components/src/components/TextInput.vue
+++ b/vue-components/src/components/TextInput.vue
@@ -29,9 +29,9 @@
 	</div>
 </template>
 
 <script lang="ts">
-import { defineComponent, computed } from 'vue';
+import { defineComponent, computed, StyleValue } from 'vue';
 import ValidationMessage from './ValidationMessage.vue';
 import Input from './Input.vue';
 import generateUid from '@/components/util/generateUid';
 import { errorProp, ErrorProp, getFeedbackTypeFromProps } from '@/compositions/validatable';
@@ -47,9 +47,9 @@ export default defineComponent( {
 	setup( props: { error: ErrorProp }, context ) {
 		const { class: extraClasses, style: extraStyles, ...otherAttributes } = context.attrs;
 		return {
 			extraClasses,
-			extraStyles,
+			extraStyles: extraStyles as StyleValue,
 			otherAttributes,
 			feedbackType: computed( getFeedbackTypeFromProps( props ) ),
 		};
 	},
```

lucaswerkmeister (approved):

PR #555, opened 2022-03-10
Readd Message component to next branch

micgro42:
That Message component was deleted in #534, but we do in fact need it in the current project. So this PR adds it back. 

To make review easier, the first commit just plain copies the files from the master branch without any changes, and the second commit makes the adjustments that are needed for Vue3.

### DIFF
--- a/vue-components/src/components/Message.vue
+++ b/vue-components/src/components/Message.vue
@@ -1,0 +1,101 @@
+<template>
+	<div
+		:class="[
+			'wikit',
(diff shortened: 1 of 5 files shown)


lucaswerkmeister (approved):
Code looks okay, not sure what to make of the one failed build.

My local storybook decides to show me the “All” story before the “Basic” one, but the order also differs from the gh-pages storybook in the Button component, so that’s not a problem specific to this PR.

PR #556, opened 2022-03-11
Bump premajor version to v3.0.0-alpha.0

chukarave:
As part of [#T303503](https://phabricator.wikimedia.org/T303503) we would like to use a Vue3 version of some of the Wikit components, created in the next branch.


chukarave:
throwing this away due to unintended `package-lock.json` diff

PR #558, opened 2022-03-11
Bump premajor version to v3.0.0-alpha.0

chukarave:
As part of [#T303503](https://github.com/wmde/wikit/pull/557) we would like to use a Vue3 version of some of the Wikit components, created in the next branch.


lucaswerkmeister:
I think we should also change the two `2.1.0-alpha.12` versions in `vue-components/package-lock.json` (related to wikit-tokens) to 3.0.0-alpha.0.

lucaswerkmeister (approved):

PR #564, opened 2022-03-15
Add sentence about old version numbers after bumping version

micgro42:


micgro42:
The changed copy is ready for review. Though the styles look slightly off. Have we written the CSS for these ourselves or is that some standard CSS that wasn't created with sub-lists in mind?

sai-san:
> The changed copy is ready for review. Though the styles look slightly off. Have we written the CSS for these ourselves or is that some standard CSS that wasn't created with sub-lists in mind?

The mdx pages are using the default Storybook styling. Using asterisks instead of dashes seems to fix it. Nevertheless, since the sentence doesn't describe a "step" or action, do you mind if we include it in the parent bullet point instead of creating a new indentation? If this need to be somehow highlighted, we could add  `**Note:**` in front of this bit.

Thanks so much for this improvement, @micgro42 😊

micgro42:
> > The changed copy is ready for review. Though the styles look slightly off. Have we written the CSS for these ourselves or is that some standard CSS that wasn't created with sub-lists in mind?
> 
> The mdx pages are using the default Storybook styling. Using asterisks instead of dashes seems to fix it. Nevertheless, since the sentence doesn't describe a "step" or action, do you mind if we include it in the parent bullet point instead of creating a new indentation? If this need to be somehow highlighted, we could add `**Note:**` in front of this bit.
> 
> Thanks so much for this improvement, @micgro42 blush

I'm sorry for dropping a bit the ball on this PR. You make a good point and I have adjusted the PR accordingly.

sai-san (approved):
LGTM! Thanks again, @micgro42 💯

PR #571, opened 2022-03-29
Upgrade dependencies

micgro42:
This should only include fully semver-compatible upgrades, it was done
with `npm upgrade` in all the directories.


micgro42:
Not sure why, but for some reason this now fails with type errors when building, even though only `npm upgrade` was run everywhere. It seems somehow some type definitions changed or something around how the vue code is parsed.

This was supposed to be an easy maintenance win over merging all those dependabot PRs individually, but apparently that is still our best way forward.

PR #575, opened 2022-03-31
Update a11y message component

chukarave:
The icons in the Message component and the Validation message and such have a purely decorative purpose and that should be reflected in their markup so that screen readers do not get confused.

Bug: T303810

### DIFF
--- a/vue-components/src/components/Icon.vue
+++ b/vue-components/src/components/Icon.vue
@@ -17,6 +17,8 @@
 			fill="none"
 			xmlns="http://www.w3.org/2000/svg"
 			class="wikit-Icon__svg"
+			aria-hidden="true"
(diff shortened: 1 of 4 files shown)


micgro42:
While it should work in for "fixing" the message component, I think this would be better applied to the Icon component directly. I can't think of a case where one of those SVG icon should not get an application specific alternative text, and thus the icon itself be ideally completely ignored.

micgro42 (approved):
I feel at some point we maybe should restructure this component. There is so much duplication here :/

That being said, this PR looks good for what it set out to do :+1:

lucaswerkmeister:
> I feel at some point we maybe should restructure this component. There is so much duplication here :/

I tried something in #577.

PR #576, opened 2022-04-04
Cherry Pick "Update a11y message component (#575)"

micgro42:
* add aria-hidden attribute to decorative svg icons

* add focusable=false for IE so that svg icons are decalared decorative

* add html attributes directly to the icon component and adjust tests


lucaswerkmeister:
(Note to self or others: cherry-pick #577 after this is merged. I’m not creating a PR for that yet because it would have to be rebased anyways ^^)

PR #577, opened 2022-04-04
Extract common Icon <svg> attributes

lucaswerkmeister:
All `<svg>` elements shared five common attributes, which can be extracted into a common object using `v-bind` – except that the Search type didn’t have a `fill=none` attribute before. I assume this wasn’t intentional (or at least, there’s no discussion of it in #371), so it should be okay to add the attribute with this change. (There’s no snapshot to update for this case because there’s no snapshot specifically for the Search type.)

### DIFF
--- a/vue-components/src/components/Icon.vue
+++ b/vue-components/src/components/Icon.vue
@@ -13,12 +13,8 @@
 
 		<!-- add icon -->
 		<svg
+			v-bind="SVG_ATTRIBUTES"
 			viewBox="0 0 21 20"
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
> > (There’s no snapshot to update for this case because there’s no snapshot specifically for the Search type.)
> 
> That feels like right the next thing we should fix. Probably by merging all those snapshot tests into a parametrized one that iterates over all the item types from `iconProps.ts`

Done in #578.

PR #585, opened 2022-05-23
Add story for disabled focused Button

lucaswerkmeister:
Clicking the first button should immediately make it appear disabled in all browsers.

---

Tests #584.

PR #586, opened 2022-05-23
Add story for disabled focused Button (without fix)

lucaswerkmeister:
In Firefox, clicking the first button does not make it appear disabled; only a second click will do that (it can be a click in the exact same place).

---

Tests behavior without #584 (see #585 for behavior with fix).


lucaswerkmeister:
Decision in Mattermost was to not merge the story.

PR #587, opened 2022-05-23
Add aria attributes for screen readers

guergana:
Add more specific and complete aria attributes to allow screen
readers to announce menu and menu items correctly.

TODO:

- add role alert in a way that it is read by screen readers every time the 'no-results' panel is shown. the default behavior is that it is read only once if the text remains the same. This issue will be addressed in a follow-up PR.

Bug: [T290733](https://phabricator.wikimedia.org/T290733)

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -27,8 +27,7 @@
 			@input="$emit('input', $event)"
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
 			aria-autocomplete="list"
-			aria-haspopup="menu"
-			:aria-label="label"
+			aria-haspopup="listbox"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -26,8 +26,6 @@
 			@update:searchInput="$emit('update:searchInput', $event)"
 			@input="$emit('input', $event)"
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
-			aria-autocomplete="list"
-			aria-haspopup="listbox"
 			:label="label"
 		>
 			<template v-slot:no-results>
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -26,6 +26,9 @@
 			@update:searchInput="$emit('update:searchInput', $event)"
 			@input="$emit('input', $event)"
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
+			aria-autocomplete="list"
+			aria-haspopup="menu"
+			:aria-label="label"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -28,6 +28,7 @@
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
 			aria-autocomplete="list"
 			aria-haspopup="listbox"
+			:label="label"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -137,6 +137,11 @@ export default defineComponent( {
 		onInput( value: string ): void {
 			this.showMenu = this.canShowMenu( value );
 
+			// reset hover state id to reset aria-activedescendant
+			if ( value === '' ) {
+				this.keyboardHoverId = null;
+			}
+
 			// the following comment generates the event's description for the docs tab in storybook
 			/**
 			 * Enables the `searchInput` prop to be used with the `.sync` modifier. It's used to transport the value of
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -72,7 +72,7 @@ export default defineComponent( {
 			showMenu: false,
 			scrollIndexStart: null as ( number | null ),
 			scrollIndexEnd: null as ( number | null ),
-			keyboardHoverId: null,
+			keyboardHoverId: null as ( string | null ),
 			optionsMenuId: generateUid( 'wikit-OptionsMenu' ),
 		};
 	},
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -153,7 +153,6 @@ export default defineComponent( {
 		onFocus(): void {
 			if ( this.canShowMenu( this.searchInput ) ) {
 				this.showMenu = true;
-				this.$emit( 'expanded', true );
 			}
 		},
 		onEsc(): void {
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -29,6 +29,7 @@
 			ref="menu"
 			:aria-expanded="showMenu || 'false'"
 			:id="optionsMenuId"
+			:label="label"
 		>
 			<template v-slot:no-results>
 				<slot name="no-results" />
@@ -114,6 +115,10 @@ export default defineComponent( {
 		 * computed property to dynamically update the Lookup's `menuItems` prop.
 		 */
 		searchInput: {
+			type: String,
+			default: '',
+		},
+		label: {
 			type: String,
 			default: '',
 		},
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -121,6 +121,10 @@ export default defineComponent( {
 			type: String,
 			default: '',
 		},
+		/**
+		 * Sets the label to be passed down to the inner `<OptionsMenu>` component so it can be properly announced
+		 * by screen readers.
+		 */
 		label: {
 			type: String,
 			default: '',
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -15,6 +15,8 @@
 			v-bind="$attrs"
 			:aria-activedescendant="keyboardHoverId"
 			:aria-owns="optionsMenuId"
+			aria-autocomplete="list"
+			aria-haspopup="listbox"
 		/>
 		<OptionsMenu
 			class="wikit-LookupInput__menu"
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -7,7 +7,7 @@
 			:value="searchInput"
 			@input="onInput"
 			@focus.native="onFocus"
-			@blur.native="onBlurAndEsc"
+			@blur.native="showMenu = false"
 			:feedback-type="feedbackType"
 			:placeholder="placeholder"
 			:disabled="disabled"
@@ -28,7 +28,7 @@
 			v-show="showMenu"
 			@select="onSelect"
 			@scroll="onScroll"
-			@esc="onBlurAndEsc"
+			@esc="onEsc"
 			@keyboard-hover-change="onKeyboardHoverChange"
 			ref="menu"
 			:id="optionsMenuId"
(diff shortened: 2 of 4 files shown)


vue-components/src/components/Lookup.vue
```diff
@@ -26,6 +26,9 @@
 			@update:searchInput="$emit('update:searchInput', $event)"
 			@input="$emit('input', $event)"
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
+			aria-autocomplete="list"
+			aria-haspopup="menu"
```
micgro42, line 30: This should maybe be `listbox` instead.

Also, I think we can add this in `LookupInput` component itself.

Mainly going by this example: https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190814/examples/combobox/aria1.1pattern/listbox-combo.html#ex1_label

vue-components/src/components/LookupInput.vue
```diff
@@ -13,6 +13,8 @@
 			:disabled="disabled"
 			autocomplete="off"
 			v-bind="$attrs"
+			:aria-owns="keyboardHoverIndex > -1 ? `menu-item-${keyboardHoverIndex}` : ''"
```
micgro42, line 16: I think this should be owning the entire OptionsMenu, not the individual item

micgro42 (commented):
These are only some in-progress observations. I have yet to try it out and look into it some more

vue-components/src/components/Lookup.vue
```diff
@@ -26,6 +26,9 @@
...
+			:aria-label="label"
```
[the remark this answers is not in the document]
guergana: agree, it works the same without the label

vue-components/src/components/OptionsMenu.vue
```diff
@@ -3,6 +3,8 @@
 		:class="[ 'wikit', 'wikit-OptionsMenu' ]"
 		@scroll.passive="onScroll"
 		ref="lookup-menu"
+		role="menu"
+		:aria-expanded="showMenu || 'false'"
```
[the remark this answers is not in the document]
guergana: woops. thanks for pointing this out. i copied and pasted it in one of my tryouts.

vue-components/src/components/OptionsMenu.vue
```diff
@@ -3,7 +3,9 @@
 		:class="[ 'wikit', 'wikit-OptionsMenu' ]"
 		@scroll.passive="onScroll"
 		ref="lookup-menu"
+		role="listbox"
 		:style="{ maxHeight: maxHeight ? maxHeight + 'px' : null }"
+		aria-label="Options Menu"
```
guergana, line 8: Not sure which label to put here... Put this in the meantime. Not having an `aria-label` in the OptionsMenu generates an accesibility error.

vue-components/src/components/LookupInput.vue
```diff
@@ -23,7 +25,10 @@
 			@select="onSelect"
 			@scroll="onScroll"
 			@esc="onEsc"
+			@keyboard-hover-change="onKeyboardHoverChange"
 			ref="menu"
+			:aria-expanded="showMenu || 'false'"
```
micgro42, line 30: Mh, This doesn't work for me here. Could we try to move this `:aria-expanded="showMenu || 'false'"` to the `Input` compnent? At least with orca that seems to work better

vue-components/src/components/LookupInput.vue
```diff
@@ -140,10 +148,12 @@ export default defineComponent( {
 		onFocus(): void {
 			if ( this.canShowMenu( this.searchInput ) ) {
 				this.showMenu = true;
+				this.$emit( 'expanded', true );
```
micgro42, line 151: Is anything listening to this event?

vue-components/src/components/OptionsMenu.vue
```diff
@@ -18,6 +20,9 @@
 			@mousedown.prevent="activeItemIndex = index"
 			@mouseup="activeItemIndex = -1"
 			ref="menu-items"
+			role="option"
+			:aria-label="`${menuItem.label} - ${menuItem.description}`"
```
micgro42, line 24: Have you tried putting the description into [`aria-description`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description)?

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
micgro42: :+1:

vue-components/src/components/Lookup.vue
micgro42: If it works this simple, then let's go with this. The simpler, the better. Though we might still need it later, but we can figure that out in a second iteration.

vue-components/src/components/OptionsMenu.vue
micgro42: In storybook in the OptionsMenu story you mean? I think those lack the context of the surrounding lookup and so cause this false-positive error. As long as we use that component in a Lookup or dialog, we can probably ignore it.

guergana: Added the label from the lookup as agreed.

vue-components/src/components/OptionsMenu.vue
guergana: much better. thanks for the suggestion.

vue-components/src/components/LookupInput.vue
guergana: but what is expanded is the optionsMenu, not the Input. :thinking:

vue-components/src/components/LookupInput.vue
guergana: no, it's a leftover from something i tried. thanks for catching this. :sweat_smile:

vue-components/src/components/OptionsMenu.vue
guergana: this works when testing orca locally but later i discovered it creates the a11y checks to fail :( `role=option` is not compatible with `aria-description`. you can see the error here: https://github.com/wmde/wikit/runs/6597303500?check_suite_focus=true#step:5:193

micgro42: Ok, this is strange. It feels almost like an oversight, given that `aria-describedby` _is_ allowed in https://www.w3.org/TR/wai-aria-1.1/#option.

But we can work around this. Can you give an idea to the field with the description for each option and reference that in `aria-describedby`?

vue-components/src/components/LookupInput.vue
micgro42: I looked into it a bit more and for me it works when added to the `<Input>` if we also add the needed `role="combox"` to it. This is also mentioned in the ticket. And it also works for me here: https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190814/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html

vue-components/src/components/Lookup.vue
```diff
@@ -26,6 +26,9 @@
 			@update:searchInput="$emit('update:searchInput', $event)"
 			@input="$emit('input', $event)"
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
+			aria-autocomplete="list"
+			aria-haspopup="listbox"
```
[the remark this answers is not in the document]
guergana: ok.

vue-components/src/components/OptionsMenu.vue
guergana: That's a very good workaround! Thanks

vue-components/src/components/Lookup.vue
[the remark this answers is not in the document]
guergana: Works :)

vue-components/src/components/LookupInput.vue
guergana: It works indeed! Is it a combobox though?

vue-components/src/components/LookupInput.vue
```diff
@@ -109,6 +118,10 @@ export default defineComponent( {
 			type: String,
 			default: '',
 		},
+		label: {
```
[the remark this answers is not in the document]
guergana: Yes! Thanks for catching this.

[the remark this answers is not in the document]
guergana: The QuantityInput already has an aria-label.

vue-components/src/components/LookupInput.vue
micgro42: Yes, I think so. This sounds a lot like what we're building: https://www.w3.org/WAI/ARIA/apg/patterns/combobox/

vue-components/src/components/LookupInput.vue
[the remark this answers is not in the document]
micgro42: > The `QuantityInput` already has an aria-label.

Yes, this makes this somewhat confusing. But that aria-label is for the actual `<input>` because that component does not have a proper label (this is what the `Lookup` component adds).

On the other hand, the value coming from the  `:label="..."` prop will be used on the QuantityInput.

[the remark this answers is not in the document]
guergana: I added the label because otherwise there is an error when running the tests.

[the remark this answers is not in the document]
guergana: This is the error. 

image

micgro42:
Mh, the `role="alert"` on the "No Results" message doesn't work for, that message is never announced for me. How did you test it?

guergana:
> Mh, the `role="alert"` on the "No Results" message doesn't work for, that message is never announced for me. How did you test it?

I enabled orca.. I typed "x" in the Lookup and it says "element combobox expanded" and then it says "no match was found".

sai-san:
Hello! It's so great to finally have an accessible lookup 💯 
Leaving some findings after testing the Lookup component locally using VoiceOver on Chrome (v 102):

1. I unfortunately didn't get an alert for the "No match was found" state.
2. When you select an option from the menu, VoiceOver says, for example: "potato root vegetable not selected, (1 of 5)" and then "root vegetable, you are currently on a text element, inside a list box". So, as you can see, the description is repeated again. This shouldn't happen.
3. Lastly, the 'limited support' label is problematic: it is announced immediately after the menu option title (almost as if they were a single word) and before the description. The screenshot below displays the VoiceOver transcript in a gray box.
Would it be possible to make the 'limited support' label to be announced at the end?

Screenshot 2022-06-03 at 18 09 49

micgro42:
I tried it with NVDA, and navigating the options was fine, that is to say: the mostly same as with Codex.

About the "no-results" thingy we want probably an extra PR, because that is also shown when there are no results _yet_, because they are still loading. So _maybe_ it would make sense to leave that to the application? But maybe what is in this slot should always be announced and it is just the application deciding what to put in there? Not sure yet.

micgro42:
The point about the "limited support", I think in code we refer to it as `tag` is a good one, because I'm surprised that it is announced at all. It would probably make sense to add it somehow, but that might cause it to be read twice in voice over as well? Not sure

guergana:
> Thank you for testing, Sarai! pray
> 
> I tried it with NVDA, and navigating the options was fine, that is to say: the mostly same as with Codex.
> 
> About the "no-results" thingy we want probably an extra PR, because that is also shown when there are no results _yet_, because they are still loading. So _maybe_ it would make sense to leave that to the application? But maybe what is in this slot should always be announced and it is just the application deciding what to put in there? Not sure yet.

Maybe [aria-busy](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy) could be used when the results are not there yet, @micgro42

guergana:
> The point about the "limited support", I think in code we refer to it as `tag` is a good one, because I'm surprised that it is announced at all. It would probably make sense to add it somehow, but that might cause it to be read twice in voice over as well? Not sure

[aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby) expects one or several space separated ids as input

guergana:
@micgro42 @sai-san I have addressed the limited support tag issue in the most recent commit, could you test again in the respective screen readers? I tested in orca and the limited support is announced properly after the description.

micgro42 (approved):
I checked it again and got JAWS working on a Windows virtual box machine. It sounded ok there. It includes now the tag and only reads the tag and the description once.

Let's get this merged. This is already better than what we had before and we can potential fix issue in future pull requests. Announcing the "no-results" text will happen in context of T310521

PR #588, opened 2022-05-24
Cherry pick "Wrap button :focus, :hover and :active styles in :enabled (#584)"

lucaswerkmeister:
Firefox allows a button to be :disabled and still have :focus, if it was
programmatically disabled while focused. In this case, the :focus styles
would beat the :disabled styles, since they have higher specificity; we
want disabled buttons to always look disabled, though, so wrap all the
other styles in :enabled so they only apply when wanted.

The stylelint rule max-nesting-depth (max depth: 3), which complains
about the change, is disabled – the alternative solutions (combining
selectors together) seem worse to us in this case.

Bug: T304511

### DIFF
--- a/vue-components/src/components/Button.vue
+++ b/vue-components/src/components/Button.vue
@@ -132,29 +132,32 @@ $base: '.wikit-Button';
 		padding-block: $wikit-Button-large-padding-vertical;
 	}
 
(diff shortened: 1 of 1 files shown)


lucaswerkmeister:
Cherry-picks #584.

micgro42 (approved):

PR #590, opened 2022-05-30
Make Lookup react better to keyboard events

lucaswerkmeister:
Make the OptionsMenu prevent default actions for events it handles, but only have LookupInput send events to OptionsMenu in the first place when the menu is actually being shown, so that default actions *aren’t* being prevented when the menu is hidden (e.g. after an option was selected).

Preventing the default Enter action (form submit) when the OptionsMenu is open is what we want for T305359; the Escape and Tab versions aren’t directly related, but feel sensible to me.

Bug: T305359

----

Not yet sure if this is really the way to go, but putting it up for discussion. It seems to work well when I make the equivalent edits to my `node_modules/@wmde/wikit-vue-components/dist/wikit-vue-components.common.js` file; trying to create a build in Wikit and copy that into `node_modules/` produced some strange errors that I don’t understand.


micgro42:
Mh, that deploy to netlify is broken for me and when I try it out locally, it doesn't seem to filter results in the storybook lookup anymore?

Looking at the code it makes sense to me to treat `Enter` this way. But I'm not so sure about `Tab`, the desired behavior there is to "Select highlighted option and move focus to next element in the tab-order". Wouldn't the `event.preventDefault();` interfere with moving the focus?
What is the behavior that is prevented for the `Esc` press?

lucaswerkmeister:
> Looking at the code it makes sense to me to treat `Enter` this way. But I'm not so sure about `Tab`, the desired behavior there is to "Select highlighted option and move focus to next element in the tab-order". Wouldn't the `event.preventDefault();` interfere with moving the focus? What is the behavior that is prevented for the `Esc` press?

When I looked at the code, I felt like it made sense to generally prevent the default behavior for every event that we handle (since the ArrowUp and ArrowDown cases did this already). But I didn’t think about it too much, so let’s just remove the Tab and Esc parts again.

> Mh, that deploy to netlify is broken for me and when I try it out locally, it doesn't seem to filter results in the storybook lookup anymore?

Let’s see if this works better when we only prevent default for Enter…

lucaswerkmeister:
Pushed a new commit, but it still doesn’t seem to work properly in storybook…

lucaswerkmeister:
Hm, the strange storybook behavior already seems to happen on the next branch without this change…

lucaswerkmeister:
Okay, but with `npm -C vue-components/ run build` in wikit, and `cp -a ~/git/wikit/vue-components/dist/ node_modules/@wmde/wikit-vue-components/` in new-lexeme, the lookup seems to work fine and the Enter behavior is as I would expect it.

lucaswerkmeister:
Storybook fixed in #592, turned out to be not directly related.

micgro42 (approved):
Ok, this looks good to me, though it is hard to verify in the sense that storybook doesn't show the effect of the event bubbling or not. But based on the code change, this should do what we want it to do :+1:

micgro42:
> Hm, should we cherry pick this to the master branch as well? (Though I’m not sure how the `showMenu` access would work there, tbh.)

Not sure. Apparently, this is not an issue for query builder or mismatch finder, so I think we can keep it in the `next` branch until someone there complains. And I don't foresee any new Vue2 projects that would use Wikit.

PR #592, opened 2022-06-03
Don’t use .sync for Lookup stories

lucaswerkmeister:
.sync is Vue 2 syntax; in Vue 3, use [multiple v-model bindings][1].

[1]: https://vuejs.org/guide/components/events.html#multiple-v-model-bindings


lucaswerkmeister:
Hm, the Storybook link seems to be broken. Not sure why, or whether it’s related to this commit. (Locally, the Storybook seems to work.)

PR #593, opened 2022-06-03
Update docs/ Storybook config

lucaswerkmeister:
--static-dir (-s) is deprecated since Storybook 6.4 [1].

[1]: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated---static-dir-cli-flag

### DIFF
--- a/docs/.storybook/main.js
+++ b/docs/.storybook/main.js
@@ -1,6 +1,7 @@
 const path = require( 'path' );
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
The docs storybook is currently broken in general, so we probably don’t want to merge this until we’ve figured out what’s going on (this isn’t enough to unbreak it).

micgro42 (dismissed):

micgro42 (approved):

PR #595, opened 2022-06-03
docs: Use React 16 instead of 17

lucaswerkmeister:
@storybook/html requires React 16 (though many other storybook packages are compatible with 16 || 17 || 18), and having multiple React versions in the tree seems to cause issues.

The explicit react-is dependency can be removed from package.json, we don’t use it directly. (It’s still pulled in by other packages.)


micgro42 (approved):
We have a storybook again! :100:

PR #598, opened 2022-06-15
Next: Update some dependencies

micgro42:
The `TextInput` component was adjusted to satisfy the error message:
```
94:3  error  Name "Input" is reserved in HTML  vue/no-reserved-component-names
```


micgro42:
I don't get it. The error is here as well!?

Also, about that npm warning - where the heck is the `eslint-plugin-vue@6.2.2` coming from in the first place?

micgro42:
> > The `TextInput` component was adjusted to satisfy the error message:
> 
> Doesn’t that also need to be changed in `LookupInput.vue`?

Mh, good point. But it didn't complain for LookupInput.vue :thinking:.  The only difference that I notice as first glance would be how the `Input` component is imported in each case?

lucaswerkmeister:
This is very strange, everything seems to be green locally (so, the package/lock errors aren’t there, but also, eslint isn’t complaining about `LookupInput`). I’m confused…

.github/workflows/test-and-demo.yaml
```diff
@@ -38,6 +38,10 @@ jobs:
       uses: actions/setup-node@v2
       with:
         node-version: '16'
+    - name: Print node version
+      run: node --version
+    - name: Print npm version
+      run: npm --version
```
lucaswerkmeister, line 44:
```suggestion
```

lucaswerkmeister (changes requested):
LGTM apart from some leftover debug output (I assume); also works for me in Fresh

lucaswerkmeister (approved):

PR #599, opened 2022-06-16
Cherry pick "Add aria attributes for screen readers (#587)"

micgro42:
Add more specific and complete aria attributes to allow screen
readers to announce menu and menu items correctly.

Note: adding an alert role that is read by screen readers every time the 'no-results' panel is shown is not implemented in this PR. the default behavior is that it is read only once if the text remains the same. This issue will be addressed in a follow-up PR.

Bug: T290733

### DIFF
--- a/vue-components/tests/a11y/AllComponents.spec.ts
+++ b/vue-components/tests/a11y/AllComponents.spec.ts
@@ -151,7 +151,8 @@ describe( 'OptionsMenu', () => {
 			{ label: 'potato', description: 'root vegetable' },
 			{ label: 'duck', description: 'aquatic bird' },
 		];
-		const wrapper = mount( OptionsMenu, { props: { menuItems } } );
+		const lookupLabel = 'Label of the parent Lookup';
+		const wrapper = mount( OptionsMenu, { propsData: { menuItems, label: lookupLabel } } );
 		const results = await axe( wrapper.element, {
 			rules: {
 				'region': { enabled: false },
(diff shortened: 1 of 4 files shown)


vue-components/tests/a11y/AllComponents.spec.ts
```diff
@@ -151,7 +151,8 @@ describe( 'OptionsMenu', () => {
 			{ label: 'potato', description: 'root vegetable' },
 			{ label: 'duck', description: 'aquatic bird' },
 		];
-		const wrapper = mount( OptionsMenu, { props: { menuItems } } );
+		const lookupLabel = 'Label of the parent Lookup';
+		const wrapper = mount( OptionsMenu, { propsData: { menuItems, label: lookupLabel } } );
```
lucaswerkmeister, line 155:
```suggestion
		const wrapper = mount( OptionsMenu, { props: { menuItems, label: lookupLabel } } );
```

https://test-utils.vuejs.org/migration/#propsdata-is-now-props

lucaswerkmeister (commented):
LGTM, just noticed one thing in the tests; not sure if we should test this with a screenreader again?

micgro42:
> LGTM, just noticed one thing in the tests; not sure if we should test this with a screenreader again?

I don't think so. Or rather, we will keep checking when the no-results part is developed

lucaswerkmeister (approved):

PR #600, opened 2022-06-17
 Bump version to 3.0.0-alpha.5

chukarave:
npm run bump-version prerelease

Bug: T290733

### DIFF
--- a/docs/package-lock.json
+++ b/docs/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "@wmde/wikit-docs",
-  "version": "3.0.0-alpha.4",
+  "version": "3.0.0-alpha.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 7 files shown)


lucaswerkmeister:
The previous version bumps only had 10 insertions and deletions, why is this one three times as large?

chukarave:
I had to change a bunch of versions manually, since they weren't updated automatically and I remember it was an issue last time I did this. maybe that wasn't necessary?

lucaswerkmeister:
We didn’t do it the previous times, and the [(pre-)release docs](https://wmde.github.io/wikit/?path=/story/documentation-release-management--page#creating-a-pre-release) mention that some old version numbers will stay behind and that’s okay.

lucaswerkmeister:
We definitely discussed manual changes a few prereleases ago (the first one I did, I think), but didn’t include them in the end.

PR #606, opened 2022-06-20
Add pause after loading Dialog story

lucaswerkmeister:


lucaswerkmeister:
#608 is an improved version of this.

PR #618, opened 2022-07-26
Add aria-required attribute to Lookup, LookupInput

lucaswerkmeister:
We want to add this attribute to lookups on the new lexeme special page, and adding it to the surrounding `<div>` isn’t enough.

Bug: T313891


lucaswerkmeister:
(Draft for now, in case we break down [T313113](https://phabricator.wikimedia.org/T313113) tomorrow and this commit should be attached to a subtask.)

guergana (approved):
Looks good!

PR #620, opened 2022-07-27
vue-components: update chromedriver, geckodriver, nightwatch

lucaswerkmeister:
This resolves several npm audit warnings.


micgro42 (approved):
Impressive to see that a bump of twenty major versions goes entirely without issues. But I guess chromedriver is essentially just chrome

PR #624, opened 2022-08-04
Display WiKit Input placeholder correct color

guergana:
Bug: [T312633](https://phabricator.wikimedia.org/T312633)

### DIFF
--- a/tokens/properties/alias.json
+++ b/tokens/properties/alias.json
@@ -381,7 +381,7 @@
         "overlay": {
             "value": "{opacity.65.value}"
         },
-        "placeholder": {
+        "base": {
             "value": "{opacity.100.value}"
         }
     },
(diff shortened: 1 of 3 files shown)


micgro42 (commented):
I can confirm that the placeholder has now the correct color for the `Input` and `Lookup` components. (Though I'm a bit confused that the ui-review CI job didn't trigger a Design review.)
Also, the CSS looks fine from my side.

I'll leave this open for @sai-san to decide if the tokens are named/located fine, or if they should be alias tokens with `placeholder` in the name or something.

sai-san:
All looking great both from a design and tokens implementation side 👍  The placeholder color was verified using Safari v15.6, Chrome v103 and Firefox 103.0.1 on macOS Monterey.

sai-san (approved):
All looking great, both from a design and a tokens implementation side 👍  The placeholder color (and its opacity) was verified using Safari v15.6, Chrome v103 and Firefox 103.0.1 on macOS Monterey.

PR #627, opened 2022-08-10
Apply full opacity to TextArea's placeholder

sai-san:
This is reusing the corresponding Input token, like we did with all other TextArea properties.

Related issue: [T312633](https://phabricator.wikimedia.org/T312633)


guergana (approved):
Looks great!

PR #632, opened 2022-10-10
Add random icon

guergana:
Bug: [T302289](https://phabricator.wikimedia.org/T302289)


vue-components/src/components/iconProps.ts
```diff
@@ -12,6 +12,7 @@ export enum IconTypes {
 	INFOOUTLINED = 'info-outlined',
 	LINK = 'link',
 	NEWWINDOW = 'newwindow',
+	RANDOM = 'random',
```
sai-san, line 15:
```suggestion
	DIE = 'die',
```

vue-components/src/components/Icon.vue
```diff
@@ -159,6 +159,15 @@
 			<path fill="currentColor" d="M11 1L14.29 4.29L8.56 10.02L9.98 11.44L15.71 5.71L19 9V1H11Z" />
 		</svg>
 
+		<!-- random icon -->
+		<svg
+			v-bind="SVG_ATTRIBUTES"
+			viewBox="0 0 20 20"
+			v-else-if="type === IconTypes.RANDOM"
```
sai-san, line 166:
```suggestion
			v-else-if="type === IconTypes.DIE"
```

sai-san: This icon's name is actually 'die'

sai-san (commented):
Thanks for the addition, Guergana! We should just replace 'random' by the official name of this icon: 'die'.

guergana:
> Thanks for the addition, Guergana! We should just replace 'random' by the official name of this icon: 'die'.

Yes, I wasn't sure of the what the name should be. Thanks! May I ask... why such a dark name? :sweat_smile:

sai-san (approved):
LGTM!

vue-components/package.json
```diff
@@ -14,7 +14,7 @@
     "build:main": "vue-cli-service build --target lib --formats commonjs --name wikit-vue-components src/main.ts --report --report-json",
     "build:tsdecl": "tsc --emitDeclarationOnly",
     "build:vue3compat": "VUE_APP_VUE3COMPAT=true vue-cli-service build --target lib --formats commonjs --name wikit-vue-components-vue3compat --no-clean src/main.ts",
-    "test:unit": "vue-cli-service test:unit",
+    "test:unit": "vue-cli-service test:unit -u",
```
lucaswerkmeister, line 17: `-u` should not be part of the regular test command.

guergana: Can you please offer more constructive review? What can be another way of updating the snapshot?

lucaswerkmeister: `-u` is the right way to update the snapshot, but it shouldn’t have been committed. Either remove it from the `package.json` again before creating the commit, or run something like `npm run test:unit -u` or `npx jest  -u` manually.

Though when I try it now, Jest seems to update the snapshot even without the `-u` option, which I find quite confusing – no idea what’s going on there…

lucaswerkmeister:
(For future reference: the `die` icon isn’t in the right position alphabetically, because it was originally called `random`. Maybe we can fix this the next time we add an icon. See also #640.)

PR #635, opened 2022-10-13
Create pre-release v2.1.0-alpha.14

guergana:
This pre-release includes icon 'die' that will be used in [Wikidata Mismatch Finder](https://github.com/wmde/wikidata-mismatch-finder).


lucaswerkmeister (changes requested):
https://github.com/wmde/wikit/pull/632#discussion_r994864551 should be fixed before this pre-release is made.

guergana:
> [#632 (comment)](https://github.com/wmde/wikit/pull/632#discussion_r994864551) should be fixed before this pre-release is made.

Fix is [PR 636](https://github.com/wmde/wikit/pull/636) :)

lucaswerkmeister (approved):
I just noticed that this downgrades the `tokens/package-lock.json` lockfileVersion from 2 to 1, which doesn’t seem nice – but the other three lockfiles also have version 1, so I guess it’s the `tokens` lockfile that was inadvertently updated at some earlier point (in 7c75971bf4, apparently). In the `next` branch, all the lockfiles are on version 2, so I think it’s fine to leave it like this now (version 1 on the legacy / master branch and 2 on the newer branch).

PR #641, opened 2022-11-15
Let extra Lookup attributes fall trough to the <input>

micgro42:
Bug: T322686

### DIFF
--- a/vue-components/src/components/Lookup.vue
+++ b/vue-components/src/components/Lookup.vue
@@ -1,6 +1,7 @@
 <template>
 	<div
-		:class="[ 'wikit', 'wikit-Lookup' ]"
+		:class="[ 'wikit', 'wikit-Lookup', extraClasses ]"
+		:style="extraStyles"
 	>
 		<span class="wikit-Lookup__label-wrapper">
 			<label
@@ -20,7 +21,6 @@
 			:feedback-type="feedbackType"
 			:menu-items="menuItems"
 			:disabled="disabled"
-			:aria-required="ariaRequired"
 			:placeholder="placeholder"
 			:value="value"
 			:search-input="searchInput"
@@ -28,6 +28,7 @@
 			@input="$emit('input', $event)"
 			@scroll="(firstIndex, lastIndex) => $emit('scroll', firstIndex, lastIndex)"
 			:label="label"
+			v-bind="otherAttributes"
 		>
 			<template #no-results>
 				<slot name="no-results" />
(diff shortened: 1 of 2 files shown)


lucaswerkmeister:
Looks good to me, but in #538/#540 we made the change on master and then cherry-picked it to next – should we cherry-pick this back to master as well?

Also, I’d like to see some tests for this; but since we didn’t add any tests for `TextInput` either, we can do that in a separate PR and add tests for both components at once there. (And if it doesn’t work out for some reason, it doesn’t block this PR.)

Not sure what to do about ui-review :/

micgro42:
UI review is happy now after I fixed it on the `next` branch. I made a retro note about it.

I'll look into adding a test.

Should we drop the `aria-required` prop and let it now be handled by the arg binding?

Backporting it to master is the plan :+1:

lucaswerkmeister:
> Should we drop the `aria-required` prop and let it now be handled by the arg binding?

I like that idea – it should even double as a test for this feature, since we have an existing test that the `<input>` ends up with an `aria-required` attribute :)

sai-san:
> UI review is happy now after I fixed it on the `next` branch. I made a retro note about it.
> 
> I'll look into adding a test.
> 
> Should we drop the `aria-required` prop and let it now be handled by the arg binding?
> 
> Backporting it to master is the plan 👍

I approved the UI changes in the relevant Chromatic build. Weird enough, it was complaining about spacing changes in the Icon component page.

micgro42:
> I approved the UI changes in the relevant Chromatic build some minutes ago. Weird enough, it was complaining about spacing changes in the Icon component page.

Ah, thank you! I hadn't noticed that! I thought my clicking all the buttons had worked eventually, but apparently it didn't  😅

Test has been added :)

micgro42:
> I like that idea – it should even double as a test for this feature, since we have an existing test that the `<input>` ends up with an `aria-required` attribute :)

* [ ]  TODO for tomorrow 👆

micgro42:
Done. Should be ready for another review :)

lucaswerkmeister (approved):
I thought we were going to keep the `ariaRequired` tests and only add the new test in a separate change, but okay ^^ I tested locally that the old tests also still pass if they’re added back.

micgro42:
> I thought we were going to keep the `ariaRequired` tests and only add the new test in a separate change, but okay ^^ I tested locally that the old tests also still pass if they’re added back.

I considered it, but judged it to be more noise than what is worth it. These tests implied that there is indeed an aria-required prop. So I merged the test that previously added `aria-required: true` into the new test, and dropped the one that just asserted the default behavior.

PR #656, opened 2022-12-15
Fix Lookup menu navigation index reset issue on arrow down

guergana:
Bug: T324743
Co-authored-by: Michael Große <micgro42@users.noreply.github.com>

### DIFF
--- a/vue-components/stories/Lookup.stories.ts
+++ b/vue-components/stories/Lookup.stories.ts
@@ -117,6 +117,8 @@ export function all(): Component {
 		template: `
 			<div>
 				<div style="margin-bottom: 20px">
+					<p>Try pressing <kbd>a</kbd> to see several search results. 
+					Some additional results will load when you scroll.</p>
 					<Lookup
 						:search-input.sync="search"
 						label="Label"
--- a/vue-components/stories/Lookup.stories.ts
+++ b/vue-components/stories/Lookup.stories.ts
@@ -67,7 +67,7 @@ const vegetableItems = [
 	},
 ];
 
-const addOnScroll = [
+const itemsToAddOnScroll = [
 	{
 		label: 'passionfruit',
 		description: 'sour sweet tropical fruit',
@@ -110,7 +110,7 @@ export function all(): Component {
 			onScroll( firstIndex: number, lastIndex: number ): void {
 				this.visibleItems = { firstIndex, lastIndex };
 				if ( lastIndex === this.menuItems.length - 1 ) {
-					this.menuItems = [ ...vegetableItems, ...addOnScroll ];
+					this.menuItems = [ ...vegetableItems, ...itemsToAddOnScroll ];
 				}
 			},
 		},
--- a/vue-components/stories/Lookup.stories.ts
+++ b/vue-components/stories/Lookup.stories.ts
@@ -98,7 +98,7 @@ export function all(): Component {
 		},
 		computed: {
 			menuItems: {
-				get(): MenuItem {
+				get(): MenuItem[] {
 					return this.menuVegetableItems.filter( ( item ) => item.label.includes( this.search ) );
 				},
 				set( newMenuItems: MenuItem[] ): void {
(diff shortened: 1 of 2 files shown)


vue-components/stories/Lookup.stories.ts
```diff
@@ -75,16 +93,25 @@ export function all(): Component {
...
 				selectedItem: null,
 				visibleItems: null,
+				menuVegetableItems: vegetableItems,
 			};
 		},
 		computed: {
-			menuItems(): MenuItem[] {
-				return vegetableItems.filter( ( item ) => item.label.includes( this.search ) );
+			menuItems: {
+				get(): MenuItem {
```
lucaswerkmeister, line 101:
```suggestion
				get(): MenuItem[] {
```

vue-components/stories/Lookup.stories.ts
```diff
@@ -67,6 +67,24 @@ const vegetableItems = [
 	},
 ];
 
+const addOnScroll = [
```
lucaswerkmeister, line 70: This variable name sounds a bit confusing to me (more like an action than a list). How about `fruitItems`, by analogy with `vegetableItems`? (Are sweet potatoes fruit? :sweat_smile:)

vue-components/stories/Lookup.stories.ts
```diff
@@ -1,7 +1,7 @@
 import Lookup from '@/components/Lookup';
 import Icon from '@/components/Icon';
-import { Component } from 'vue';
 import { MenuItem } from '@/components/MenuItem';
+import { Component } from 'vue';
```
lucaswerkmeister, line 4: I guess it makes a lot more sense to keep all the `@/components` imports together, yeah :+1:

lucaswerkmeister (commented):
I think it would be useful to add some information to the story about how to trigger the scroll behavior (press <kbd>a</kbd> – <kbd>e</kbd> doesn’t match enough items, even though it’s the most common letter in English).

vue-components/stories/Lookup.stories.ts
guergana: I don't think sweet potatoes are a fruit. I added this name because I think this makes it clearer as to what is the purpose of this variable. You are right about adding the instructions, but for me pressing `a` gives enough items to make the test. There are a total of 11 items with the letter `a`, the original menuItem count with `a` is 8 and then it increases to 11

guergana: I am not sure where to add these instructions, tho... I don't see any other Component with instructions.

lucaswerkmeister: > but for me pressing a gives enough items to make the test

Yes, for me too, that’s why I think it should be in the instructions ^^

vue-components/stories/Lookup.stories.ts
```diff
@@ -75,16 +93,25 @@ export function all(): Component {
...
 		},
 		methods: {
 			onScroll( firstIndex: number, lastIndex: number ): void {
 				this.visibleItems = { firstIndex, lastIndex };
+				if ( lastIndex === this.menuItems.length - 1 ) {
+					this.menuItems = [ ...vegetableItems, ...addOnScroll ];
+				}
 			},
 		},
 		template: `
```
lucaswerkmeister, line 117: I think we could add the instructions somewhere below here (GitHub won’t let me comment on the right line :rage:), something like:

```html
<p>
Try pressing <kbd>a</kbd> to see several search results. Some additional results will load when you scroll.
</p>
```

vue-components/stories/Lookup.stories.ts
guergana: Ah, then I misunderstood. I thought you were not getting extra results when testing with the letters a and e.

lucaswerkmeister: No, it works for me with a, but my first attempt was e :) (and then a bunch of random letters)

lucaswerkmeister: I still think this variable name is confusing, though. How about `itemsToAddOnScroll`?

guergana: sure. that sounds like a good name. :)

lucaswerkmeister (approved):
As far as I can tell, this works great, including the case where the lookup is closed and reopened (the selection is preserved).

PR #657, opened 2022-12-21
Create pre-release of version 2.1.0-alpha.15

guergana:


lucaswerkmeister (changes requested):
I think we should add the changes to the `CHANGELOG.md` too – something like:

> Let Lookup attributes fall through to `<input>`
> Fix Lookup navigation on arrow down

guergana:
> I think we should add the changes to the `CHANGELOG.md` too – something like:
> 
> > Let Lookup attributes fall through to `<input>`
> > Fix Lookup navigation on arrow down

I thought changlog updates were not necessary for pre-releases

lucaswerkmeister:
Hm, then the docs don’t seem to match our current practice :sweat_smile: I noticed this based on the changes in earlier pre-releases (e.g. 1d8da10467, 533965be1e); I also think it can help to collect the changes earlier, so we don’t have as much work to do when we make the final release?

guergana:
> Hm, then the docs don’t seem to match our current practice :sweat_smile: I noticed this based on the changes in earlier pre-releases (e.g. 1d8da10467, 533965be1e); I also think it can help to collect the changes earlier, so we don’t have as much work to do when we make the final release?

I probably never added a release change before when making a prerelease.

lucaswerkmeister (approved):

PR #658, opened 2022-12-21
Fix Lookup menu navigation index reset issue on arrow down

guergana:
Bug: T324743


lucaswerkmeister (approved):
Seems to work fine AFAICT :+1:

PR #661, opened 2022-12-29
Replace set output

guergana:
Bug: T322548

This is a followup to #660 which was created as a test, but can only be tested on PR open.


guergana:
Wanted to test that action on open was working but didn't push the correct commit. Closing to repeat this process on a new PR.

PR #662, opened 2022-12-29

guergana:
Bug: T322548


guergana:
Same as #660 and #661

guergana:
Same as #660 and #661

PR #665, opened 2022-12-29

guergana:
Bug: T322548


itamargiv (approved):
Looking good, thanks!

PR #674, opened 2023-01-13
Add aria-controls attribute to LookupInput

lucaswerkmeister:
This should be set to the same ID as the `aria-owns` attribute: the input is not only a functional parent of the options menu (`aria-owns`), but also controls its contents and presence (`aria-controls`).

Bug: T326681

### DIFF
--- a/vue-components/src/components/LookupInput.vue
+++ b/vue-components/src/components/LookupInput.vue
@@ -14,6 +14,7 @@
 			autocomplete="off"
 			v-bind="$attrs"
 			:aria-activedescendant="keyboardHoverId"
+			:aria-controls="optionsMenuId"
 			:aria-owns="optionsMenuId"
 			aria-autocomplete="list"
 			aria-haspopup="listbox"


micgro42:
Mh, reading more about `aria-controls`, it seems its primary purpose is to allow jumping the controlled element. However, this is not something we want, because it would remove the focus from the input and close the menu.
At least it looks like that from reading https://tink.uk/using-the-aria-controls-attribute/ and https://a11ysupport.io/tech/aria/aria-controls_attribute

However, when looking at MDN, then a combobox is explicitly listed as one of the example use-cases: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

So, I'm not really sure what the right thing to do here is.

micgro42:
Mh. Then let's move forward and see if we happen to get any feedback from anyone.

micgro42 (approved):

lucaswerkmeister:
Okay, let’s try it…

PR #675, opened 2023-01-16
Cherry-pick 'Add aria-controls attribute to LookupInput (#674)'

lucaswerkmeister:
This should be set to the same ID as the aria-owns attribute: the input is not only a functional parent of the options menu (aria-owns), but also controls its contents and presence (aria-controls).

Bug: T326681


guergana (approved):
LGTM. Thanks

PR #682, opened 2023-08-08
Fix Lookup component scrollbar height after open and closing menu

guergana:
Note: this is not reproducible in Wikit, please reproduce in special-new-lexeme. Instructions in the phabricator ticket.

Bug: [T325822](https://phabricator.wikimedia.org/T325822)


micgro42 (approved):
I reproduced this error in storybook on `master` (with wrapping the added results in a `window.setTimeout()`) and confirmed that it works as expected with this change.

guergana:
> I reproduced this error in storybook on `master` (with wrapping the added results in a `window.setTimeout()`) and confirmed that it works as expected with this change.

Super! It didn't occur to me before to test that way ,thanks for the tip.

PR #684, opened 2023-08-14
[WiP] Update node version to 16 in github ci

guergana:
Checks are failing at [this step](https://github.com/wmde/wikit/actions/runs/5800751927/job/15723582350#step:5:92) in the checks. This error was probably introduced when upgrading github actions that don't support node 12 in [this commit](https://github.com/wmde/wikit/commit/6a7c6cbc18833eb82110b479c3745a5741f96235).

Upgrading to node 16 works partially but now tests are complaing about the deprecation of `lerna bootstrap`.

### DIFF
--- a/.github/workflows/pre-release.yaml
+++ b/.github/workflows/pre-release.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup
         uses: actions/setup-node@v3
         with:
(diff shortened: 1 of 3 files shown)


guergana:
> 

oh my, i had changed it but forgot to save the files. :see_no_evil:

PR #685, opened 2023-08-14
Fix Lookup component scrollbar height after open and closing menu

guergana:
Bug: T325822


micgro42 (approved):
Same as #682, so I assume that this is fine.

PR #688, opened 2023-08-15
Fix CI for PRs on master

micgro42:
This does three things:
1. it makes sure we explicitly use _our_ lerna, not the global one that seems to come with `ubuntu_latest` in the build step
2. it makes sure we use the storybook CLI that matches our storybook version when extracting the stories, `npx sb` would install the latest version if it is not already in node_modules
3. it updates chromatic because there was an issue with the chromatic part failing despite a change already being accept in the chromatic web interface (this PR does not touch components at all)

### DIFF
--- a/.github/workflows/test-and-demo.yaml
+++ b/.github/workflows/test-and-demo.yaml
@@ -57,7 +57,7 @@ jobs:
       run: npm ci
     - name: Build  # only runs in case of cache hit, otherwise as part of npm ci
       if: steps.cache.outputs.cache-hit == 'true'
-      run: lerna bootstrap
+      run: npm run lerna bootstrap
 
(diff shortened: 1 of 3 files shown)


.github/workflows/test-and-demo.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
       run: npm ci
     - name: Build  # only runs in case of cache hit, otherwise as part of npm ci
       if: steps.cache.outputs.cache-hit == 'true'
-      run: lerna bootstrap
+      run: npm run lerna bootstrap
```
guergana, line 60: i see, good catch!

chukarave (approved):

PR #689, opened 2023-08-15
Explicitly use consistent node version in post-merge workflow

micgro42:
If that is not explicitly set, CI will use whatever comes with `ubuntu_latest` and that is not what we want for this workflow. Node 14 is consistent with what is used in the other workflows. The PR to upgrade these is #684.

This should be merged after #688 and make CI green again on merged branches.


chukarave (approved):
Let's see how the post-merge workflow find this!

